PR #125, opened 2021-02-19
Allow oe_num_mem_pages to be overridden

paulcallen:
Calculation of this setting is done via sighing and packaging.
If the unsigned version is used it uses the default enclave setting
which is not always big enough.
This change allows that setting to be updated for unsigned binaries.
This changed also updated algorithm a little as the reserved space was
not enough for a number of scenarios.


tools/myst/host/utils.h
```diff
@@ -34,4 +34,6 @@ int cli_getopt(
     const char* opt,
     const char** optarg);
 
+#define make_oe_num_heap_pages(rootfs_size) ((rootfs_size + (50 * 1024 * 1024)) / PAGE_SIZE)
```
mikbras, line 37: Our coding conventions for macros is upper case.

mikbras (changes requested):
Looks great. The only thing I could find is a nit. :) Our coding convention uses upper case for macros. You could make it an inline function for better type checking.

tools/myst/host/utils.h
mikbras:
```
MYST_INLINE size_t make_oe_num_heap_pages(size_t rootfs_size)
{
    const size_t extra_bytes = 50 * 1024 * 1024;
    return (rootfs_size + extra_bytes) / PAGE_SIZE;
}
```

paulcallen:
Abandoning

PR #131, opened 2021-02-22
Document exposed OE APIs

jxyang:

### DIFF
--- a/doc/exposed-oe-apis.md
+++ b/doc/exposed-oe-apis.md
@@ -1,0 +1,61 @@
+# Exposed APIs from Open Enclave SDK in Mystikos
+
+Mystikos is linked with [Open Enclave SDK](https://openenclave.io/sdk/)
+libraries. Through an extended system call interface, Mystikos is able
+to expose the following set of OE APIs to user applications. Please consult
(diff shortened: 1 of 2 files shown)


mikbras (commented):
I think it would be hard to use the OE syscalls directly without the types (consider **oe_report_t**). Including **<openenclave/enclave.h>** solves half of the problem by making the data types available that are passed as OE syscall parameters. The second part of the problem is solved by linking with **liboeenclave.a**, which provides stubs for invoking the OE syscalls in a type-safe manner.

But as we discussed, it would probably be better to provide an abstract SGX API that would work for both the **OE SDK** and the **Intel SGX SDK**. Maybe **<myst/sgx.h>**? I am not sure whether we should add this document before deciding how we want to surface SGX features.

jxyang:
Agreed. We should hold the publishing of this document until a Mystikos implementation that is TEE-agnostic and SDK-agonistic. Basically, Mystikos has to translate between general purpose TEE-relevant structures and the underlying specific TEE structures.

This document serves the purpose of identifying what generic APIs user applications expect from Mystikos. Most of the content can be reused later, when time is ready.

PR #132, opened 2021-02-22
Add syscall/kernel limitations

jxyang:

### DIFF
--- a/doc/kernel-limitations.md
+++ b/doc/kernel-limitations.md
@@ -1,0 +1,169 @@
+# Notable Kernel limitations in the current Mystikos implementation
+
+The first hardware TEE that Mystikos based on is Intel SGX. Therefore, Mystikos
+carries most limitations imposed by Intel SGX. However, we expect those
+limitations can be relaxed as better hardware TEE platforms emerge, or when
+Mystikos overcomes a hardware limitation with software emulation. In fact,
+[SGX2](https://caslab.csl.yale.edu/workshops/hasp2016/HASP16-16_slides.pdf)
+has already relaxed a few limitations imposed by SGX1, as shown below.
+
+By and large, Mystikos implements a Linux-compatible kernel so it could
+run Linux applications with little or no modifications. There are still
+incompatibilities/limitations worth noting as summarized below:
+
+| Limitation                    | Description         | Potential impacts on applications |
+| ------------------------------|---------------------|---------|
+| [Multi-process limitation](#limitations-arisen-from-sgxs-single-process-model) | No forking of child processes. However `posix_spawn` is supported. | Communication and address space isolation between processes |
+| [Thread limitations](#limitations-arisen-from-sgx1-limitations) | Thread creation is slow and limited to a user setting. Thread scheduling is subject to attacks by host. | Required Configuration / Thread starvation / D.o.S. |
+| [Memory limitations](#limitations-arisen-from-sgx1-limitations) | User stack/heap memory is limited to a user setting. Page permission enforcement is under host control. | Required Configuration / OOM / Page table manipulation |
+| [Clock limitation](#limitations-arisen-from-lack-of-access-to-time-source) | Clock ticks and resolution are controlled by host | Clock resolution / Untrusted time |
+| [Exception handling limitation](#hardware-exception-handling-limitations) | Direct #PF/#GP exception handling inside the Mystikos environment is not supported | Missed or arbitrary exceptions / Exception handling with long delays |
+| [Network limitation](#network-limitations)  | Networking has to pass through host | Missed or delayed packages / Eavesdropping / D.o.S. |
+| [File system limitations](#file-System-limitations) | Support ramfs, hostfs, and ext2 only. Mounting of a file system has to be explicit. Changes are not always persisted. | Required FS conversion / Unprotected files once persisted to outside of TEE   |
+
+* Note, many attacks above are only possible with a malicious host OS/kernel,
+which is deliberately excluded from TCB by the SGX threat model.
+
+## Limitations arisen from SGX's single process model
+
+With SGX, an enclave application runs as a single process in a single
+address space. Mystikos tries to emulate a process with a thread. When we
+create a child process, we actually create a child thread with proper
+process specific information, including `pid`, `ppid`, `fd table`, etc.
+
+This approach would allow child processes created with `posix_spawn` to
+run alongside the parent process. Sure, there is no memory address space
+separation between the processes, but the child or the parent would have
+no handle to reference the other side's memory.
+
+On the other hand, a child process created with `fork` is supposed to
+share variables with the parent. It's significantly more challenging to
+emulating a forked child process as a thread without underlying support
+of address space separation.
+
+
+## Limitations arisen from SGX1 limitations
+
+### SXG1 Thread model
+
+With SGX1, the number of threads running inside an enclave, i.e.,
+**ethread**, has to be statically declared. Once the application starts,
+no more ethreads can be dynamically created. Meanwhile, the stack size
+of each ethread is also statically declared and allocated while creating
+the enclave. An application attempting to create more ethreads than what
+is declared will crash.
+
+A Mystikos application declares how many user threads are needed in
+config.json. A potential mitigation of the limitation is to make the setting big
+enough. The caveat is that we should ensure the stack size for each ethread
+is reasonable so we don't waste too much stack memory on unused threads.
+
+This limitation is likely to be relaxed when EDMM of SGX2 is officially
+supported in Mystikos.
+
+When an user application calls `pthread_create`, Mystikos makes an OCALL
+to the host to create a host thread, and re-enters the enclave with an
+ECALL, and once inside the enclave, jump to the thread routine. When an user
+application calls `pthread_yield`, Mystikos makes an OCALL to the host to
+request the host to suspend the calling thread. The long
+chain of events leads to delays and performance problems occasionally.
+
+### SGX1 Memory model
+
+With SGX1, the heap size of the enclave application has to be statically
+declared. Once the application starts, allocating more memory than what's
+declared causes a crash due to out of memory.
+
+A Mystikos application declares how big the heap size is needed in config.json.
+A potential mitigation of the limitation is to make the setting big enough.
+The downside of doing that is: 1) the SGX runtime will take more time to
+add/extend heap pages one by one during the initialization, and 2) The heap
+size is backed by physical EPC memory of the system. If the declared heap
+size is much more than the available physical EPC memory size, the application
+performance is likely to suffer due to severe `EPC paging`.
+
+Also with SGX1, the untrusted host can manipulate the page permissions at
+will because the page tables are solely controlled by the host. The implies
+the host can trigger page fault whenever and/or at whichever address it likes.
+
+These limitations are likely to be relaxed when EDMM of SGX2 is officially
+supported in Mystikos.
+
+## Limitations arisen from lack of access to time source
+
+Typically hardware time stamp counters are not available in user space. With
+SGX, the enclave application runs in ring 3, therefore it's generally
+forbidden to call `RDTSC` or `RDTSCP`. (They are legal inside
+an enclave for processors that support SGX2, but subject to the value of
+`CR4.TSD`).
+
+Without a direct time source, Mystikos relies on a dedicated host thread that
+periodically updates a clock shared by the host and the enclave. The host
+decides how often to update and what's the updated value. Obviously this
+has security implications.
+
+Mystikos ensures clock monotonicity. But the host can still advance the clock
+with a faster or slower speed. An application depending on trusted time should
+query a trusted time server to detect such attacks.
+
+## Hardware exception handling limitations
+
+When hardware exceptions happen, the execution context is saved, and the ISR
+of the kernel is invoked. This is even true for exceptions triggered by code
+running inside an SGX enclave. It has to perform an `enclave exit` (`AEX`)
+before invoke the kernel handler, and re-enter enclave with `AEP` after it.
+This basically means the untrusted host is trusted with the confidential
+execution context, e.g., the register values, when a hardware exception
+occurs. It also means that the untrusted host is in the middle of the handling
+loop, thus is able to withhold, manipulate, or manufacture hardware
+exceptions to the enclave.
+
+This security impact is likely to be mitigated when SGX2 is officially
+supported in Mystikos. With SGX2, the CPU saves full `EXIT_INFO`
+for #PF/#GP inside the enclave before `AEX`. With the full
+`EXIT_INFO`, the enclave exception handler can handle #PF/#GP exception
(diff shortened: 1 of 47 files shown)


doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+or slower speed. An application depending on trusted time should query
+a trusted time server to detect such attacks.
+
+## Limitations arisen from handling of hardware exceptions by OS
+
+When hardware exceptions happen, the execution context is saved, and the ISR
+of the kernel is invoked. This is even true for exceptions triggered by code
+running inside an SGX enclave. It has to perform an `enclave exit` (`AEX`)
+before invoke the kernel handler, and re-enter enclave with `AEP` after it.
+This basically means the untrusted host is trusted with the confidential
```
bodzhang, line 118:
```suggestion
The SGX CPU makes sure the untrusted kernel handler cannot observe SGX enclave execution context, e.g., CPU register values, by recording the execution context within the Enclave memory and replacing the context with a "synthetic" context before invoking the kernel ISR. With host assistance, trusted exception handler inside the Enclave can be invoked to handle the exception or abort the enclave execution. If the untrusted host does not invoke the trusted exception handler and "resume" the enclave execution, the same HW exception would be triggered again and no forward progress will be made in Enclave execution. With SGX1, for #PF/#GP faults, only partial exception context is recorded within the Enclave memory. Lack of full exception context inside the Enclave for certain exceptions makes it impossible to support trusted exception handling inside the Enclave for the affected exceptions.
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+execution context, e.g., the register values, when a hardware exception
```
bodzhang, line 119:
```suggestion
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+occurs. It also means that the untrusted host is in the middle of the handling
```
bodzhang, line 120:
```suggestion
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+loop, thus is able to withhold, manipulate, or manufacture hardware
```
bodzhang, line 121:
```suggestion
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+exceptions to the enclave.
```
bodzhang, line 122:
```suggestion
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+separation between the processes, but the child or the parent would have
+no handle to reference the other side's memory.
+
+On the other hand, a child process created with `fork` is supposed to
+share variables with the parent. It's significantly more challenging to
+emulating a forked child process as a thread without underlying support
+of address space separation.
+
+
+## Limitations arisen from SGX's thread model
```
bodzhang, line 42:
```suggestion
## Thread limitation
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+order of ethreads.
+
+The security and performance impact of this limitation can be mitigated with
+an enclave-resident thread scheduler, which uses M:N mapping to map ethreads
+to user application pthreads, and performs context switching between pthreads.
+
+This limitation is likely to be relaxed when EDMM of SGX2 is officially
+supported in Mystikos.
+
+## Limitations arisen from SGX's memory model
```
bodzhang, line 72:
```suggestion
## Memory limitation
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+
+A Mystikos application declares how big the heap size is needed in config.json.
+A potential mitigation of the limitation is to make the setting big enough.
+The downside of doing that is: 1) the SGX runtime will take more time to
+add/extend heap pages one by one during the initialization, and 2) The heap
+size is backed by physical EPC memory of the system. If the declared heap
+size is much more than the available physical EPC memory size, the application
+performance is likely to suffer due to severe `EPC paging`.
+
+Also with SGX1, the untrusted host can manipulate the page permissions at
```
bodzhang, line 86:
```suggestion
Also with SGX1, the Enclave cannot change Enclave enforced page access restriction after the Enclave is initialized. It's possible for the enclave SW to request the host to enforce the restriction in the page tables, but with the SGX thread model, the Enclave SW should assume the host might silently drop the requests.
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+will because the page tables are solely controlled by the host. The implies
```
bodzhang, line 87:
```suggestion
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+the host can trigger page fault whenever and/or at whichever address it likes.
```
bodzhang, line 88:
```suggestion
```
In both SGX thread model and many other TEEs' thread model, page access related, #PF or EPT fault can be triggered anytime by the privileged system SW outside of the TCB.

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+decides how often to update and what's the updated value. Obviously this
+has security implications.
+
+To protect against a malicious host, Mystikos ensures the reading of the
+clock is at least one over the previous reading. The clock monotonicity
+is maintained this way. The host can still advance the clock with a faster
+or slower speed. An application depending on trusted time should query
+a trusted time server to detect such attacks.
+
+## Limitations arisen from handling of hardware exceptions by OS
```
bodzhang, line 112:
```suggestion
## Exception handling limitation
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+is declared will crash.
+
+A Mystikos application declares how many user threads are needed in
+config.json. A potential mitigation of the limitation is to make the setting big
+enough. The caveat is that we should ensure the stack size for each ethread
+is reasonable so we don't waste too much stack memory on unused threads.
+
+When an user application calls `pthread_create`, Mystikos makes an OCALL
+to the host to create a host thread, and re-enters the enclave with an
+ECALL, and once inside the enclave, jump to the thread routine. The long
```
bodzhang, line 58:
```suggestion
ECALL, and once inside the enclave, jump to the thread routine. When an user application calls `pthread_yield`, Mystikos makes an OCALL to the host to request the host to suspend the calling thread. The long
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+chain of events leads to delays and performance problems occasionally.
+
+The security implication of using host threads is that the untrusted host
```
bodzhang, line 61:
```suggestion
The host is able to starve or switch execution order of ethreads, but because Mystikos implement synchronization primitives inside the SGX Enclave, the host cannot use the thread scheduling interface to cause unexpected synchronization error as long as the Enclave SW utilizes the synchronization primitives properly.
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+controls the thread scheduling, and may decide to starve or switch execution
```
bodzhang, line 62:
```suggestion
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
...
+order of ethreads.
```
bodzhang, line 63:
```suggestion
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,174 @@
+# Notable Kernel limitations of Mystikos
```
bodzhang, line 1: This doc primarily discusses  Mystikos-SGX security design. Most of  the covered security design topics are captured in Open Enclave design docs and the Mystikos implemented same design as Open Enclave.  If the doc is posited as a security design doc, it misses many other security design topics. I think for the syscall/kernel limitation docs, we can focus on the limitation, without the in-depth discussion of the security design choices behind the limitation. We can capture security design topics in separate docs.

bodzhang (changes requested):

doc/kernel-limitations.md
jxyang: This doc tries to answer these questions: 
> As an application developer, what do I need to know regarding Mystikos kernel vs. Linux kernel in order to write a performant and safe program that can be lift-n-shifted into a TEE? And when it runs in a TEE, what could be the incompatibility from running on Linux?

These challenges require a holistic approach from the beginning. Security really should be a first consideration, and shouldn't be deferred to another project that doesn't have a kernel.

I think we can take out the security design choices but leave the security limitations in place.

doc/syscall-limitations.md
```diff
@@ -0,0 +1,142 @@
...
+| SYS_chroot | change root directory | Unsupported |
+| SYS_statfs / SYS_fstatfs | get file system status | Partial |
+| SYS_sync / SYS_syncfs | commit filesystem caches to disk | Unsupported |
+
+## System calls related to low level memory management
+
+| Syscall names        | Description             | Compatibility |
+| -------------------- |-------------------| --------------|
+| SYS_mmap                | allocate memory pages | Partial |
+| SYS_msync               | flush mmaped files | Partial |
```
mikbras, line 120: Thank you for updating.

doc/kernel-limitations.md
bodzhang: Untrusted privileged system SW involved in MMU control, for example, OS page table manger, VMM EPT manager can always manipulate the page permission in the construct it controls and trigger page fault. Protection against DoS attack from the privileged system SW is typically outside of the scope of TEE Technologies. This is true for Intel SGX, as well as AMD SEV. The security objective is that the TEE can implement its own page permission restriction, even if the untrusted privileged system SW does not enforce the policy (TEE's page permission policy is a subset of the untrusted policy). Even if the untrusted VMM (for both Intel SGX and AMD SEV) does not enforce any page access permission policy, with Intel SGX, the Enclave can enforce its own page permission policy in SGX EPCM(OS page table policy is untrusted for SGX too) , and with AMD SEV, the trusted kernel in a SEV protected VM can enforce its own page permission policy through OS page table. The result is that the trusted page access restriction maintained by the TEE is honored by the MMU, not the potentially more relaxed policy set by the untrusted system SW. The feature limitation of SGX1 is that the Enclave cannot change the Enclave enforced page access restriction after the Enclave is initialized. SGX2.0 does not have this limitation.

doc/syscall-limitations.md
```diff
@@ -0,0 +1,142 @@
...
+
+
+## Miscellaneous system calls
+
+| Syscall names        | Description             | Compatibility |
+| -------------------- |-------------------| --------------|
+| SYS_syslog              | read and/or clear kernel message ring buffer | Unsupported |
+| SYS_futex               | futex operations | Partial |
+| SYS_get_thread_area     | get thread-local storage | Unsupported |
+| SYS_clock_nanosleep     | high-resolution sleep | Unsupported |
```
bodzhang, line 136: It seems that nanosleep() is supported through an OCALL

doc/kernel-limitations.md
```diff
@@ -0,0 +1,169 @@
...
+has already relaxed a few limitations imposed by SGX1, as shown below.
+
+By and large, Mystikos implements a Linux-compatible kernel so it could
+run Linux applications with little or no modifications. There are still
+incompatibilities/limitations worth noting as summarized below:
+
+| Limitation                    | Description         | Potential impacts on applications |
+| ------------------------------|---------------------|---------|
+| [Multi-process limitation](#limitations-arisen-from-sgxs-single-process-model) | No forking of child processes. However `posix_spawn` is supported. | Communication and address space isolation between processes |
+| [Thread limitations](#limitations-arisen-from-sgx1-limitations) | Thread creation is slow and limited to a user setting. Thread scheduling is subject to attacks by host. | Required Configuration / Thread starvation / D.o.S. |
```
bodzhang, line 17:
```suggestion
| [Thread limitations](#limitations-arisen-from-sgx1-limitations) | Thread creation is slow and limited to a user setting. Thread scheduling is under host control. | Required Configuration / Thread starvation / D.o.S. |
```

doc/kernel-limitations.md
```diff
@@ -0,0 +1,169 @@
...
+| [Memory limitations](#limitations-arisen-from-sgx1-limitations) | User stack/heap memory is limited to a user setting. Page permission enforcement is under host control. | Required Configuration / OOM / Page table manipulation |
+| [Clock limitation](#limitations-arisen-from-lack-of-access-to-time-source) | Clock ticks and resolution are controlled by host | Clock resolution / Untrusted time |
+| [Exception handling limitation](#hardware-exception-handling-limitations) | Direct #PF/#GP exception handling inside the Mystikos environment is not supported | Missed or arbitrary exceptions / Exception handling with long delays |
```
bodzhang, line 20:
```suggestion
| [Exception handling limitation](#limitations-arisen-from-sgx1-limitations) | Direct #PF/#GP exception handling inside the Mystikos environment is not supported |  Exception handling with long delays |
```
#FP/#GP exception will not be missed. If exception is triggered either based on OS/VMM page table or SGX EPCM, the SGX enclave will go through SGX AEX and the Enclave execution will not make forward progress even if the OS does not invoke the trusted exception handler inside the Enclave. If the OS invokes the trusted exception handler, the handler will abort Enclave execution, when it detects that the trusted exception info is not present inside the Enclave. "Arbitrary exception" terminology does not paint the full picture. The system SW can cause interrupt or #PF exception, which typically is a security issue for TEE. See discussion in another comment.

doc/kernel-limitations.md
```diff
@@ -0,0 +1,169 @@
...
+| [Network limitation](#network-limitations)  | Networking has to pass through host | Missed or delayed packages / Eavesdropping / D.o.S. |
```
bodzhang, line 21:
```suggestion
| [Network limitation](#network-limitations)  | Networking has to pass through host | Missed or delayed packets / Eavesdropping of unencrypted packets / D.o.S. |
```

doc/syscall-limitations.md
jxyang: That's true for `SYS_nanosleep`. `clock_nanosleep` is more like an advanced `nanosleep` in which you can choose which clock to sleep on.

doc/kernel-limitations.md
jxyang: @bodzhang Are you referring to the futex primitive inside the enclave? I feel the suggested changes are too broad and too detailed. Can we abstract/scope it somehow?

jxyang: We decided to take this one.

jxyang:
@bodzhang I believe all your comments are addressed. Please review.

jxyang:
Seems `commit suggestions on github` caused difficulty to rebase. I will drop this one and create another PR.

jxyang:
Superseded by https://github.com/deislabs/mystikos/pull/156

PR #133, opened 2021-02-23
msync syscall

mikbras:
Add support for msync syscall. Add msync mapping structures that
are created by mmap(), syncrhoized by msync(), and removed by
munmap().

### DIFF
--- a/kernel/libc.c
+++ b/kernel/libc.c
@@ -7,6 +7,7 @@
 #include <string.h>
 
 #include <myst/backtrace.h>
+#include <myst/syscall.h>
 #include <myst/defs.h>
 #include <myst/eraise.h>
 #include <myst/kernel.h>
@@ -720,7 +721,7 @@ ssize_t pread(int fd, void* buf, size_t count, off_t offset)
     return myst_syscall_ret(myst_syscall_pread(fd, buf, count, offset));
 }
 
-ssize_t myst_pwrite(int fd, const void* buf, size_t count, off_t offset)
+ssize_t pwrite(int fd, const void* buf, size_t count, off_t offset)
 {
     return myst_syscall_ret(myst_syscall_pwrite(fd, buf, count, offset));
 }
@@ -798,6 +799,15 @@ ssize_t readlink(const char* pathname, char* buf, size_t bufsiz)
 int symlink(const char* target, const char* linkpath)
 {
     return (int)myst_syscall_ret(myst_syscall_symlink(target, linkpath));
+}
+
+int fcntl(int fd, int cmd, ...)
+{
+    va_list ap;
+    va_start(ap, cmd);
+    long arg = va_arg(ap, long);
+    va_end(ap);
+    return (int)myst_syscall_ret(myst_syscall_fcntl(fd, cmd, arg));
 }
 
 /*
--- a/kernel/libc.c
+++ b/kernel/libc.c
@@ -801,15 +801,6 @@ int symlink(const char* target, const char* linkpath)
     return (int)myst_syscall_ret(myst_syscall_symlink(target, linkpath));
 }
 
-int fcntl(int fd, int cmd, ...)
-{
-    va_list ap;
-    va_start(ap, cmd);
-    long arg = va_arg(ap, long);
-    va_end(ap);
-    return (int)myst_syscall_ret(myst_syscall_fcntl(fd, cmd, arg));
-}
-
 /*
 **==============================================================================
 **
--- a/kernel/mmanutils.c
+++ b/kernel/mmanutils.c
@@ -15,6 +15,7 @@
 #include <myst/mmanutils.h>
 #include <myst/process.h>
 #include <myst/strings.h>
+#include <myst/round.h>
 
 #define SCRUB
 
@@ -335,7 +336,7 @@ static int _release_msync_mappings(void* addr, size_t length)
 
                     /* update the left mapping length */
                     p->length = llength;
-                    prev = p;
+                    prev = rm;
                 }
                 else if (llength)
                 {
@@ -385,12 +386,11 @@ int myst_munmap(void* addr, size_t length)
     int ret = 0;
 
     /* address cannot be null and must be aligned on a page boundary */
-    if (!addr || ((uint64_t)addr % PAGE_SIZE))
+    if (!addr || ((uint64_t)addr % PAGE_SIZE) || !length)
         ERAISE(-EINVAL);
 
-    /* length cannot be 0 and must be aligned on a page boundary */
-    if (!length || (length % PAGE_SIZE))
-        ERAISE(-EINVAL);
+    /* align length to a page boundary */
+    ECHECK(myst_round_up(length, PAGE_SIZE, &length));
 
     ECHECK(myst_mman_munmap(&_mman, addr, length));
 
--- a/kernel/mmanutils.c
+++ b/kernel/mmanutils.c
@@ -2,19 +2,19 @@
 // Licensed under the MIT License.
 
 #include <assert.h>
-#include <stdlib.h>
-
 #include <limits.h>
-#include <myst/file.h>
-#include <myst/mmanutils.h>
-#include <myst/strings.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/mman.h>
 #include <unistd.h>
 
+#include <myst/syscall.h>
 #include <myst/eraise.h>
+#include <myst/file.h>
+#include <myst/mmanutils.h>
 #include <myst/process.h>
+#include <myst/strings.h>
 
 #define SCRUB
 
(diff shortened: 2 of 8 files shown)


kernel/mmanutils.c
```diff
@@ -99,14 +107,48 @@ static ssize_t _map_file_onto_memory(
...
+        {
+            msync_mapping_t* fm;
+
+            if (!(fm = calloc(1, sizeof(msync_mapping_t))))
+            {
+                myst_spin_unlock(&_msync_mappings_lock);
+                ERAISE(-ENOMEM);
+            }
+
+            fm->fd = fd;
```
jxyang, line 138: It seems inefficient to create a msync_mapping_t entry for every page. Can we optimize?

kernel/mmanutils.c
```diff
@@ -184,9 +226,59 @@ void* myst_mremap(
...
+
+    myst_spin_lock(&_msync_mappings_lock);
+    {
+        msync_mapping_t* p = _msync_mappings;
+        msync_mapping_t* prev = NULL;
+        void* addr_end = (uint8_t*)addr + length;
+
+        while (p)
+        {
+            msync_mapping_t* next = p->next;
```
jxyang, line 242: This is also very inefficient.

kernel/mmanutils.c
```diff
@@ -308,3 +401,130 @@ int myst_release_process_mappings(pid_t pid)
...
+
+    myst_spin_lock(&_msync_mappings_lock);
+    {
+        void* addr_end = (uint8_t*)addr + length;
+        msync_mapping_t* head = NULL;
+
+        /* flush any msync mappings contained by this address range */
+        for (msync_mapping_t* p = _msync_mappings; p; p = p->next)
+        {
+            void* p_addr_end = (uint8_t*)p->addr + p->length;
```
jxyang, line 460: Same performance problem.

kernel/mmanutils.c
```diff
@@ -308,3 +401,130 @@ int myst_release_process_mappings(pid_t pid)
...
+
+/* notified on close to remove msync mappings involving fd */
+int myst_mman_close_notify(int fd)
+{
+    int ret = 0;
+
+    if (fd < 0)
+        ERAISE(-EINVAL);
+
+    myst_spin_lock(&_msync_mappings_lock);
```
jxyang, line 500: Skip if file was opened with READ only.

tests/msync/msync.c
```diff
@@ -0,0 +1,184 @@
...
+    /* map the file onto memory */
+    const size_t length = st.st_size;
+    const int prot = PROT_READ | PROT_WRITE;
+    const int flags = MAP_PRIVATE | MAP_SHARED;
+    uint8_t* addr = mmap(NULL, length, prot, flags, fd, 0);
+    assert(addr != MAP_FAILED);
+
+    /* perform a second mapping of the same file */
+    uint8_t* addr2 = mmap(NULL, length, prot, flags, fd, 0);
+    assert(addr2 != MAP_FAILED);
```
jxyang, line 112: We should also check addr2 != addr

kernel/mmanutils.c
```diff
@@ -99,14 +107,48 @@ static ssize_t _map_file_onto_memory(
...
             r -= (size_t)n;
             bytes_read += n;
         }
     }
 
-    /* restore the file position */
-    if (lseek(fd, save_pos, SEEK_SET) == (off_t)-1)
-        goto done;
+    /* get the fd flags */
+    ECHECK(flags = fcntl(fd, F_GETFL));
```
jxyang, line 117: We don't record the file open modes in FD table? It seems unnecessary to use a syscall to get the modes while we are inside the kernel.

kernel/mmanutils.c
```diff
@@ -63,34 +77,28 @@ int myst_teardown_mman(void)
...
     off_t offset,
-    void* data,
-    size_t size)
+    void* addr,
+    size_t length,
+    int mmap_flags)
 {
-    ssize_t ret = -1;
+    ssize_t ret = 0;
     ssize_t bytes_read = 0;
```
jxyang, line 85: This should be some [checking](https://man7.org/linux/man-pages/man2/mmap.2.html) upfront:

> EACCES A file descriptor refers to a non-regular file.  Or a file
              mapping was requested, but fd is not open for reading.  Or
              MAP_SHARED was requested and PROT_WRITE is set, but fd is
              not open in read/write (O_RDWR) mode.  Or PROT_WRITE is
              set, but the file is append-only.

kernel/mmanutils.c
mikbras: I updated the PR to fix this. Now msync mappings cover the entire range and are reduced or split when unmappings occur.

kernel/mmanutils.c
mikbras: Fixed also.

kernel/mmanutils.c
mikbras: Fixed

kernel/mmanutils.c
mikbras: Fixed. Thanks.

mikbras: Fixed, thanks.

tests/msync/msync.c
mikbras: Removed multiple mappings of the same file feature. Created ATTN in code for syscall limitation. This feature was unstable since it did not cover the same file being mapped through different descriptors.

kernel/mmanutils.c
mikbras: I replaced **fcntl()** with a direct call to **myst_syscall_fcntl()** (the former is a wrapper for calling that latter). The fd flags are not stored in the fd-table so no other choice. The function does an O(1) lookup on the fd and calls fs->fcntl().

kernel/mmanutils.c
mikbras: Fixed. I added all the checks above.

kernel/mmanutils.c
```diff
@@ -308,3 +517,140 @@ int myst_release_process_mappings(pid_t pid)
...
+            //      maxlo     minhi
+
+            if (maxlo < minhi)
+            {
+                ECHECK(_sync_file(
+                    p->fd,                     /* fd */
+                    p->offset + (maxlo - plo), /* offset */
+                    maxlo,                     /* addr */
+                    minhi - maxlo));           /* length */
+            }
```
jxyang, line 597: Could we have two FDs mmaped to overlapping memory regions? If not, we can break out of the loop here.

kernel/mmanutils.c
```diff
@@ -184,9 +267,134 @@ void* myst_mremap(
...
 {
-    int ret = myst_mman_munmap(&_mman, addr, length);
+    int ret = 0;
+
+    /* address cannot be null and must be aligned on a page boundary */
+    if (!addr || ((uint64_t)addr % PAGE_SIZE))
+        ERAISE(-EINVAL);
+
+    /* length cannot be 0 and must be aligned on a page boundary */
+    if (!length || (length % PAGE_SIZE))
```
jxyang, line 392: I think we relaxed this. The expected behavior is to align LENGTH to a multiple of the page size. See the implementation of _munmap.

kernel/mmanutils.c
```diff
@@ -184,9 +267,134 @@ void* myst_mremap(
...
+                        ERAISE(-ENOMEM);
+                    }
+
+                    /* insert right mapping into list */
+                    rm->next = p->next;
+                    p->next = rm;
+
+                    /* update the left mapping length */
+                    p->length = llength;
+                    prev = p;
```
jxyang, line 338: We just inserted a new node after p. And `next` is the node after the new node, so `prev` should be the new node, right?

mikbras: Good catch! Fixed. I tested by hand that we hit all case but it still didn't detect this bug.

kernel/libc.c
```diff
@@ -800,6 +801,15 @@ int symlink(const char* target, const char* linkpath)
     return (int)myst_syscall_ret(myst_syscall_symlink(target, linkpath));
 }
 
+int fcntl(int fd, int cmd, ...)
```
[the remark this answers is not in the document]
mikbras: Right. gone now.

kernel/mmanutils.c
mikbras: You are right, although the man page was unclear:

```
EINVAL We don't like addr, length, or offset (e.g., they are too large, or not aligned on a page boundary).
```

I wrote this program to verify:

```
#include <sys/mman.h>
#include <stdio.h>

int main()
{
    int prot = PROT_READ | PROT_WRITE;
    int flags = MAP_PRIVATE | MAP_ANONYMOUS;
    void* addr = mmap(NULL, 4097, prot, flags, -1, 0);

    printf("addr=%p\n", addr);

    return 0;
}
```

kernel/mmanutils.c
mikbras: Yes, you can map two contiguous regions of memory, map different files to each and then, and then free them as one.

kernel/mmanutils.c
jxyang: I remember dotnet runtime does free a memory region with a length that is not aligned to page size.

jxyang (approved):

PR #134, opened 2021-02-24
Support hostfs as a rootfs

mikbras:
This change allows a hostfs to be mounted as a rootfs and extends the fs test to cover ramfs, ext2fs, and hostfs. Accordingly, it fixes buts in all three implementations.

Hostfs-rootfs support only applied to debug builds and is intended for running file system tests.

### DIFF
--- a/kernel/enter.c
+++ b/kernel/enter.c
@@ -33,6 +33,7 @@
 #include <myst/thread.h>
 #include <myst/times.h>
 #include <myst/ttydev.h>
+#include <myst/hostfs.h>
 
 #define WANT_TLS_CREDENTIAL "MYST_WANT_TLS_CREDENTIAL"
 
@@ -132,21 +133,22 @@ done:
 static int _create_standard_directories(void)
 {
     int ret = 0;
+    const mode_t mode = 0777;
 
     /* create /tmp if it does not already exist */
-    if (myst_mkdirhier("/tmp", 777) != 0)
+    if (myst_mkdirhier("/tmp", mode) != 0)
     {
         myst_eprintf("cannot create the /tmp directory\n");
         ERAISE(-EINVAL);
     }
 
-    if (myst_mkdirhier("/proc/self/fd", 777) != 0)
+    if (myst_mkdirhier("/proc/self/fd", mode) != 0)
     {
         myst_eprintf("cannot create the /proc/self/fd directory\n");
         ERAISE(-EINVAL);
     }
 
-    if (myst_mkdirhier("/usr/local/etc", 777) != 0)
+    if (myst_mkdirhier("/usr/local/etc", mode) != 0)
     {
         myst_eprintf("cannot create the /usr/local/etc directory\n");
         ERAISE(-EINVAL);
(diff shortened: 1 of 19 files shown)


kernel/enter.c
```diff
@@ -446,15 +470,26 @@ int myst_enter_kernel(myst_kernel_args_t* args)
...
+#if defined(MYST_ENABLE_HOSTFS) && defined(MYST_DEBUG)
+        /* setup and mount the HOSTFS file system */
+        if (!loaded_fs && _setup_hostfs(args->rootfs) == 0)
+        {
+            loaded_fs = true;
+        }
+#endif
+
+        if (!loaded_fs)
+            myst_panic("failed to mount rootfs: %s", args->rootfs);
```
jxyang, line 492: Maybe more informative message here? Like: `Failed to mount the rootfs %s as either RAMFS (why?), EXT2 (why?), or HOSTFS (why?)`

kernel/enter.c
```diff
@@ -446,15 +470,26 @@ int myst_enter_kernel(myst_kernel_args_t* args)
...
+        if (_setup_ext2(args->rootfs) == 0)
+        {
+            loaded_fs = true;
+        }
+#endif
 
-#else  /* MYST_ENABLE_EXT2FS */
-        myst_panic("ext2 not supported");
-#endif /* MYST_ENABLE_EXT2FS */
+#if defined(MYST_ENABLE_HOSTFS) && defined(MYST_DEBUG)
```
jxyang, line 483: There should be a comment somewhere says: `if you enable HOSTFS, additionally if you are in debug mode, then we allow the hostfs to be mounted as root fs`

kernel/enter.c
```diff
@@ -446,15 +470,26 @@ int myst_enter_kernel(myst_kernel_args_t* args)
     }
     else
     {
-#ifdef MYST_ENABLE_EXT2FS
```
jxyang, line 449: Can we move the file system mounting operations to a separate function?

kernel/enter.c
```diff
@@ -203,6 +203,30 @@ static int _setup_ext2(const char* rootfs)
...
 #endif /* MYST_ENABLE_EXT2FS */
 
+#if defined(MYST_ENABLE_HOSTFS) && defined(MYST_DEBUG)
+static int _setup_hostfs(const char* rootfs)
+{
+    int ret = 0;
+
+    if (myst_init_hostfs(&_fs) != 0)
+    {
+        myst_eprintf("failed initialize the RAM files system\n");
```
jxyang, line 213: Should the message be `failed to initialize the host files system`?

kernel/enter.c
```diff
@@ -203,6 +203,30 @@ static int _setup_ext2(const char* rootfs)
...
+        ERAISE(-EINVAL);
+    }
+
+    if (myst_mount(_fs, rootfs, "/") != 0)
+    {
+        myst_eprintf("cannot mount root file system\n");
+        ERAISE(-EINVAL);
+    }
+
+    _create_standard_directories();
```
jxyang, line 223: What's the impact of the created folders? Are they written back to the host?

kernel/enter.c
jxyang: We should use the `debug` flag in config.json. That's a more truthful reflection of whether the enclave is in debug mode or not.

tools/myst/enc/syscall.c
```diff
@@ -1336,6 +1336,25 @@ static long _fstatfs(int fd, struct myst_statfs* buf)
...
 
+#ifdef MYST_ENABLE_HOSTFS
+static off_t _lseek(int fd, off_t offset, int whence)
+{
+    long ret = 0;
+    long retval;
+
+    if (myst_lseek_ocall(&retval, fd, offset, whence))
+    {
+        ret = -EINVAL;
```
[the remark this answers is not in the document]
mikbras: This errno is returned in the **retval** parameter. This return is for the failure of the OCALL itself (which should return OE_OK). To be more explicit I added "!= OE_OK".

kernel/enter.c
mikbras: Good point. I cleaned all this up with a new function called **_get_fstype()** that determines the file system type and then we use a switch statement that is much cleaner. Also, I added error string output arguments to the setup functions so we can provide better error messages.

kernel/enter.c
mikbras: I will add to the file systems doc when read.

Good idea about the debug flag. Rather than relying on the config.json flag, I added a function to use **oe_get_report_v2()** and **oe_parse_report()** and then I check that then I check the following mask.

```
parsed_report.identity.attributes & OE_REPORT_ATTRIBUTES_DEBUG
``

kernel/enter.c
mikbras: Good idea. I moved it to **_mount_rootfs()**.

kernel/enter.c
mikbras: Thanks. Fixed.

kernel/enter.c
mikbras: They are written back to the hostfs source directory. This will vanish after we have RAMFS file systems for /tmp and /proc.

jxyang (approved):

PR #138, opened 2021-02-24
Fixes for open tests

mikbras:
Add fixes so that the following ltp open tests will pass.

```
/ltp/testcases/kernel/syscalls/open/open03
/ltp/testcases/kernel/syscalls/open/open07
/ltp/testcases/kernel/syscalls/open/open09
```

This involved the following changes:
- access checks in read and write functions (O_RDONLY/O_WRONLY)
- add umask syscall
- add chown syscall as no-op
- add setpgid syscall
- update ltp tests to use ext2 (instead of ramfs)
- fix handling of filenames ending in slash (denoting /.)

### DIFF
--- a/include/myst/process.h
+++ b/include/myst/process.h
@@ -8,6 +8,8 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#define MYST_DEFAULT_UMASK (S_IWGRP | S_IWOTH)
+
 MYST_INLINE pid_t myst_getsid(void)
 {
     return myst_thread_self()->sid;
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -3074,13 +3074,28 @@ long myst_syscall(long n, long params[6])
             BREAK(_return(n, myst_syscall_fchmod(fd, mode)));
         }
         case SYS_chown:
-            break;
+        {
+            const char* pathname = (const char*)x1;
+            uid_t owner = (uid_t)x2;
+            gid_t group = (gid_t)x3;
+
+            _strace(n, "pathname=%s owner=%u group=%u", pathname, owner, group);
+
+            /* owner is a no-op for now since kernel executes as root */
+            BREAK(_return(n, 0));
+        }
         case SYS_fchown:
             break;
         case SYS_lchown:
             break;
         case SYS_umask:
-            break;
+        {
+            mode_t mask = (mode_t)x1;
+
+            _strace(n, "mask=%o", mask);
+
+            BREAK(_return(n, myst_syscall_umask(mask)));
+        }
         case SYS_gettimeofday:
         {
             struct timeval* tv = (struct timeval*)x1;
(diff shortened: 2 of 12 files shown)


include/myst/process.h
```diff
@@ -8,6 +8,8 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#define MYST_DEFAULT_UMASK (S_IWGRP | S_IWOTH)
```
jxyang, line 11: Why do we need permission `S_IWOTH`?

kernel/umask.c
```diff
@@ -0,0 +1,21 @@
+#include <myst/syscall.h>
+#include <myst/spinlock.h>
+#include <myst/thread.h>
+#include <myst/eraise.h>
+
+long myst_syscall_umask(mode_t mask)
```
jxyang, line 6: Don't feel we should dedicate a file just for a single function. The cons are increased compile time and decreased code readability.

Maybe move this and a few other file permission related functions to a separate file?

kernel/syscall.c
```diff
@@ -3074,13 +3074,28 @@ long myst_syscall(long n, long params[6])
             BREAK(_return(n, myst_syscall_fchmod(fd, mode)));
         }
         case SYS_chown:
-            break;
+        {
+            const char* pathname = (const char*)x1;
+            uid_t owner = (uid_t)x2;
+            gid_t group = (gid_t)x3;
```
jxyang, line 3080: Is this needed for this PR?

kernel/syscall.c
```diff
@@ -3181,7 +3196,18 @@ long myst_syscall(long n, long params[6])
             BREAK(_return(n, MYST_DEFAULT_GID));
         }
         case SYS_setpgid:
-            break;
+        {
+            gid_t gid = (gid_t)x1;
```
jxyang, line 3200: Is this needed for this PR?

kernel/syscall.c
jxyang: Some other ltp tests require SYS_chown. Let's keep this.

include/myst/process.h
mikbras: This is is the "typical default value" as defined by the umask manpage.

```
 typical default value for the process  umask  is  S_IWGRP | S_IWOTH
       (octal  022).
```

kernel/umask.c
mikbras: Fixed. I moved it to kernel/fs.c (since it is file-system related).

kernel/syscall.c
mikbras: Yes, one of the LTP tests called it.

hukoyu:
@mikbras I need this PR to be merged so that I can submit PR to enable more LTP tests. Most of the LTP tests call chown syscall.

jxyang (approved):

PR #145, opened 2021-02-26
Make sample pattern easier to follow.

catblade:
Clean up sample code to be easier to follow, following this protocol:
https://medium.com/swlh/return-early-pattern-3d18a41bba8

Also, returning 1, because I'm assuming that the "return 0" is only in the case that the code actually executed the secret block.


AevaOnline (approved):
lgtm!

jxyang (approved):

AevaOnline:
Thanks for the contribution, @catblade , and congratulations on your first merged PR to Mystikos! :)

PR #149, opened 2021-03-01
Drop closed FDs from epoll internal list

jxyang:

This PR also improved the test case for epoll to having two clients making requests to the server simultaneously.


mikbras (approved):
Looks great. I see a potential bug if there are multiple threads calling wait, but that should not occur in a well behaved program.

jxyang:
Thanks Mike. I address the potential race condition with a spinlock.

PR #154, opened 2021-03-02
Add simple installation instructions to README

### LINKED ISSUE #152: missing documentation for binary installations
https://github.com/deislabs/mystikos#from-binaries has a *TODO* but no instructions yet for how to get started with a .tar.gz binary release.

AevaOnline:
This commit adds simple bash installation instructions to the README file.

Closes #152

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,24 @@ kernel while others are delegated to the target.
 You can download the latest binary release [from our builds
 page](https://github.com/deislabs/mystikos/releases)
 
-**TODO**: Include installation instructions. 
+```
+# change this to match the latest version
+LATEST='0.1.2'
+RELEASE="mystikos-${LATEST}-x86_64"
+
+# this will create the "mystikos" directory within your current working directory
+curl -sSL --ssl https://github.com/deislabs/mystikos/releases/download/v${LATEST}/${RELEASE}.tar.gz | tar -xzf -
+
+# you can use mystikos from your home directory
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,14 +46,56 @@ kernel while others are delegated to the target.
...
+```
+## Add Intel's repository & install the sgx libraries
+
+```bash
+echo 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu bionic main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
+wget -qO - https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo apt-key add -
+
+sudo apt update
+
+sudo apt -y install libsgx-enclave-common libsgx-dcap-ql libsgx-dcap-ql-dev
```
jxyang, line 78: Might be worth pointing out that for ACC VMs, az-dcap-client is a required package for remoted attestation.

jxyang (approved):

README.md
AevaOnline: Wouldn't that be better in an Azure-specific example, perhaps in the docs directory?

Other parts of this README are not cloud specific (and even call out that this can be tested on _any_ sgx-capable hardware).

jxyang: Without az-dcap-client, a bunch of tests/solutions would fail. For this reason, we prefer users of Mystikos to get an ACC VM. We can put the note here or in another document. I am fine either way.

PR #157, opened 2021-03-02
Rewrite appbuilder to improve usability

### LINKED ISSUE #153: myst-appbuilder should use random temporary file and resource names
using static resource names, such as the directory name `appdir` or the docker image name `temp-extract`, is insecure (a would-be attacker could predict the resource name and inject changes) and it is prone to conflicts if there are multiple concurrent runs on the same host.

all temporary resources used when building a myst `cpio` archive should use random identifiers, and furthermore should use signal traps to ensure they are cleaned up if the program exits early or unexpectedly.

AevaOnline:
This is a pretty hefty rewrite of 'appbuilder', including the following changes:
- introduce proper temp file handling, and stop using static file names for temp objects
- also use a temp directory for decompressing the tar file
- do not use a fixed name for the temporary docker container or docker image, instead
  using the docker auto-generated names for these
- use 'getopts' for better option handling
- improve help message output

Closes #153

### DIFF
--- a/scripts/appbuilder
+++ b/scripts/appbuilder
@@ -8,13 +8,14 @@
 function print_usage {
     echo "$(basename $0): a useful shell script for converting Docker images into Mystikos cpio archives"
     echo
-    echo "Usage: $(basename $0) [-o OUTPUT_FILENAME] [-d Dockerfile] [-i IMAGE_NAME]"
+    echo "Usage: $(basename $0) [-o file] [-d [file]] [[-i] image]"
     echo
-    echo "  -d      Build image from a Dockerfile"
-    echo "  -i      Specify an existing Docker image name to build from"
-    echo "          If neither -d nor -i are specified, will look for Dockerfile in current dir"
-    echo "  -f      Specify the output format. Options: dir, cpio. Default: 'dir'"
-    echo "  -o      Specify the file name of output file for directory"
+    echo "  -d file     Build docker image from the named file"
+    echo "              Defaults to './Dockerfile' if no option specified"
+    echo "  -i image    Specify an existing Docker image name to build from"
+    echo "              If neither -d nor -i are specified, will look for Dockerfile in current dir"
+    echo "  -f format   Specify the output format. Options: dir, cpio. Default: 'dir'"
+    echo "  -o      Specify the name of the output file or directory"
     echo "          If not specified, defaults to 'appdir'"
     echo "  -h      Print this help message"
     echo
@@ -24,7 +25,7 @@ function print_usage {
 while getopts 'hvd:i:f:o:' OPTION; do
     case "$OPTION" in
         d )
-            DOCKER_FILE="$OPTARG"
+            DOCKER_FILE=${OPTARG:-Dockerfile}
             ;;
         f )
             OUTPUT_FORMAT="$OPTARG"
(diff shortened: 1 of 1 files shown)


AevaOnline:
Looks like this is breaking tests, as expected, because I changed both the input options and output file type for the myst-appbuilder command. I'll need to update several tests as well as documentation in several places.

I find this flow for myst-appbuilder to be significant improvement because: a) it's simpler, and b) it removes an unnecessary step that leaks temporary state.

What do others think?

AevaOnline:
Updated commit to address broken tests and doc cases by preserving current default behaviour, while adding more streamlined functionality.

scripts/appbuilder
```diff
@@ -1,62 +1,151 @@
...
+function print_usage {
+    echo "$(basename $0): a useful shell script for converting Docker images into Mystikos cpio archives"
+    echo
+    echo "Usage: $(basename $0) [-o OUTPUT_FILENAME] [-d Dockerfile] [-i IMAGE_NAME]"
+    echo
+    echo "  -d      Build image from a Dockerfile"
+    echo "  -i      Specify an existing Docker image name to build from"
+    echo "          If neither -d nor -i are specified, will look for Dockerfile in current dir"
+    echo "  -f      Specify the output format. Options: dir, cpio. Default: 'dir'"
+    echo "  -o      Specify the file name of output file for directory"
```
jxyang, line 17: Should be `Specify the name of the output file or directory`

scripts/appbuilder
```diff
@@ -1,62 +1,151 @@
...
-    docker export temp-extract -o temp-extract.tar
-    sudo chmod 777 temp-extract.tar
-    docker rm temp-extract
+    # note that we have to start up the image to flatten it
+    TEMP_INAME=$(docker run -d $IMAGE_NAME)
+    docker stop $TEMP_INAME >/dev/null
+    docker export $TEMP_INAME -o $TEMP_FILE
+    docker rm $TEMP_INAME >/dev/null
+    if [ -z $DELETE_IMAGE ]; then
+	docker rmi $IMAGE_NAME
```
jxyang, line 129: Better use `-f`. I remember we had to add it to the previous version.

scripts/appbuilder
```diff
@@ -1,62 +1,151 @@
...
+    fi
 
-    rm -rf appdir && mkdir appdir
-    tar xf temp-extract.tar -C appdir
-}
+    # finally, create the output cpio object from the docker export
+    tar xf $TEMP_FILE -C $TEMP_DIR
```
jxyang, line 133: Can we take care of the name server issue at this point?

jxyang:
Please ensure the changes pass the pipeline.

scripts/appbuilder
AevaOnline: I think that can be addressed in a separate PR, as it's a separate issue.

scripts/appbuilder
AevaOnline, line 17:
```suggestion
    echo "  -o      Specify the name of the output file or directory"
```

scripts/appbuilder
AevaOnline, line 129:
```suggestion
	docker rmi -f $IMAGE_NAME
```

jxyang (approved):

PR #160, opened 2021-03-03
Support absolute symlinks across file systems

vtikoo:
### Summary
This affects file operations which deal with pathnames.
During path to inode lookup, if an absolute symlink is encountered on an
intermediate path component; path lookup is terminated early.

The unresolved path is then resolved with the mount resolve callback,
which the filesystems receive during initial setup.

With the target filesytem and the unresolved path in hand, the file
operation is then delegated to the target filesystem.

Limitations:
- Only absolute symlinks can cross filesystem boundary.
- Symlinks into hostfs are supported, but not the other way. This is
because hostfs file operations are carried by the host OS, and we can't
supplement it with the mount point information within mystikos.

### DIFF
--- a/ext2/ext2.c
+++ b/ext2/ext2.c
@@ -921,7 +921,7 @@ static int _load_file_by_path(
     if (!path || !data_out || !size_out)
         ERAISE(-EINVAL);
 
-    ECHECK(ext2_open(&ext2->base, path, O_RDONLY, 0000, &file));
+    ECHECK(ext2_open(&ext2->base, path, O_RDONLY, 0000, NULL, &file));
     ECHECK(_load_file(ext2, file, &data, &size));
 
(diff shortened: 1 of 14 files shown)


jxyang:
Can you provide background? Why do we want to symlink into hostfs?

vtikoo:
Some procfs entries are symlinks, like `/proc/self/exe` and `/proc/self/fd/[fd]`. Currently these are created within rootfs. With a separate proc filesystem, which gets mounted on the rootfs at `/proc`, these symlinks become cross file system symlinks.

Also as a result any filesystem which can be a rootfs in mystikos would atleast need to support links into it.

mikbras (approved):
This looks great Vikas! Thank you for this work.

PR #174, opened 2021-03-09
LTP: add new ltp2 test and fix related issues

mikbras:
This change add the **test/ltp2** test, enables more tests by fixing related issues, including:
- support ftruncate() with O_APPEND
- fix problems with hostfs readv/writev exposed by LTP tests
- fix error code returns to align with LTP test expectations
- add fsync syscall (as advisory only since file systems are ephemeral and cannot really be sync'd)
- ignore the mmap fd parameter when flags contain MAP_ANONYMOUS
- support fcntl F_GETFL
- support EPIPE as expected by LTP tets
- check for bad addresses passed to kernel (expected by LTP tests)
- support fallocate syscall (as no-op)


vtikoo (approved):
Looks good to me.

PR #176, opened 2021-03-10
Temporary file system support

mikbras:
Building on @vtikoo's symlink work, this change introduces temporary file systems for the following:

```
/tmp
/run
/proc
```

This change prevents temporary files from being dumped to the hostfs when used as a rootfs.

The implementation mounts a unique RAMFS for each of the mount points above.


vtikoo (approved):
LGTM.

PR #177, opened 2021-03-10
enable more ltp tests

salsal97:


mikbras (approved):
Thank you for doing this.

PR #178, opened 2021-03-10
Separate ramfs based proc filesystem

vtikoo:
### Summary
- Mounts procfs, a ramfs based filesystem, at /proc. 
- Added pid specific link entries: /proc/[pid]/fd and /proc/[pid]/exe
- /proc/self will resolve to the current pid.
- Initial implementation of /proc/meminfo.

### DIFF
--- a/kernel/ramfs.c
+++ b/kernel/ramfs.c
@@ -2186,3 +2186,57 @@ done:
 
     return ret;
 }
+
+int myst_release_tree(
+    myst_fs_t* fs,
+    const char* pathname
+)
+{
+    int ret = 0;
+    ramfs_t* ramfs = (ramfs_t*)fs;
+    inode_t *parent, *self;
+
+    if (!_ramfs_valid(ramfs))
+        ERAISE(-EINVAL);
+
+    if (!pathname)
+        ERAISE(-EINVAL);
+
+    ECHECK(_path_to_inode(ramfs, pathname, true, &parent, &self, NULL, NULL));
+
+    if (!_inode_valid(parent) || !_inode_valid(self))
+        ERAISE(-EINVAL);
+
+    /* Release all inodes in the sub-tree under self*/
+    {
+        int type, mode = self->mode;
+        if (S_ISDIR(mode))
+            type = DT_DIR;
+        else if (S_ISREG(mode))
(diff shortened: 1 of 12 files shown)


kernel/ramfs.c
```diff
@@ -738,6 +760,9 @@ static int _fs_open(
 
         if ((flags & O_APPEND))
             file->offset = inode->buf.size;
+
+        if (inode->vcallback)
+            ECHECK((*inode->vcallback)(&file->vbuf));
```
jxyang, line 765: Why don't we use `inode->buffer` to unify the getter functions?

tests/Makefile
```diff
@@ -66,9 +66,6 @@ ifndef MYST_SKIP_LIBCXX_TESTS
 DIRS += libcxx
 endif
 
-DIRS += tlscert
-DIRS += wake_and_kill
```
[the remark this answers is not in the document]
vtikoo: These seem to be duplicates of L61-62.

kernel/ramfs.c
vtikoo: Thread-safety was one reason for keeping the buffer at the file level.

kernel/thread.c
```diff
@@ -718,6 +722,18 @@ static long _syscall_clone_vfork(
         child->main.prev_process_thread = parent_main_thread;
         parent_main_thread->main.next_process_thread = child;
         myst_spin_unlock(&myst_process_list_lock);
+
+        /* Create /proc/[pid]/fd directory for new thread */
```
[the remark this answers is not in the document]
vtikoo: Good point! fixed.

kernel/ramfs.c
jxyang: This doesn't eliminate race condition still, right?

jxyang (approved):

kernel/ramfs.c
vtikoo: Yes agree :( There is still a race condition, concurrent accesses on the same file within a thread group. @mikbras and I had discussed this but couldn't find a suitable solution.
Also Linux also stores /proc virtual files under `file->private_data`.

PR #180, opened 2021-03-10
Add samples for python Flask and uWSGI

AevaOnline:
Neither sample works right now due to syscall limitations. This is documented within the sample. Uploading to share the test cases.

### DIFF
--- a/solutions/python_web_frameworks/Makefile
+++ b/solutions/python_web_frameworks/Makefile
@@ -1,0 +1,39 @@
+TOP = $(abspath ../..)
+include $(TOP)/defs.mak
+
+APPBUILDER=$(TOP)/scripts/appbuilder
+ROOTFS=hello_server.cpio
+
+ifdef STRACE
+OPTS = --strace
+endif
+
+all: app run
+
+app:
+	$(APPBUILDER) -f cpio -o $(ROOTFS) Dockerfile 
+
+run:
+	test -f server.pid && kill -9 `cat server.pid` || true
+	@echo "=== starting flask server ==="
+	@$(MYST_EXEC) $(OPTS) $(ROOTFS) /usr/local/bin/python3 /app/flask_app.py & echo $$! > server.pid
+	@sleep 10
+	curl 127.0.0.1:8000 -w "\n"
+	@echo "=== stopping flask server ==="
+	@test -f server.pid && kill -9 `cat server.pid` && rm server.pid || true
+	@echo "=== starting uwsgi server ==="
+	@$(MYST_EXEC) $(OPTS) $(ROOTFS) /usr/local/bin/uwsgi /app/uwsgi.ini & echo $$! > server.pid
+	@sleep 10
+	curl 127.0.0.1:8000 -w "\n"
+	@echo "=== stopping flask server ==="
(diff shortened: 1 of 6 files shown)


jxyang:
Could you retry with #267  merged? @AevaOnline

bodzhang:
With [#267](https://github.com/deislabs/mystikos/pull/267) merged, Flask sample test now fails due to Python 3 subprocess package [calling fork/exec](https://github.com/python/cpython/blob/76cd81d60310d65d01f9d7b48a8985d8ab89c8b4/Lib/subprocess.py#L1774). 

Python subprocess will [use POSIX_SPAWN on glibc version 2.24 and above ](https://github.com/python/cpython/blob/76cd81d60310d65d01f9d7b48a8985d8ab89c8b4/Lib/subprocess.py#L678). But currently Mytikos implementation doesn't support Ubuntu/glibc Python3 yet.

bodzhang:
With [#267](https://github.com/deislabs/mystikos/pull/267) merged, Python uWSGI server sample works.  I did need to add "protocol = http" to uwsgi.ini file.

bodzhang:
Setting Flask to not use multiple process, Flask server also works.  In flask_app.py:

`app.run(debug=True, host='0.0.0.0', port='8000', use_reloader=False)`

solutions/python_web_frameworks/Makefile
```diff
@@ -0,0 +1,26 @@
...
+ifdef STRACE
+OPTS = --strace
+endif
+
+all: app 
+
+app:
+	$(APPBUILDER) -f cpio -o hello_server.cpio Dockerfile 
+
+#run:
```
bodzhang, line 15: Replace with the following:

run:
	@echo "===start flask server"
	@$(MYST_EXEC) $(OPTS) rootfs /usr/local/bin/python3 /app/flask_app.py & echo $$! > server.pid
	@sleep 10
	curl 127.0.0.1:8000 -w "\n"
	@echo "===stop flask server"
	@test -f server.pid && kill -9 `cat server.pid` && rm server.pid || true
	@echo "===start uwsgi server"
	@$(MYST_EXEC) $(OPTS) rootfs /usr/local/bin/uwsgi /app/uwsgi.ini & echo $$! > server.pid
	@sleep 10
	curl 127.0.0.1:8000 -w "\n"
	@test -f server.pid && kill -9 `cat server.pid` && rm server.pid || true
	@echo "===stop flask server"

server-flask:
	$(MYST_EXEC) $(OPTS) rootfs /usr/local/bin/python3 /app/flask_app.py

server-uwsgi:
	$(MYST_EXEC) $(OPTS) rootfs /usr/local/bin/uwsgi /app/uwsgi.ini

bodzhang (changes requested):

AevaOnline:
Thanks for the suggestions, @bodzhang ! Incorporated and tested and seems to be passing locally!

bodzhang (approved):

bodzhang:
@AevaOnline , just realized that the PR did not add the new solution to /solutions/Makefile. Can you add the needed Makefile change?

bodzhang (approved):
Can address the needed  change in solutions directory Makefile in a separate PR.

PR #181, opened 2021-03-10
Add Rust hello-world example

rylev:
The example is similar to the C example except done in Rust. Unfortunately, I can't yet get normal Rust apps that rely on the Rust standard library to work, but this should hopefully be resolved at some point, and we can update these docs

### DIFF
--- a/doc/user-getting-started-rust.md
+++ b/doc/user-getting-started-rust.md
@@ -2,15 +2,11 @@
 
 Please see [README](../README.md) for how to install Mystikos or build it from source code.
 
-This guide assumes you have Rust installed through [`rustup`](https://rustup.rs) and that you're running on a Linux machine.
-
-You'll need to install the `x86_64-unknown-linux-musl` target which you can by running:
-
-```bash
-rustup target add x86_64-unknown-linux-musl
-```
+This guide assumes you have Rust and cargo installed (which you can do using [`rustup`](https://rustup.rs)) and that you're running on a Linux machine.
 
 # Write the program
+
+We can run normal Rust binaries inside the TEE.
 
 First, we create a new project using cargo:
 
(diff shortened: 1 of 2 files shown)


mikbras (approved):
@rylev: Thank you so much for doing this. The document clear and easy to read (and taught me a little about cargo as well). :)

I noticed this version still mentions the std lib limitation. Is that something you wanted to update here or later?

rylev:
@mikbras indeed I will update this PR with the documentation for how to get a Rust std lib binary working.

doc/user-getting-started-rust.md
```diff
@@ -0,0 +1,148 @@
...
+    "-C", "target-feature=-crt-static", # Don't statically link the musl C runtime. We'll dynamically link to it.
+]
+```
+
+## Build the program
+
+Compile the project like so:
+
+```bash
+cargo build --release --target=x86_64-unknown-linux-musl
```
jxyang, line 81: Did you try using a regular Linux target? We recently added glibc support.

doc/user-getting-started-rust.md
```diff
@@ -0,0 +1,148 @@
...
+panic = "abort" # abort on panic instead of unwinding
+lto = true # make our binary smaller and faster by performing link time optimizations
+```
+
+Lastly, we need to slightly change how Cargo builds our Rust project. We do this by making some changes in a Cargo config file which we create in our project directory at the path `.cargo/config`:
+
+```toml
+[target.'cfg(target_env = "musl")']
+rustflags = [
+    "-C", "linker=myst-gcc", # use `myst-gcc` for linking. `myst-gcc` is a wrapper for `musl-gcc` which links to MUSL libc instead of the default glibc.
```
jxyang, line 71: myst-gcc is not a complete replacement of gcc. I'd recommend to use regular gcc if possible. See my comment about glibc support below.

doc/user-getting-started-rust.md
```diff
@@ -0,0 +1,148 @@
+# Getting started with a native Rust program
+
+Please see [README](../README.md) for how to install Mystikos or build
+it from source code.
+
+Due to an issue in Rust's standard library, it is not currently possible to write Rust programs that use the standard library. See the end of this guide for more information on this issue.
```
jxyang, line 6: I'd like to work with you to enable Rust standard library in this PR.

rylev:
@mikbras @jxyang I've updated the docs to use the Rust std library and glibc. As the docs indicate there are two shared libraries that Rust binaries rely on which are not present in the TEE: `libgcc` and `ld`. In the docs I simple suggest the work around of copying them into the `appdir` folder.

AevaOnline:
This is great! Thanks, @rylev !

PR #182, opened 2021-03-11
Fix pipeline regression

mikbras:
This patch fixes the pipeline break by backing out the tmpfs feature (through conditional compilation).

Also these two LTP tests are broken and removed for now.
```
/ltp/testcases/kernel/syscalls/epoll/epoll-ltp
/ltp/testcases/kernel/syscalls/syslog/syslogtst
```

### DIFF
--- a/kernel/enter.c
+++ b/kernel/enter.c
@@ -130,9 +130,12 @@ done:
     return ret;
 }
 
+#ifdef USE_TMPFS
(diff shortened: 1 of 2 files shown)


vtikoo (approved):
LGTM.
Should we have an item to track to remove the `USE_TMPFS` compile flag?

PR #185, opened 2021-03-11
Enable LTP testing of ramfs

mikbras:
This change enables LTP testing of ramfs (see tests/ltp2). The most notable change is a fix to the faulty inode link counting in ramfs. It adds an inode count to ensure that all inodes are freed.


salsal97 (approved):
LGTM

PR #190, opened 2021-03-12
Abandon use of OE heap

mikbras:
This is a quick POC to eliminate use of the OE heap. Instead, the kernel now uses its own internal memory manager to obtain memory. This change adds dlmalloc to the kernel and handles all dlmalloc's with mmap/mremap/munmap to and leaves the sbrk region for use by the user application.

The OE heap size was decreased from 128mb to 16kb. The kernel boot time is now 3 times faster than before.


mikbras:
Created new PR and closed this one.

PR #192, opened 2021-03-15
Remove dependency on OE heap

mikbras:
This PR adds an allocator (dlmalloc) to the Mystikos kernel and removes the dependency on the OE heap. This eliminates fragmentation caused by splitting the memory space into **user-memory** and **kernel-memory**.

This change cuts the kernel boot time by three.

Notes: the dlmalloc is configured to use mmap'd memory only (no sbrk). This leaves the application to control the brk region.


jxyang:
https://github.com/deislabs/mystikos/issues/193

mikbras:
I'm looking for a better option name than --ram=<size>. What about --heap-size=<size>?

jxyang:
The memory is used for both stack and heap of user threads (pthreads). How about `--memory-size`?

include/myst/tcall.h
```diff
@@ -18,8 +18,6 @@
 typedef enum myst_tcall_number
 {
     MYST_TCALL_RANDOM = 2048,
-    MYST_TCALL_ALLOCATE = 2049,
-    MYST_TCALL_DEALLOCATE = 2050,
```
jxyang, line 22: Should we shift the following syscall numbers upward so there is no "holes" in them?

kernel/malloc.c
```diff
@@ -0,0 +1,301 @@
+#include <limits.h>
```
jxyang, line 1: Copyright header is missing.

Which version of dlmalloc is this? Also should we pay tribute to the dlmalloc author?

solutions/aspnet/Makefile
```diff
@@ -8,6 +8,8 @@ APPNAME=webapp
 #OPTS += --strace
 #endif
 
+OPTS = --ram=256m
```
jxyang, line 11: This setting should be shadowed by the one in config.json. Does it make a difference for you?

solutions/dotnet/Makefile
```diff
@@ -8,6 +8,8 @@ APPNAME=HelloWorld
 #OPTS += --strace
 #endif
 
+OPTS = --ram=256m
```
jxyang, line 11: Same question as aspnet.

solutions/python_webserver/Makefile
```diff
@@ -18,7 +20,7 @@ rootfs:
 run:
 	test -f server.pid && kill -9 `cat server.pid` || true
 	$(MYST_EXEC) $(OPTS) rootfs /usr/local/bin/python3 /app/hello_server.py & echo $$! > server.pid
-	sleep 10
+	sleep 20
```
jxyang, line 23: Maybe some performance regressions. It used to launch within 10 seconds.

tools/myst/host/dump.c
```diff
@@ -74,10 +74,6 @@ static int dump_enclave_properties(oe_sgx_enclave_properties_t* properties)
 
     printf("xfrm=%#016" PRIx64 "\n", properties->config.xfrm);
 
-    printf(
-        "num_heap_pages=%" PRIu64 "\n",
-        properties->header.size_settings.num_heap_pages);
```
jxyang, line 79: Why don't we print heap pages reserved for OE? It's fixed at 32 pages, right?

tools/myst/host/exec.c
```diff
@@ -204,31 +185,24 @@ int exec_launch_enclave(
     return retval;
 }
 
-#define USAGE_EXEC_SGX \
-    "\
+#define USAGE_EXEC_SGX "\n\
 \n\
-Usage: %s exec-sgx <rootfs> [options] <application> <app_args...>\n\
+Usage: %s exec-linux [options] <rootfs> <application> <args...>\n\
```
jxyang, line 190: This usage string is for SGX target, isn't it?

tools/myst/host/exec.c
```diff
@@ -204,31 +185,24 @@ int exec_launch_enclave(
...
 Where:\n\
-    exec-sgx      -- execute the specified <application> from within the\n\
-                     <rootfs> with the <app_arguments> within an SGX enclave\n\
-    <rootfs>      -- This is the CPIO archive (created via mkcpio) of the\n\
-                     application directory\n\
-    <application> -- the application path from within <rootfs> to run within\n\
-                     the SGX enclave\n\
-    <app_args>    -- the application arguments to pass through to\n\
-                     <application>\n\
+    exec-linux    -- execute an application within <rootfs> in a trusted\n\
```
jxyang, line 193: exec-linux -> exec-sgx

tools/myst/host/exec.c
```diff
@@ -204,31 +185,24 @@ int exec_launch_enclave(
...
+    <app_args>    -- arguments to passed through to the <application>\n\
 \n\
-and [options] are one or more of:\n\
-    --help                          -- this message\n\
-    --user-mem-size <size>          -- for running an unsigned binary this overrides the\n\
-                                       default user memory size for an application.\n\
-                                       The <size> format is a number that is in\n\
-                                       bytes (<size>), in kilobytes (<size>k), \n\
-                                       in megabytes (<size>m), or gigabytes (<size>g)\n\
-    --app-config-path <config.json> -- specifies the configuration json file for running an\n\
```
jxyang, line 229: Any reason we want to get rid of `--app-config-path`?

include/myst/tcall.h
mikbras: Done. Thanks.

kernel/malloc.c
mikbras: Fixed. I added comments that give tribute and point to the location of the license.

mikbras: Version is 2.8.6. Same as OE.

solutions/aspnet/Makefile
mikbras: Good catch. I change the name in the config and docs to HeapSize but continued to support for UserMemSize for backwards compatibility.

solutions/dotnet/Makefile
mikbras: Fixed

solutions/python_webserver/Makefile
mikbras: I increased this because it fails on my machine, which is slower. It takes about 17 seconds to launch on my machine.

tools/myst/host/dump.c
mikbras: Okay, I added it back but it should be transparent to the user.

tools/myst/host/exec.c
mikbras: Fixed, thanks.

tools/myst/host/exec.c
mikbras: Fixed, thanks.

tools/myst/host/exec.c
mikbras: Oh my! Thanks for finding this. It was a cut-n-paste error.

mikbras:
> The memory is used for both stack and heap of user threads (pthreads). How about `--memory-size`?

perfect!

jxyang (approved):

jxyang:
Thanks for the optimization. Looks great!

PR #201, opened 2021-03-16

salsal97:

Added passing test for each filesystem into the suite, with more information grouping failing tests based on their reason for failure. Hanging tests are in a separate file to allow the script to run when we want to update these details. 

To update, change the $FS var in the run script and run it.


salsal97:
closing duplicate as #206

PR #202, opened 2021-03-16
Add set and get thread name prctl options

vtikoo:

### DIFF
--- a/include/myst/thread.h
+++ b/include/myst/thread.h
@@ -187,6 +187,9 @@ struct myst_thread
     myst_spinlock_t* thread_lock;
     struct myst_thread* group_prev;
     struct myst_thread* group_next;
+
+    /* thread name */
+    char name[16];
 };
 
 MYST_INLINE bool myst_valid_thread(const myst_thread_t* thread)
@@ -294,5 +297,12 @@ int myst_get_num_threads(void);
 myst_thread_t* myst_find_thread(int tid);
 
 size_t myst_kill_thread_group();
+
+MYST_INLINE char* myst_get_thread_name(myst_thread_t* thread)
+{
+    return thread->name;
+}
+
+int myst_set_thread_name(myst_thread_t* thread, const char* n);
 
 #endif /* _MYST_THREAD_H */
--- a/kernel/thread.c
+++ b/kernel/thread.c
@@ -605,7 +605,14 @@ static long _syscall_clone(
         child->crt_td = newtls;
         child->run_thread = myst_run_thread;
         child->thread_lock = parent->thread_lock;
-
+        {
+            char buf[BUFSIZ];
+            /* ATTN: we don't take a lock on _num_threads,
+             thread names could be duplicates */
+            snprintf(buf, sizeof(buf), "thread-%ld", _num_threads);
+            myst_set_thread_name(child, buf);
+        }
+        
         // Link up parent, child, and the previous head child of the parent,
         // if there is one, in the same thread group.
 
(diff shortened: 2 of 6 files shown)


include/myst/thread.h
```diff
@@ -187,6 +187,9 @@ struct myst_thread
     myst_spinlock_t* thread_lock;
     struct myst_thread* group_prev;
     struct myst_thread* group_next;
+
+    /* thread name */
+    char name[16];
```
jxyang, line 192: Is this what Linux expected?

kernel/thread.c
```diff
@@ -680,6 +687,13 @@ static long _syscall_clone_vfork(
         child->run_thread = myst_run_thread;
         child->main.thread_group_lock = MYST_SPINLOCK_INITIALIZER;
         child->thread_lock = &child->main.thread_group_lock;
+        {
+            char buf[BUFSIZ];
+            /* ATTN: we don't take a lock on _num_threads,
+             thread names could be duplicates */
+            snprintf(buf, sizeof(buf), "thread-%ld", _num_threads);
+            myst_set_thread_name(child, buf);
```
jxyang, line 695: Why don't we just sprintf into the myst_thread_t name buffer?

kernel/thread.c
```diff
@@ -605,7 +605,14 @@ static long _syscall_clone(
         child->crt_td = newtls;
         child->run_thread = myst_run_thread;
         child->thread_lock = parent->thread_lock;
-
+        {
+            char buf[BUFSIZ];
+            /* ATTN: we don't take a lock on _num_threads,
+             thread names could be duplicates */
+            snprintf(buf, sizeof(buf), "thread-%ld", _num_threads);
+            myst_set_thread_name(child, buf);
```
jxyang, line 613: Same as below.

kernel/thread.c
```diff
@@ -813,3 +827,23 @@ size_t myst_kill_thread_group()
...
+{
+    int ret = 0;
+
+    if (!thread || !n)
+        ERAISE(-EINVAL);
+
+    /* Copy atmost 15 bytes */
+    strncpy(thread->name, n, sizeof(thread->name) - 1);
+    /* Set null terminator if string pointed by n
+        longer than the 16-byte fixed buffer for thread->name */
```
jxyang, line 841: Is this the same behavior as Linux?

include/myst/thread.h
vtikoo: Yes. 

https://elixir.bootlin.com/linux/latest/source/include/linux/sched.h#L957
https://elixir.bootlin.com/linux/latest/source/include/linux/sched.h#L215

jxyang:
Can you check if the document on syscall limitations should be updated?

kernel/thread.c
vtikoo: Yes. According to the man prctl page - the name can be atmost 16 bytes, including the null termination character.

kernel/thread.c
vtikoo: Probably a theoritical concern, but `_num_threads` being a unsigned long int, could cause the thread name to go over 16 bytes. So this was delegating to `myst_set_thread_name`, which does the bounds checking anyways.

jxyang: We can check `_num_threads` here. Checking an integer is always faster than checking a string.

vtikoo: Changed to not use `myst_set_thread_name`.

kernel/thread.c
```diff
@@ -680,6 +687,13 @@ static long _syscall_clone_vfork(
         child->run_thread = myst_run_thread;
         child->main.thread_group_lock = MYST_SPINLOCK_INITIALIZER;
         child->thread_lock = &child->main.thread_group_lock;
+        {
+            char buf[16];
+            /* ATTN: we don't take a lock on _num_threads,
+             thread names could be duplicates */
+            snprintf(buf, sizeof(buf), "thread-%ld", _num_threads % 99999999);
```
jxyang, line 694: Actually, I meant to replace the whole block with one line:
```
snprintf(child->name, sizeof(buf), "thread-%ld", _num_threads % 99999999);
```

vtikoo: Oh right, thats much simpler 👍

vtikoo: Fixed.

jxyang (approved):

PR #203, opened 2021-03-17
Use gcompat for a more compatible layer with glibc

jxyang:


third_party/gcompat/Makefile
```diff
@@ -0,0 +1,13 @@
+URL=https://github.com/AdelieLinux/gcompat.git
+
+TAG=1.0.0
+
+all: src
+
+clean: 
+	rm -rf gcompat
+
+src:
```
mikbras: This fails when I type make the second time. I renamed src to gcompat and pushed (this prevents it from trying to clone again if you type make the second time). 

I also added the **distclean** target, which is required if you type "make distclean" in the directory above.

mikbras (approved):
Looks great. I pushed changes to keep make from failing the second time in the third_party/gcompat directory.

I also added a "distclean" target to that makefile.

third_party/gcompat/Makefile
jxyang: Thank you Mike.

PR #206, opened 2021-03-18

salsal97:

- Added testing for 3 filesystems - hostfs, ext2fs and ramfs
- Added passing test for each filesystem into the suite
- added more information grouping failing tests based on if the syscall is missing or it was another error

To update, change the $FS var in the run script and run it.


tests/Makefile
```diff
@@ -76,6 +76,7 @@ endif
 ifdef MYST_NIGHTLY_TEST
 DIRS += ltp
 DIRS += sockperf
+DIRS += ltp2
```
mikbras, line 79: Do you plan to consolidate ltp2 and ltp into one test?

tests/ltp2/Makefile
```diff
@@ -133,5 +141,3 @@ endif
 
 t:
```
mikbras: I removed this rule and pushed.

mikbras (changes requested):
Thank you Saloni for doing this. It looks like there might be an unused directory. Also I was wondering whether you plan to consolidate the ltp and ltp2 tests (maybe in a later PR)?

tests/fs_ltp/fstests
```diff
@@ -0,0 +1,522 @@
+/ltp/testcases/kernel/syscalls/access/access01
```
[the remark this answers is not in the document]
salsal97: Hi Mike, I'm working to separate the ltp tests that have fs syscalls and run them in the PR pipeline to give better exposure. I pushed this file as an effort towards that, it is still WIP

tests/Makefile
salsal97: Yes, once I separate out the fstests I will be doing that as well

mikbras (approved):

PR #212, opened 2021-03-19

salsal97:

- Added testing for 3 filesystems - hostfs, ext2fs and ramfs
- Added passing test for each filesystem into the suite
- added more information grouping failing tests based on if the syscall is missing or it was another error
- runs fstests in PR pipeline and all the LTP syscall tests in the nightly pipeline

To update, run the run_* script


mikbras (approved):
Thank you so much for doing this Saloni.

Issue #217, opened 2021-03-22
appbuilder should be able to take extra "docker build" options


jxyang:
Some docker images require command line arguments to build, as detailed [here](https://vsupalov.com/docker-build-pass-environment-variables/). `appbuilder` as a docker drop-in replacement should support them as well.

Propose: `appbuilder [-e [extra build options]]`

jxyang:
Fixed by #226

PR #226, opened 2021-03-29
Fix appbuilder issues

### LINKED ISSUE #216: appbuilder doesn't work with -d
```bash
appbuilder  -d ./Dockerfile
unable to prepare context: context must be a directory: ...
```

jxyang:
1) Interpret "-d" option correctly;
2) Add "-e" option for extra options to the container image build command;
3) Generate name server on appdir and remove the now redundant commands in Makefiles;
4) Produce more meaningful help messages for appbuilder.

Fix #216 #217

### DIFF
--- a/scripts/appbuilder
+++ b/scripts/appbuilder
@@ -8,21 +8,23 @@
 function print_usage {
     echo "$(basename $0): a useful shell script for converting Docker images into Mystikos cpio archives"
     echo
-    echo "Usage: $(basename $0) [-o file] [-d [file]] [[-i] image]"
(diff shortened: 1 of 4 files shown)


praenubilus:
can we add extra param to turn off container build silent mode?

jxyang:
> can we add extra param to turn off container build silent mode?

You can use `-v` to force appbuilder to print more messages. In terms of `docker build`, I haven't find a way to turn off `-q` but still get the image name cleanly.

vtikoo (approved):

Francis-Liu (approved):
LGTM

PR #227, opened 2021-03-30
updates ltp test roster

salsal97:


jxyang:
Thanks for keeping this updated!

jxyang:
Seems there are some consistent pipeline failures. Is ltp/run_tests_locally.sh  used by the pipeline?

PR #235, opened 2021-04-01
Print friendly ip address in strace

jxyang:

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -74,6 +74,8 @@
 
 #define DEV_URANDOM_FD MYST_FDTABLE_SIZE
 
+#define MAX_IPADDR_LEN 64
+
 #define COLOR_RED "\e[31m"
 #define COLOR_BLUE "\e[34m"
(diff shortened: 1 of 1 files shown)


kernel/syscall.c
```diff
@@ -600,6 +602,29 @@ static long _return(long n, long ret)
     return ret;
 }
 
+static int _socketaddr_to_str(
+    const struct sockaddr* addr,
+    char out[],
+    size_t limit)
+{
+    int ret = 0;
+    const uint8_t* p = (uint8_t*)addr->sa_data;
```
mikbras, line 611: Maybe add a null check for the addr parameter. Otherwise the related syscalls will crash when passed null address parameters.

mikbras (changes requested):

kernel/syscall.c
jxyang: Good catch. Changed.

mikbras (approved):
This will be nice to have. Thank you for doing this.

PR #239, opened 2021-04-04
Increase max threads from 32 to 1024

mikbras:
This change increases the maximum thread count from 32 to 1024, requiring the following changes:
- Increase OE_SGX_MAX_TCS from  32 to 1024
- Use tiny OE stacks to allow 1024 threads by default
- Switch from OE TCS stacks to dynamically allocated stacks.


kernel/thread.c
```diff
@@ -567,6 +575,42 @@ long myst_run_thread(uint64_t cookie, uint64_t event)
...
+
+    /* get the thread corresponding to this cookie */
+    if (!(thread = _put_cookie(cookie)))
+        ERAISE(-EINVAL);
+
+    /* the stack size is determined by the thread type (process or regular) */
+    if (myst_is_process_thread(thread))
+        stack_size = process_stack_size;
+    else
+        stack_size = regular_stack_size;
```
jxyang, line 596: Why create a temporary stack for regular threads? They are going to switch to a new one on line 561 anyway.

tools/myst/enc/enc.c
```diff
@@ -628,6 +633,40 @@ int myst_enter_ecall(
     return ret;
 }
 
+#define ENTER_STACK_SIZE (132 * 1024)
```
jxyang, line 636: Need a comment here, like "This is the stack size of the main thread."

include/myst/thread.h
```diff
@@ -305,4 +305,7 @@ MYST_INLINE char* myst_get_thread_name(myst_thread_t* thread)
 
 int myst_set_thread_name(myst_thread_t* thread, const char* n);
 
+/* execute a function on the given stack */
+long myst_switch_stack(void* stack, long (*func)(void* arg), void* arg);
```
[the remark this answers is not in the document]
mikbras: Good point. To keep it short, I went with ``myst_call_on_stack``.

kernel/thread.c
mikbras: I wanted to do that but there may not be enough room on the OE stack for ``__run_thread`` to execute. The OE stack is only two-pages and some of that is already used by edger8r stubs and the call frame into myst_run_thread(). Also ``_run_thread`` calls other functions as well that may use up stack space. Even if there is enough space now, we would have to keep a close eye on the evolution of the ``_run_thread`` function to guarantee that it never consumes more than the remainder of the OE heap (less than two pages). Worse, we don't know how much stack is left.

tools/myst/enc/enc.c
mikbras: Good idea. Done.

jxyang (approved):

PR #241, opened 2021-04-05
Syscall instruction exception handler, with low level plumbing and test codes from @mikbras

bodzhang:


tools/myst/enc/enc.c
```diff
@@ -37,6 +37,13 @@
 #define IRETFRAME_EFlags IRETFRAME_SegCs + 8
 #define IRETFRAME_Rsp IRETFRAME_EFlags + 8
 
+static myst_kernel_args_t kargs;
+
+long myst_syscall(long n, long params[6])
```
jxyang, line 42: This syscall is unique in that the caller is from a layer below the kernel. Suggest rename this function to differentiate from the other `myst_syscall` that is originated from libc.

tools/myst/enc/enc.c
```diff
@@ -112,6 +121,37 @@ static uint64_t _vectored_handler(oe_exception_record_t* er)
 
         return OE_EXCEPTION_CONTINUE_EXECUTION;
     }
+    if (er->code == OE_EXCEPTION_ILLEGAL_INSTRUCTION &&
```
jxyang, line 124: Can we reorg the branches now so it's more efficient? Like:
```c
if (er->code == OE_EXCEPTION_ILLEGAL_INSTRUCTION)
{
   switch (opcode)
   {
       ....
```

jxyang (approved):

PR #246, opened 2021-04-06
Add a simple OpenMP sample app

bodzhang:


samples/openmp/Makefile
```diff
@@ -0,0 +1,28 @@
...
+all: appdir rootfs
+
+appdir:
+	$(APPBUILDER) Dockerfile
+
+rootfs: appdir
+	sudo $(MYST) mkext2 appdir ext2rootfs
+	sudo chown $(USER).$(USER) ext2rootfs
+	$(MYST) fssig --roothash ext2rootfs > roothash
+	# $(MYST) mkcpio appdir rootfs
```
jxyang, line 19: Why don't we use CPIO archive?

bodzhang: I prefer using EXT2 rootfs to reduce EPC consumption.

jxyang: For this size of applications, EXT2fs and ramfs probably don't have much performance difference.

bodzhang: @jxyang , any specific concern about using EXT2fs instead of CPIO for this sample?

jxyang: Just curious because ext2fs requires more step in the makefile and uses one more `myst` argument.  But I can live with that if you prefer.

jxyang (approved):

PR #252, opened 2021-04-07
remove mbedtls_debug_set_threshold to run the release pipeline solutions

salsal97:


mikbras:
If this passes the pipeline, then a permanent fix would be to define NDEBUG when building the enclave. For example, you can the following to ``solutions/attested_tls/oe_enclave/enc/Makefile``:

```
ifeq ($(MYST_RELEASE),1)
    CFLAGS += -DNDEBUG
endif
```

jxyang:
Could you address Mike's comment? @salsal97

salsal97:
> Could you address Mike's comment? @salsal97

Hi! We tried it out offline and the problem still persists with the solution Mike mentioned since we are not building with the release flag (on purpose), so we concluded that the commenting out would be the way to go. @jxyang

jxyang (approved):

jxyang:
Sounds good. Please resolve the merge conflicts before merging.

PR #263, opened 2021-04-08
Run make clean to get rid of root-owned files

jxyang:


jxyang:
Incorporated in #261

PR #265, opened 2021-04-08
Debug malloc feature

mikbras:
This change adopts the debug-malloc feature from Open Enclave. This debug allocator checks for three kinds of errors:
- leak detection on shutdown
- use of uninitialzed memory from malloc (memory is filled with 0xaa)
- use of freed memory (memory is filled with 0xdd on free)

To enable pass the ``--memcheck`` option to ``myst``

### DIFF
--- a/include/myst/debugmalloc.h
+++ b/include/myst/debugmalloc.h
@@ -1,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#ifndef _MYST_DEBUGMALLOC_H
+#define _MYST_DEBUGMALLOC_H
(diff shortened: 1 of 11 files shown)


jxyang:
I noticed with debug malloc, the pipeline time increased ~50%, from 25 times to 38 minutes. Is there a way we can improve the performance?

mikbras:
> I noticed with debug malloc, the pipeline time increased ~50%, from 25 minutes to 38. Is there a way we can improve the performance? Or make it non-default?

It is off by default. I can't see any reason this would increase the pipeline.

paulcallen (approved):

PR #267, opened 2021-04-08
Robust list feature

mikbras:
Implements ``SYS_sched_get_robust_list`` and ``SYS_sched_get_robust_list`` system calls.

### DIFF
--- a/include/myst/thread.h
+++ b/include/myst/thread.h
@@ -57,6 +57,13 @@ typedef struct
     uint64_t mask;
 } posix_sigaction_t;
 
+typedef struct myst_robust_list_head
+{
+    volatile void* volatile head;
+    long off;
+    volatile void* volatile pending;
+} myst_robust_list_head_t;
+
 struct myst_thread
 {
     /* MYST_THREAD_MAGIC */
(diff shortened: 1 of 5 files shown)


include/myst/thread.h
```diff
@@ -57,6 +57,13 @@ typedef struct
     uint64_t mask;
 } posix_sigaction_t;
 
+typedef struct myst_robust_list_head
+{
+    volatile void* volatile head;
+    long off;
+    volatile void* volatile pending;
+} myst_robust_list_head_t;
```
jxyang, line 65: I am not sure we need this structure. https://man7.org/linux/man-pages/man2/set_robust_list.2.html states:
> These system calls deal with per-thread robust futex lists.
       These lists are managed in user space: the kernel knows only
       about the location of the head of the list.

It's up to the user space, in our case, MUSL, to interpret the robust list. The kernel merely records where is the head of the list, and the size of each item in the list. We can use `void*` to represent the list, and `size_t` to represent the size.

include/myst/thread.h
```diff
@@ -189,6 +196,11 @@ struct myst_thread
     struct myst_thread* group_prev;
     struct myst_thread* group_next;
 
+    /* robust list (see SYS_set_robust_list & SYS_get_robust_list) */
+    struct myst_robust_list_head* robust_list_head;
+    size_t robust_list_len;
```
jxyang, line 201: See comments above.

kernel/syscall.c
```diff
@@ -2313,6 +2313,53 @@ long myst_syscall_futimesat(
     return ret;
 }
 
+long myst_syscall_get_robust_list(
+    int pid,
+    myst_robust_list_head_t** head_ptr,
```
jxyang, line 2318: See comments above.

jxyang (approved):

PR #268, opened 2021-04-08
mkext2 now creates sudo-user owned files

mikbras:
The ``sudo myst mkext2`` command previously created files owned by root. This changes that command to create files owned by the sudo user (the user that executed sudo). The sudo program passes the following environment variables:

- ``$SUDO_USER``
- ``$SUDO_UID``
- ``$SUDO_GID``

When the latter two are defined, ``myst mkext2`` changes the owner of output files and directories to $SUDO_UID/$SUDO_GID.

### DIFF
--- a/host/file.c
+++ b/host/file.c
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #include <fcntl.h>
+#include <stdlib.h>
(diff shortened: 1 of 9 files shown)


jxyang:
LGTM

jxyang (approved):

PR #270, opened 2021-04-08
add --max-threads option

mikbras:
This PR adds the ``--max-threads`` option for ``exec-sgx`` only. If this option is specified, it creates a temporary copy of the enclave image file and patches the ``num_tcs`` in the enclave properties.

This option is aimed at speeding up the pipeline. It will be used to optimize the load time for testing purposes. For tests that only require 2 threads (instead of 1024), myst loads 2.3 times faster.


mikbras:
It is taking a very long time to run the tests for unknown reason. Looking in to it now.

PR #272, opened 2021-04-09
removed failed tests that fail outside of mystikos

salsal97:


jxyang:
Do we know why they are failing outside of Mystikos? Is it because of MUSL/Alpine?

salsal97:
> Do we know why they are failing outside of Mystikos? Is it because of MUSL/Alpine?

Hey, yes it seems they are because of the environment rather

jxyang:
Do we still want this? Please rebase and run pipeline again.

PR #280, opened 2021-04-13
fix Dockerfile name not recognized issue

### LINKED ISSUE #305: appbuilder should output the progress while building the container image
The current implementation prints nothing while building the image with `docker build`, leaving the user in dark and wonder whether the tool is hanging or not. This is bad user experience. It's also less helpful with debugging pipeline failures.

praenubilus:
Update the default build behavior. Only Dockerfile name will be accepted as default argument. 
Fix #305 
Add Sample Usage in help info

### DIFF
--- a/scripts/appbuilder
+++ b/scripts/appbuilder
@@ -8,18 +8,31 @@
 function print_usage {
     echo "$(basename $0): a useful shell script for converting Docker images into Mystikos cpio archives"
     echo
-    echo "Usage: $(basename $0) [-o file] [-d [file]] [[-i] image] [-e extras] [-v]"
+    echo "Usage: $(basename $0) [-o file] [-i image] [-e extras] [-v] [[-d] file]"
     echo
     echo "  -d file     Build docker image from the named file"
-    echo "              Defaults to './Dockerfile' if no option specified"
     echo "  -i image    Specify an existing Docker image name to build from"
-    echo "              If neither -d nor -i are specified, will look for Dockerfile in current dir"
     echo "  -f format   Specify the output format. Options: dir, cpio. Default: 'dir'"
     echo "  -e extras   Specify the extra options while building the container image"
     echo "  -o      Specify the name of the output file or directory"
     echo "          If not specified, defaults to 'appdir'"
     echo "  -v      Verbose"
     echo "  -h      Print this help message"
+    echo
+    echo
+    echo "Sample Usage:"
+    echo
+    echo "  Build a Dockerfile without '-d' or '-o' option:"
+    echo "    $(basename $0) Dockerfile"
+    echo
+    echo "  Build Dockerfile with irregular file name"
+    echo "    $(basename $0) Dockerfile.release"
+    echo
+    echo "  Build existing image from remote registry"
+    echo "    $(basename $0) -i hello-world"
+    echo
+    echo "  Build specified Docker image with extra options in verbose"
+    echo "    $(basename $0) -d Dockerfile.debug -v -e '--build-arg USERNAME=username --build-arg PASSWORD=password'"
     echo
 }
 
@@ -47,6 +60,7 @@ while getopts 'hvd:i:f:o:e:' OPTION; do
             ;;
         v )
             VERBOSE=true;
+            QUIET=false;
             set -x
             ;;
         * )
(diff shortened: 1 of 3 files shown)


scripts/appbuilder
```diff
@@ -8,12 +8,10 @@
 function print_usage {
     echo "$(basename $0): a useful shell script for converting Docker images into Mystikos cpio archives"
     echo
-    echo "Usage: $(basename $0) [-o file] [-d [file]] [[-i] image] [-e extras] [-v]"
+    echo "Usage: $(basename $0) [-o file] [-i image] [-e extras] [-v] [[-d] file]"
```
jxyang, line 11: Some people don't like to be forced to place an argument at a particular location.

scripts/appbuilder
```diff
@@ -8,12 +8,10 @@
...
     echo
     echo "  -d file     Build docker image from the named file"
-    echo "              Defaults to './Dockerfile' if no option specified"
```
jxyang, line 14: Why remove it?

scripts/appbuilder
```diff
@@ -8,12 +8,10 @@
...
     echo "  -i image    Specify an existing Docker image name to build from"
-    echo "              If neither -d nor -i are specified, will look for Dockerfile in current dir"
```
jxyang, line 16: Why remove it?

scripts/appbuilder
```diff
@@ -58,9 +56,16 @@ done
 
 ### defaults
 
-# if no options specified, assume first argument is an image name
-shift $((OPTIND-1))
-IMAGE_NAME=${IMAGE_NAME:-$@}
+if [ $# -lt 1 ]; then
+        echo "Dockerfile or ImageName has not been specified"
```
jxyang, line 60: Does this mean we can't run `appbuilder` byitself anymore?

scripts/appbuilder
```diff
@@ -8,12 +8,10 @@
...
-    echo "Usage: $(basename $0) [-o file] [-d [file]] [[-i] image] [-e extras] [-v]"
+    echo "Usage: $(basename $0) [-o file] [-i image] [-e extras] [-v] [[-d] file]"
     echo
     echo "  -d file     Build docker image from the named file"
-    echo "              Defaults to './Dockerfile' if no option specified"
     echo "  -i image    Specify an existing Docker image name to build from"
-    echo "              If neither -d nor -i are specified, will look for Dockerfile in current dir"
     echo "  -f format   Specify the output format. Options: dir, cpio. Default: 'dir'"
     echo "  -e extras   Specify the extra options while building the container image"
     echo "  -o      Specify the name of the output file or directory"
```
jxyang, line 17: Please provide a few example usages here.

scripts/appbuilder
praenubilus: the location matters only for the default situation(no other options like -o -i )

scripts/appbuilder
praenubilus: As we discussed, a docker file name is mandatory for default situation. 
```bash
appbulder <dockerfile-name>
```

scripts/appbuilder
praenubilus: see comments above

scripts/appbuilder
praenubilus: Yes. And this situation(run appbuilder by itself) does not happen to any tests or samples in current repo.

scripts/appbuilder
praenubilus: That depends on the definition for the appbuilder args. After the definition(sequence, default option) has been confirmed, I will add examples.

jxyang:
LGTM. Please add sample commands.

Do you want to address #305 in this PR?

praenubilus:
> LGTM. Please add sample commands.
> 
> Do you want to address #305 in this PR?

Sure. Let me include #305 fix in next commit.

jxyang (approved):

jxyang:
The pipeline failure is most likely caused by this line: https://github.com/deislabs/mystikos/blob/e9db8133479baef87c75fe383400fdf66c11d643/tests/libcxx/Makefile#L31

Can you remove `$(MAKE) clean` and try again?

praenubilus:
> The pipeline failure is most likely caused by this line:
> 
> https://github.com/deislabs/mystikos/blob/e9db8133479baef87c75fe383400fdf66c11d643/tests/libcxx/Makefile#L31
> 
> Can you remove `$(MAKE) clean` and try again?

The root cause is in this line, it will break the new enforcement. The `-i` option is needed. I have fixed it with a new commit.

https://github.com/deislabs/mystikos/blob/e9db8133479baef87c75fe383400fdf66c11d643/tests/libcxx/Makefile#L49

PR #285, opened 2021-04-13
Changing the agent pool for pipelines

jxyang:


jxyang:
Drop the plan as the original agent pool is revived.

Issue #87, opened 2021-02-04
add mount parameters to appconfig


mikbras:
Mount parameters will be needed in the **appconfig** file so that the Mystikos kernel may automatically mount these file systems during startup. These mount parameters include (see `man 2 mount`):
- source
- target
- filesystemtype
- mountflags

paulcallen:
Now have auto-mounts of all three filesystems. ramfs filesystems need to be located in the rootfs.
No flags are currently supported for the mount syscall.
The signing and encryption options are not currently implemented as we need to refactor the ext2 filesystem to handle the flags better

PR #289, opened 2021-04-15
DESIGN: Allow automatic mounting of filesystems into the TEE at start

paulcallen:
This design document explains how to achieve this.
Issue for this: #87

### DIFF
--- a/doc/design/mount-config-design.md
+++ b/doc/design/mount-config-design.md
@@ -1,0 +1,115 @@
+# Automatic filesystem mount configuration design
+
+During startup we need the ability to mount directories from the insecure host into the enclave.
+Currently an application can do this through a system call,
+but this needs to be automated at start-up if the necessary configuration is given.
+Configuration needed for a mount includes the mount source path which is on an insecure filesystem,
+mount target directory within the TEE filesystem,
+and other data relating configuration like if it is read-only or not.
+
+## Overview
+
+Currently three file-systems are supported by Mystikos:
+
+* ramfs is a CPIO archive that is totally TEE memory resident once loaded.
+It is used for rootfs file-systems and is included as part of the tee integrity measurement.
+A ramfs archive can also be included in the rootfs image and an application can mount that on a target mount point.
+No encryption is supported.
+
+* ext2fs filesystem image that is hosted in the non-TEE filesystem for a rootfs.
+It is currently used for the rootfs filesystem and can be integrity protected and encrypted.
+It can also be mounted from the rootfs directory to a target mount point by the application.
+The integrity and encryption keys are part of the TEE measurement.
+
+* hostfs is a directory residing within the insecure host that is mounted into the TEE.
+More hostfs directories can be mounted to a target directory by the application.
+There is no integrity checks done on this type of filesystem.
+
+The first dilemma with mounts is the source of the mount is untrusted for ext2 and hostfs.
+Only ext2 filesystem images support integrity and encryption where the integrity can be validated within the TEE.
+If the decryption key can be achieved securely within the TEE then the decryption can be done securely too without the key being passed in from the host.
+
+The second dilemma is that every host this runs on *can* have different disk layouts and the source of the mount may change every time.
+This is especially true in a container world where different filesystem images may be downloaded to different temporary locations that may differ with each run.
+
+When mounting file-systems into the TEE it is good to have the configuration that is secure being part fo the measured configuration of the application,
+but keep the host specific mount information separate and specifiable from separate configuration that is passed on the command-line.
+
+## TEE configuration
+
+A new node will be added to the JSON configuration that is used in the signing of the application through sign-sgx or package-sgx commands:
+
+```json
+mount: [
+    {
+        "target": "<target>",
+        "type": "<type>",
+        "flags": [ "<flag>" ],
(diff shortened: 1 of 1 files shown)


doc/design/mount-config-design.md
```diff
@@ -0,0 +1,115 @@
...
+
+During the signing of the application through sign-sgx or package-sgx commands via the myst command line tool the TEE mount configuration from the json configuration file is added to the signed application image and cannot be changed.
+
+
+On a host machine or container the application will be run either as an application in a directory with signed binaries, or via the self contained package binary.
+The new mount commandline parameter will be added to whatever command was used before, with one mount parameter option per mount point.
+
+The myst command will then launch the Mystikos runtime into the TEE and will pass the mount command line parameters in.
+
+Once the Mystikos kernel is entered the kernel will use the TEE-side JSON configuration and the command line parameters to call a set of myst_syscall_mount() calls, one for each mount point that is specified on the command line whos target path matches that of the mount JSON configuration.
```
bodzhang, line 111: With the debug mode, where `myst` command is used to load a rootfs inside an enclave, does the mount commandline parameters require a config.json file with mount targets info be used in the commandline? Current `myst` commandline support skipping config.json. Should the debug mode skip the TEE-side configuration check if no mount target info is provided in config.json or the default config.json is implied (no config.json provided in Commandline)? For debug mode, if no mount target info is provided through config.json, I think the implementation can assume mounting flag is r/w.

doc/design/mount-config-design.md
```diff
@@ -0,0 +1,115 @@
...
+
+When mounting file-systems into the TEE it is good to have the configuration that is secure being part fo the measured configuration of the application,
+but keep the host specific mount information separate and specifiable from separate configuration that is passed on the command-line.
+
+## TEE configuration
+
+A new node will be added to the JSON configuration that is used in the signing of the application through sign-sgx or package-sgx commands:
+
+```json
+mount: [
```
bodzhang, line 43: TEE enforceable mount info (target, type, flags, etc, not source) might be included in secret release policy. With mount info represented as json array (following OCI spec),  it's complex to describe a policy such as "if only these 3 volumes  (with the specific properties) are mounted", or "as long as these 3 volumes are mounted". Intuitively, converting the array to name/value pairs, for example disk0:{target, type...}, disk1:{target, type ...}, might make the policy syntax simpler. But at this point, I don't have a full design. Starting with json array for mount info following OCI spec makes sense, but we should consider the possibility that it might need to change in the future.

paulcallen: There is going to be a full re-design of the config at some point so I am expecting changes. 
The format you suggested was one of the designs I considered. I also thought about "target": {"type":"ext2", "flags","flag"} which I thought was a little cleaner, but it did not conform to the OCI spec and chose to not do that.
The parameters that are needed for ext2 that are used from the data mount parameter for keys we can also include in the flags as a "extencryptkey=<key>". Breaking with OCI would allow cleaner configuration though so these parameters can be individual parameters instead of flags.

doc/design/mount-config-design.md
paulcallen: This is in fact a good question.
I am getting increasingly worried about the divergence between debug and signed/packaged mode at runtime in terms of environment variables and more.
I think the behaviour should be consistent across both so the config.json needs to be passed as a parameter and the mount parameters will behave the same in all three modes.

doc/design/mount-config-design.md
```diff
@@ -0,0 +1,115 @@
+# Automatic filesystem mount configuration design
+
+During startup we need the ability to mount directories from the insecure host into the enclave.
+Currently an application can do this through a system call,
+but this needs to be automated at start-up if the necessary configuration is given.
+Configuration needed for a mount includes the mount source path which is on an insecure filesystem,
+mount target directory within the TEE filesystem,
+and other data relating configuration like if it is read-only or not.
+
```
jxyang, line 9: What if the user want to mount to a location from host that already exists in the enclave? For example, the enclave might have files under `/etc`, and if we map host `/etc` to enclave `/etc` with the config file and mount option, what's the behavior? The host /etc overwrites the enclave /etc, or they co-exist?

paulcallen: If you mount a source directory over a target directory that has files in then you can no longer see the old files, only the new ones. I believe mount -bind allows you to see what is already there as well.
I assume the same is also the case with a file being mounted into target directory where the file exists... that one will become hidden and the newly mounted one will be seen.
That said, if someone is already connected to a inode (like a bash shell) and is in a directory at the time the mount happens, that directory will still look the same after the mount has happened until the context moves outside the mount location and back in, in which case they will now see the new mounted directory.
Note: this is the linux kernel implementation based on what I read [here](https://unix.stackexchange.com/questions/198542/what-happens-when-you-mount-over-an-existing-folder-with-contents?rq=1)
I do not know the current behaviour in Mystikos

doc/design/mount-config-design.md
jxyang: I think it's reasonable to provide `-mount` option to users in debug mode with or without a config.json file.

jxyang (approved):

jxyang:
LGTM. We can always revisit this during the implementation.

PR #295, opened 2021-04-16
Thread processor affinity

mikbras:
Add support for the following syscalls:

- **SYS_sched_getaffinity** - delegated to target
- **SYS_sched_setaffinity** - delegated to target
- **SYS_getcpu** - delegated to target
- **SYS_mbind** - stub only for now (to enable Python webserver solution)

Also adds a way to map from a **kernel thread TID** to a **target thread TID**.


tests/ltp/Makefile
```diff
@@ -185,6 +185,3 @@ one:
 	$(MYST) mkcpio ramfs.appdir ramfs
 	sudo -E $(RUNTEST) $(MYST_EXEC) $(OPTS) $(FS) $(TEST) $(NL)
 endif
-
```
bodzhang, line 188: Is the deletion here intentional?

include/myst/thread.h
```diff
@@ -88,6 +88,9 @@ struct myst_thread
     /* unique thread identifier (same as pid for main thread) */
     pid_t tid;
 
+    /* the value returned by gettid() on the target (for this thread) */
```
bodzhang, line 91: `myst_gettid()` still returns the internal `tid` above. `myst_sched_getaffinity()` and myst_sched_setaffinity()' perform the internal-tid to `target_tid` translation. The comment should be fixed.

bodzhang (changes requested):

solutions/aspnet/Makefile
```diff
@@ -4,11 +4,11 @@ include $(TOP)/defs.mak
...
 
-#ifdef STRACE
-#OPTS += --strace
-#endif
+ifdef STRACE
+OPTS += --strace
+endif
 
-OPTS = --memory-size=256m
+OPTS += --memory-size=2048m
```
jxyang, line 11: This setting should has no effect as the same setting in config.json takes precedence.

solutions/aspnet/Makefile
```diff
@@ -17,7 +17,7 @@ run: appdir private.pem
 	test -f server.pid && kill -9 `cat server.pid` || true
 	$(MYST) package appdir private.pem config.json
 	myst/bin/$(APPNAME) $(OPTS) & echo $$! > server.pid
-	sleep 15
+	sleep 60
```
jxyang, line 20: 60 seconds is probably too much for waiting. Why it takes much longer now to boot Mystikos?

solutions/python_webserver/Makefile
```diff
@@ -7,7 +7,7 @@ ifdef STRACE
 OPTS = --strace
 endif
 
-OPTS = --memory-size=512m
+OPTS = --memory-size=1024m
```
jxyang, line 10: See my comment above.

include/myst/kernel.h
```diff
@@ -95,6 +95,9 @@ typedef struct myst_kernel_args
     /* The number of threads that can be created (including the main thread) */
     size_t max_threads;
 
+    /* The tid/pid of the main thread passed from the host */
+    pid_t target_tid;
```
jxyang, line 99: This applies to all ethreads. So the comment about just "main thread" is not accurate.

tools/myst/enc/enc.c
```diff
@@ -735,7 +740,7 @@ int myst_load_fssig(const char* path, myst_fssig_t* fssig)
 #define ENCLAVE_PRODUCT_ID 1
 #define ENCLAVE_SECURITY_VERSION 1
 #define ENCLAVE_DEBUG true
-#define ENCLAVE_HEAP_SIZE 131072
+#define ENCLAVE_HEAP_SIZE 262144
```
jxyang, line 743: Is this the OE heap? Why does it need to double?

bodzhang (approved):

mikbras:
I found that it takes 5 minutes on a bare metal SGX NUC for the ``aspnet`` solution to run. Previously it took only 1 minute. To work around this, I added the UNIPROCESSOR macro. When defined ``sched_getaffinity`` only reports a single processor, causing ``aspnet`` to run in 60 seconds again.

include/myst/mman.h
```diff
@@ -140,4 +150,13 @@ int myst_mman_free_size(myst_mman_t* mman, size_t* size);
 
 void myst_mman_dump_vads(myst_mman_t* mman);
 
+int myst_mman_mprotect(myst_mman_t* mman, void* addr, size_t len, int prot);
```
jxyang, line 153: Do we have to mingle `mprotect` work in this PR?

samples/openmp/Makefile
```diff
@@ -19,10 +19,10 @@ rootfs: appdir
 	# $(MYST) mkcpio appdir rootfs
 
 run:
-	$(MYST_EXEC) $(OPTS) --roothash=roothash ext2rootfs /app/openmp-test
+	OMP_PROC_BIND=true $(MYST_EXEC) $(OPTS) --roothash=roothash ext2rootfs /app/openmp-test
```
jxyang, line 22: What's this for?

PR #306, opened 2021-04-20
PLEASE DO NOT REVIEW: Maps

mikbras:
This example verifies permission propagation performed by mremap. Conclusions:
- when mremap() makes a mapping the same size or smaller, the permissions of the overlapping pages are propagated.
- when mremap() makes a mapping larger, the permissions for all pages must be the consistent (them same), else mremap() fails with EFAULT.


mikbras:
Closing since it PR was for demonstrative purposes only.

PR #309, opened 2021-04-21
Kernel Stacks

mikbras:
This PR introduces the **kernel stacks region**, which sets up a kernel stack for each thread (1024 currently but will be configurable later with the --max-threads option). 

The kernel switches to a kernel stack when handling a syscall and switches back to the user stack upon return. Every kernel stack has a guard page at the top of the stack, which causes an easy-to-trace segmentation fault on stack overflow (so that the exact site of overflow can be determined in a backtrace in gdb).

This change also reduces kernel stack usage by putting large buffers onto the heap (e.g., PATH_MAX buffers). Functions like this:

```C++
int f(const char* path)
{
    int ret = 0;
    char dirname[PATH_MAX]; /* 4096 bytes */
    char filename[PATH_MAX]; /* 4096 bytes */
    
    split(path, dirname, filename);

    /* do something here with v->dirname and v->filename */    

done:
    return ret;
}
```

Are transformed to use heap memory as follows.

```C++
int f(const char* path)
{
    int ret = 0;
    struct vars
    {
        char dirname[PATH_MAX]; /* 4096 bytes */
        char filename[PATH_MAX]; /* 4096 bytes */
    };
    struct vars* v = NULL;

    if (!(v = malloc(sizeof struct vars)))
        ERAISE(-ENOMEM);
    
    ECHECK(split_path(path, v->dirname, v->filename));

    /* do something here with v->dirname and v->filename */    

done:

    if (v)
        free(v);

    return ret;
}
```

In cases where a certain function  (**split_path()** in this example) is used very often, the function could be modified to produce heap-allocated output parameters. This PR postpones such an approach for later to minimize possible introduction of bugs. The above pattern (struct vars) therefore has two man advantages.
- It is less prone to introduce bugs (as opposed to rewriting many functions)
- It can group several buffers into a single allocation (more efficient).

The current PR adds compiler pragmas like this one:

```C++
#pragma GCC diagnostic error "-Wstack-usage=256"
```

And compiler flags like this one:

```
-Wstack-usage=512
```

These detect functions that use more than the given stack-usage limit.

The ``myst_syscall`` function itself is unoptimized and allows for a 4096 byte stack. It will be the focus of future work.

The original kernel stack size was 128 kilobytes but was reduced to 64 kilobytes as a result of this work. 32 kilobytes works in most places except during attestation operations where the OE stack grows very deep. This is a place for further optimization.

### DIFF
--- a/ext2/ext2.c
+++ b/ext2/ext2.c
@@ -190,16 +190,16 @@ static ssize_t _read(myst_blkdev_t* dev, size_t offset, void* data, size_t size)
     uint32_t rem;
     uint8_t* ptr;
     const size_t blksz = MYST_BLKSIZE;
-    struct vars
+    struct locals
     {
         uint8_t blk[MYST_BLKSIZE];
     };
-    struct vars* v = NULL;
+    struct locals* locals = NULL;
 
     if (!dev || !data)
         goto done;
 
-    if (!(v = malloc(sizeof(struct vars))))
+    if (!(locals = malloc(sizeof(struct locals))))
         goto done;
 
     /* calculate the block number */
@@ -226,7 +226,7 @@ static ssize_t _read(myst_blkdev_t* dev, size_t offset, void* data, size_t size)
             uint32_t off; /* offset into this block */
             uint32_t len; /* bytes to read from this block */
 
-            if (dev->get(dev, i, v->blk) != 0)
+            if (dev->get(dev, i, locals->blk) != 0)
                 goto done;
 
             /* If first block */
@@ -240,7 +240,7 @@ static ssize_t _read(myst_blkdev_t* dev, size_t offset, void* data, size_t size)
             if (len > rem)
                 len = rem;
 
-            memcpy(ptr, &v->blk[off], len);
+            memcpy(ptr, &locals->blk[off], len);
             rem -= len;
             ptr += len;
         }
@@ -250,8 +250,8 @@ static ssize_t _read(myst_blkdev_t* dev, size_t offset, void* data, size_t size)
 
 done:
 
-    if (v)
-        free(v);
+    if (locals)
+        free(locals);
 
     return ret;
 }
@@ -268,11 +268,11 @@ static ssize_t _write(
     uint32_t rem;
     uint8_t* ptr;
     const size_t blksz = MYST_BLKSIZE;
-    struct vars
+    struct locals
     {
         uint8_t blk[MYST_BLKSIZE];
     };
-    struct vars* v = NULL;
+    struct locals* locals = NULL;
 
     if (!dev || !data)
         goto done;
@@ -295,7 +295,7 @@ static ssize_t _write(
     }
     else
     {
-        if (!(v = malloc(sizeof(struct vars))))
+        if (!(locals = malloc(sizeof(struct locals))))
             goto done;
 
         for (i = blkno, rem = size, ptr = (uint8_t*)data; rem; i++)
@@ -304,7 +304,7 @@ static ssize_t _write(
             uint32_t len; /* bytes to write from this block */
 
             /* Fetch the block */
-            if (dev->get(dev, i, v->blk) != 0)
+            if (dev->get(dev, i, locals->blk) != 0)
                 goto done;
 
             /* If first block */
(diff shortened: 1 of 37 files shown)


jxyang:
I am confused about this line:
> Every kernel stack has a guard page at the top of the stack, which causes an easy-to-trace segmentation fault on stack overflow

The stack grows downward. Do you actually mean the guard page is at the bottom?

mikbras:
> I am confused about this line:
> 
> > Every kernel stack has a guard page at the top of the stack, which causes an easy-to-trace segmentation fault on stack overflow
> 
> The stack grows downward. Do you actually mean the guard page is at the bottom?

The page is at the top of the stack but at the bottom of the memory chunk. I am using stack nomenclature. The stack has the following layout.

```
top                          bottom
[guard page][stack pages...]
```

ext2/ext2.c
```diff
@@ -273,13 +300,12 @@ static ssize_t _write(
...
         for (i = blkno, rem = size, ptr = (uint8_t*)data; rem; i++)
         {
-            uint8_t blk[MYST_BLKSIZE];
             uint32_t off; /* offset into this block */
             uint32_t len; /* bytes to write from this block */
 
             /* Fetch the block */
-            if (dev->get(dev, i, blk) != 0)
-                return -1;
+            if (dev->get(dev, i, v->blk) != 0)
```
jxyang, line 307: We should only need to malloc `v->blk` for the last block in this branch.

ext2/ext2.c
```diff
@@ -568,22 +604,33 @@ static int _put_blkno(ext2_t* ext2, uint32_t blkno)
...
+
     return ret;
 }
 
 static int _get_blkno(ext2_t* ext2, uint32_t* blkno)
 {
     int ret = 0;
-    ext2_block_t bitmap;
+    struct vars
+    {
```
jxyang, line 625: Define `struct vars` at the file scope so we don't have to define it over and over again.

ext2/ext2.c
```diff
@@ -898,7 +971,8 @@ static int _load_file(
 {
     int ret = 0;
     struct stat st;
-    char block[1024];
+    char* block = NULL;
+    const size_t block_size = 1024;
```
jxyang, line 975: Make it `MYST_BLKSIZE`.

jxyang:
Some early feedbacks. Not yet finished.

ext2/ext2.c
```diff
@@ -181,14 +184,23 @@ static bool _zero_filled_u32(const uint32_t* s, size_t n)
...
+    {
+        uint8_t blk[MYST_BLKSIZE];
+    };
+    struct vars* v = NULL;
 
     if (!dev || !data)
-        return -1;
+        goto done;
+
+    if (!(v = malloc(sizeof(struct vars))))
```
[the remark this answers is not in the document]
mikbras: I started down that path initially (optimizing for all these special cases) but the change list was getting very large and very hard to verify for correctness. So I decided to break the PR into two parts: phase-1 and phase-2, which will focus on the optimizations you describe here. For example, I plan to use the following pattern and others.

```
#include <myst/bufalloc.h>
#include <errno.h>

int f(size_t size)
{
    char buf[128];
    void* p;

    if (!(p = myst_buf_malloc(buf, sizeof(buf), size)))
        return -ENOMEM;
    ...

    myst_buf_free(buf, p);
    return 0;
}
```
```

ext2/ext2.c
mikbras: I was aware of this case but wanted to address in phase-2 due to risk. But I did update the PR in this case because it is relatively easy to verify (fingers crossed). :)

ext2/ext2.c
mikbras: The definition varies from function to function.

ext2/ext2.c
mikbras: ``MYST_BLKSIZE`` is 512 and I am afraid to change it. I am just trying to preserve existing behavior for now.

jxyang:
You don't have to close and reopen the PR to run the pipeline again. Just click on `mystikos-pr-pipeline... details` and there is a rerun button on the page.

jxyang (approved):

mikbras:
> You don't have to close and reopen the PR to run the pipeline again. Just click on `mystikos-pr-pipeline... details` and there is a rerun button on the page.

That only works if you wait for it to finish I think.

PR #317, opened 2021-04-23
PLEASE DO NOT MERGE: epoll-client bug testing

mikbras:


jxyang:
@Mike, can you translate this into an enhancement of /tests/epoll, i.e., we test both the in-enclave clients and the outside-enclave clients?

PR #320, opened 2021-04-26
add --memcheck option for packaged mode

mikbras:


jxyang:
Should we add it to exec-sgx as well?

jxyang:
Looks like we can pass `--memcheck` for packaged apps already. @Francis-Liu Is that the case?

Francis-Liu:
@jxyang Yes, PR #319 added `--memcheck` to packaged mode. And I have been using it.

paulcallen (approved):

PR #324, opened 2021-04-27
multi-threaded poll()

mikbras:
This PR allows multiple-threads to call poll(). It works around a quirk with ``solutions/aspnet`` that caused a hang before.

### DIFF
--- a/solutions/aspnet/Makefile
+++ b/solutions/aspnet/Makefile
@@ -8,7 +8,7 @@ ifdef STRACE
 OPTS += --strace
 endif
 
-OPTS = --memory-size=1024m
+OPTS += --memory-size=1024m
 
 all: appdir private.pem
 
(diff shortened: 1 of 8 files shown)


solutions/aspnet/Makefile
```diff
@@ -17,13 +17,28 @@ run: appdir private.pem
 	test -f server.pid && kill -9 `cat server.pid` || true
 	$(MYST) package appdir private.pem config.json
 	myst/bin/$(APPNAME) $(OPTS) & echo $$! > server.pid
-	sleep 15
+	sleep 60
```
jxyang, line 20: This would increase pipeline run time by 45 seconds. On my VM, `sleep 20` is enough to launch the server.

target/shared/poll.c
```diff
@@ -31,41 +55,111 @@ static int _set_nonblock(int fd)
...
+    {
+        const uint64_t x = WAKE_MAGIC;
 
-    if (write(_wakefds[1], &x, sizeof(x)) != x)
-        return -EINVAL;
+        for (struct waker* p = _wakers; p; p = p->next)
+        {
+            if (write(p->pipefd[1], &x, sizeof(x)) != sizeof(x))
+            {
+                // return -EINVAL;
```
jxyang, line 100: Under which cases would the `write` fail?

jxyang (approved):

solutions/aspnet/Makefile
mikbras: My machine needs 50 seconds.

target/shared/poll.c
mikbras: If the pipe is full (in which case the write is not needed to wake the thread, so can safely be ignored). I will add a comment.

PR #327, opened 2021-04-27
Support SYS_mprotect with OCALL

bodzhang:
add myst_tcall_mprotect stub
Co-authored-by: Mike Brasher <mikbras@microsoft.com>

Expand mman test for mprotect

Optimized prot_vector scan

Fix mistakes in revious rebase, and cleanup


kernel/exec.c
```diff
@@ -811,7 +811,18 @@ int myst_exec(
...
+    // for certain apps, especially when on-demand stack growth is not supported
+    // yet
+    /* .Net runtime set the default minimum stack size for MUSL to 1536 * 1024,
+     *  if env. variable COMPlus_DefaultStackSize is not present or set to other
+     *  value. For MUSL, it also probes the stack limit of the primary thread
+     *  using _alloca(min_stack_size) and writing to a stack variable on top of
+     *  the stack, during CoreCLR init. Set the default stack size as 1536 *
+     *  1024 plus 8 pages reserves for kernel and earlier stack frames
+     */
+    const size_t stack_size = 392 * PAGE_SIZE;
```
jxyang, line 824: Is this still required with the recent kstack work?

kernel/mman.c
```diff
@@ -18,27 +18,28 @@
 **
 ** The memory space has the following layout.
 **
-**     <---VADs---><---BREAK---><--UNASSIGNED--><---------MAPPED---------->
+**     <-VADs/Vector-><---BREAK---><--UNASSIGNED--><---------MAPPED------->
```
jxyang, line 21: This needs to be reconciled with the mman redesign proposal from Mike.

kernel/exec.c
bodzhang: Still required, as .NET runtime require at least 384 pages for stack.

kernel/exec.c
```diff
@@ -822,7 +822,18 @@ int myst_exec(
...
+    // const size_t stack_size = 64 * PAGE_SIZE;
```
mikbras, line 836: Maybe delete this line now that we've agree on the new size.

kernel/mman.c
```diff
@@ -426,6 +429,90 @@ static uintptr_t _mman_find_gap(
...
+    {                                                          \
+        memset(                                                \
+            (MMAN)->prot_vector +                              \
+                ((uintptr_t)ADDR - (MMAN)->start) / PAGE_SIZE, \
+            PROT,                                              \
+            (LEN) / PAGE_SIZE);                                \
+    }
+
+/* check and get permission tracking, if insisitent, return -1*/
+static int _mman_get_prot(
```
mikbras, line 458: Very nice optimization. One optimization might be to use __uint128_t iteration instead, which is even faster. Also I wonder if we could generalize this into a string function, like memcmp_char()?

Anyway, this would all be the future.

mikbras (approved):
A very nice change. Some minor comments but nothing substantial. It looks like you need to run "make format" from the top before checking in.

Also, we might want to schedule some enabling of LTP mprotect tests.

PR #330, opened 2021-04-30
Remove intermediate docker files upon exit

jxyang:

### DIFF
--- a/.azure_pipelines/ci-pipeline-makefile.yml
+++ b/.azure_pipelines/ci-pipeline-makefile.yml
@@ -73,6 +73,7 @@ jobs:
           make clean -C $(Build.SourcesDirectory)/tests
           make clean -C $(Build.SourcesDirectory)/solutions
           sudo rm -rf $(Build.SourcesDirectory)/build/tests
+          docker system prune -f
         displayName: 'Cleanup test directories'
         continueOnError: true
         enabled: true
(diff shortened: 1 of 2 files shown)


.azure_pipelines/ci-pipeline-makefile.yml
```diff
@@ -73,6 +73,7 @@ jobs:
           make clean -C $(Build.SourcesDirectory)/tests
           make clean -C $(Build.SourcesDirectory)/solutions
           sudo rm -rf $(Build.SourcesDirectory)/build/tests
+          docker system prune -f
```
vtikoo, line 76: Why only intermediate images? Can we additionally use `-a` option here, that would delete all unused images.

jxyang: That probably will increase the average pipeline runtime. I am not quite sure how the docker images are cached and reused for the next image building action.

jxyang:
This PR meant to fix nightly pipeline failures. `Disk out of space` is just one of them. More investigations are needed.

Converting to draft for now.

jxyang:
The root cause of the problem is `appbuilder` doesn't delete the exported archives of docker images. This can fill up the `/tmp` directory pretty quickly.

Keep the `docker system prune -f` as an extra measure to save disk space. Also use `df` to report free disk space at the beginning of pipeline run.

vtikoo (approved):

PR #332, opened 2021-04-30
Correctly clean thread from the waiting list if it was blocked and ki…

jxyang:
…lled

### DIFF
--- a/kernel/cond.c
+++ b/kernel/cond.c
@@ -255,6 +255,7 @@ int myst_cond_requeue(
         for (myst_thread_t* p = requeues.front; p; p = next)
         {
             next = p->qnext;
+            p->signal.cond_wait = c2;
             myst_thread_queue_push_back(&c2->queue, p);
         }
     }
--- a/kernel/cond.c
+++ b/kernel/cond.c
@@ -71,6 +71,8 @@ int myst_cond_timedwait(
             return EBUSY;
         }
 
+        // assert(self->signal.cond_wait == NULL);
+        self->signal.cond_wait = c;
         for (;;)
         {
             myst_spin_unlock(&c->lock);
@@ -96,7 +98,9 @@ int myst_cond_timedwait(
             if (ret != 0)
                 break;
         }
+        self->signal.cond_wait = NULL;
     }
+
     myst_spin_unlock(&c->lock);
     myst_mutex_lock(mutex);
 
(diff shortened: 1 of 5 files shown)


kernel/thread.c
```diff
@@ -237,6 +237,28 @@ static void _free_zombies(void* arg)
 
 void myst_zombify_thread(myst_thread_t* thread)
 {
+    // If the thread was blocked on a condvar and got killed,
+    // get out of the waiting list of the condvar.
+    // If the thread exits normally, the cond_wait should be NULL.
```
Francis-Liu, line 242: Can a thread be killed but was not blocked on a condvar?

kernel/thread.c
```diff
@@ -237,6 +237,28 @@ static void _free_zombies(void* arg)
...
+    myst_cond_signal_thread(thread->signal.cond_wait, thread);
```
Francis-Liu, line 243: curious: why not check `cond_wait != NULL` before calling? To me, EINVAL makes me feel it's an error.

kernel/cond.c
```diff
@@ -216,6 +252,7 @@ int myst_cond_requeue(
         for (myst_thread_t* p = requeues.front; p; p = next)
         {
             next = p->qnext;
+            p->signal.cond_wait = c2;
```
Francis-Liu, line 255: Before your fix, requeued threads had their `cond_wait=NULL`?

kernel/cond.c
```diff
@@ -64,6 +65,14 @@ int myst_cond_timedwait(
         /* Add the self thread to the end of the wait queue */
         myst_thread_queue_push_back(&c->queue, self);
 
+        // assert(self->signal.cond_wait == NULL);
```
Francis-Liu, line 68: Why comment out this assert?

mikbras (approved):
Great work! Thanks for doing this. Approved.

Maybe there is some way to use a doubly-linked list in the future so removal of an arbitrary thread can be done with O(1) time complexity.

kernel/thread.c
jxyang: Yes. In that case, the field `thread->signal.cond_wait` would be NULL.

kernel/thread.c
jxyang: Works either way. I just prefer to keep the caller logic simpler.

kernel/cond.c
jxyang: The requeued threads have `cond_wait` to be `c1` without my fix.

kernel/cond.c
jxyang: It was failing before the final fix. I am not sure if we should enable it now.

PR #334, opened 2021-05-03
Reduce _syscall() function stack usage

mikbras:
This PR reduces the stack usage of the ``_syscall`` function from 3 kilobytes to


jxyang:
As we discussed offline, can we combine the `_pairs` table and `_syscalls` table? Each entry in the table would be `{syscall_id, handler, name}`.

syscall.c has grown to over 6000 lines. I hope we can keep the size down by moving some handlers to the files they really belong to. That's not required for this PR though.

PR #345, opened 2021-05-06
PLEASE DO NOT MERGE: Affinity version two

mikbras:


tools/myst/host/syscall.c
```diff
@@ -422,3 +423,29 @@ long myst_utimensat_ocall(
...
+}
+
+MYST_STATIC_ASSERT(sizeof(struct myst_cpu_set) == sizeof(cpu_set_t));
+
+long myst_sched_getaffinity_ocall(
+    pid_t pid,
+    size_t cpusetsize,
+    struct myst_cpu_set* mask)
+{
+    RETURN(syscall(SYS_sched_getaffinity, pid, cpusetsize, mask));
```
bodzhang, line 442: The implementation returns non-zero value on successes. On a 4 core system, I saw return value of 8, with the mask showing 4 cores. Changing the implementation to call libc sched_getaffinity() instead, the return value is 0.

mikbras: I see it too. Still trying to fix.

mikbras: Okay, the man page says that ``sched_getaffinity`` returns the number of bytes written to the mask, which is different than the number of CPUs.

```
On success, the raw sched_getaffinity() system call returns the
       number of bytes placed copied into the mask buffer; this will be
       the minimum of cpusetsize and the size (in bytes) of the
       cpumask_t data type that is used internally by the kernel to
       represent the CPU set bit mask.
```

mikbras: The ``sched_getaffinity`` ocall needed to be fixed to set ``size=cpusetsize``. I added an overwrite test (that failed initially) and then fixed it.

tests/libcxx/tests.all
```diff
@@ -1,120 +1,104 @@
-/app/CharT.pass.cpp.exe
```
bodzhang, line 1: Are the changes in tests/libcxx/tests.all intended?

mikbras: Unfortunately running the tests updates ``tests.all``. Xuejun created an issue for this.

mikbras: I restored all the libcxx test files to the ones in main to remove it from this PR. We will have to figure out why they keep getting updating when you run the tests.

solutions/aspnet/config.json
```diff
@@ -6,7 +6,7 @@
     // The stack size of each ethread
     "StackMemSize": "400k",
     // The number of ethreads
-    "NumUserThreads": 16,
+    "NumUserThreads": 32,
```
jxyang, line 9: Can we delete the "NumUserThreads" setting altogether?

solutions/python_webserver/Makefile
```diff
@@ -20,10 +20,16 @@ appdir:
 run:
 	test -f server.pid && kill -9 `cat server.pid` || true
 	$(MYST_EXEC) $(OPTS) rootfs /usr/local/bin/python3 /app/hello_server.py & echo $$! > server.pid
-	sleep 20
+	sleep 35
```
jxyang, line 23: I'd rather pass in `--max-affinity-cpus`, if that's the cause of the slow down, to save every pipeline a few seconds.

solutions/aspnet/config.json
mikbras: I don't know. Maybe but I assume it runs faster with less theads.

solutions/python_webserver/Makefile
mikbras: No, this is due to my bare metal machine being slower.

solutions/aspnet/config.json
mikbras: Oh I see. You're wondering if we can remove the setting altogether? If so, I think we should create an issue. It would be one less thing for users to worry about.

mikbras: Also we can do away with ``StackMemSize``

jxyang: I am wondering if this still works after deleting `NumUserThreads`. And if so, we can remove it from this PR. I don't want surprises when we run this test on a 8-core machine.

mikbras: I did not change this yet. I think we need to have a separate PR to consider removal of these settings since they don't just affect affinity.

solutions/python_webserver/Makefile
mikbras: I updated the ``python_webserver`` and ``aspnet`` solutions to wait on an output string from the server rather than sleeping for a fixed period.

solutions/aspnet/config.json
jxyang: Affinity definitely makes it worse because dotnet runtime tends to create more threads when it sees more CPUs from getaffinity. That's why I ask for the change in the same PR.

jxyang (approved):

PR #351, opened 2021-05-10
SYS_fchdir

vtikoo:
- Add support for fchdir.
- Update ltp test lists. 2/3 fchdir ltp tests pass. The failing test fchdir03 relies on fork. 
- Proposed fix for ramfs bug where if there is an open fd of directory and `rmdir` is called on the directory, `close(fd)` will fail as the underlying inode has been freed.


kernel/syscall.c
```diff
@@ -1690,6 +1690,57 @@ long myst_syscall_chdir(const char* path)
...
+
+    /* Get file path */
+    ECHECK((*fs->fs_realpath)(
+        fs, file, locals->realpath, sizeof(locals->realpath)));
+
+    myst_spin_lock(&process_thread->main.cwd_lock);
+
+    char* tmp = strdup(locals->realpath);
+    if (tmp == NULL)
+        ERAISE(-ENOMEM);
```
jxyang, line 1730: Move line 1728-1730 before the lock period. Ideally we should lock as short as possible.

tests/ltp/ext2fs_tests_unhandled_syscalls.txt
```diff
@@ -1,368 +1,368 @@
-/ltp/testcases/kernel/syscalls/add_key/add_key01: *** kernel panic: syscall.c(5019): myst_syscall(): unhandled syscall: SYS_add_key()
```
jxyang, line 1: If we can get rid of the line# in the error message, we wouldn't have so many false updates.

kernel/syscall.c
vtikoo: Makes sense!

jxyang (approved):

tests/ltp/ext2fs_tests_unhandled_syscalls.txt
vtikoo: Yeah agree. Have changed the local test runner script to log only the syscall name. Will update this file in the next commit.

vtikoo: Unhandled syscall list is updated now.

jxyang (approved):

PR #352, opened 2021-05-11
eventfd2 syscall

mikbras:
Implement the ``eventfd2`` syscall (and hence the ``eventfd()`` C library function) and add two tests.
- ``tests/eventfd``
- ``tests/eventfdpoll``

This change is highly orthogonal and therefore is unlikely to cause problems with existing workloads.

### DIFF
--- a/kernel/eventfddev.c
+++ b/kernel/eventfddev.c
@@ -1,0 +1,532 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include <errno.h>
+#include <fcntl.h>
+#include <limits.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/eventfd.h>
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+#include <myst/cond.h>
+#include <myst/defs.h>
+#include <myst/eraise.h>
+#include <myst/eventfddev.h>
+#include <myst/id.h>
+#include <myst/panic.h>
+#include <myst/process.h>
+#include <myst/round.h>
+#include <myst/syscall.h>
+
+#define MAGIC 0x9906acdc
+
+#define MAX_COUNTER_VALUE ((uint64_t)0xfffffffffffffffe)
+
+struct myst_eventfd
+{
+    uint32_t magic;
+    int flags;
+    int fdflags;
+    uint64_t counter;
+    myst_mutex_t mutex;
+    myst_cond_t cond;
+};
+
+MYST_INLINE bool _valid_eventfd(const myst_eventfd_t* eventfd)
+{
+    return eventfd && eventfd->magic == MAGIC;
+}
+
+static void _lock(myst_eventfd_t* eventfd)
+{
+    myst_assume(_valid_eventfd(eventfd));
+    myst_mutex_lock(&eventfd->mutex);
+}
+
+static void _unlock(myst_eventfd_t* eventfd)
+{
+    myst_assume(_valid_eventfd(eventfd));
+    myst_mutex_unlock(&eventfd->mutex);
+}
+
+static int _eventfd(
+    myst_eventfddev_t* eventfddev,
+    unsigned int initval,
+    int flags,
+    myst_eventfd_t** eventfd_out)
+{
+    int ret = 0;
+    myst_eventfd_t* eventfd = NULL;
+
+    if (!eventfddev || !eventfd_out)
+        ERAISE(-EINVAL);
+
+    if ((flags & ~(EFD_CLOEXEC | EFD_NONBLOCK | EFD_SEMAPHORE)))
+        ERAISE(-EINVAL);
+
+    /* Create the read eventfd */
+    {
+        if (!(eventfd = calloc(1, sizeof(myst_eventfd_t))))
+            ERAISE(-ENOMEM);
+
+        eventfd->magic = MAGIC;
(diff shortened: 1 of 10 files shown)


kernel/eventfddev.c
```diff
@@ -0,0 +1,537 @@
...
+    {
+        memcpy(buf, &eventfd->counter, sizeof(uint64_t));
+
+        if ((eventfd->flags & EFD_SEMAPHORE))
+            eventfd->counter--;
+        else
+            eventfd->counter = 0;
+
+        myst_cond_signal(&eventfd->cond);
+        _unlock(eventfd);
```
jxyang, line 131: Why don't we move `_unlock` to "done" section?

kernel/eventfddev.c
```diff
@@ -0,0 +1,537 @@
...
+    if (!eventfddev || !_valid_eventfd(eventfd))
+        ERAISE(-EBADF);
+
+    if (!buf)
+        ERAISE(-EINVAL);
+
+    if (count < sizeof(uint64_t))
+        ERAISE(-EINVAL);
+
+    memcpy(&value, buf, sizeof(uint64_t));
```
jxyang, line 197: Should we check the overflow first before copying?

kernel/eventfddev.c
```diff
@@ -0,0 +1,537 @@
...
+
+    if (value == 0xffffffffffffffff)
```
jxyang, line 199: Replace 0xfffffffff with `MAX_COUNTER_VALUE+1`

kernel/eventfddev.c
```diff
@@ -0,0 +1,537 @@
...
+    _lock(eventfd);
+
+    if (!_valid_eventfd(eventfd))
+    {
+        /* cannot unlock since mutex is no longer valid */
+        ERAISE(-EBADF);
+    }
+
+    /* if able to write the the counter */
+    if (eventfd->counter + value < MAX_COUNTER_VALUE)
```
jxyang, line 214: The left side could wrap around (depending on the compiler). Suggest changing to `eventfd->counter < MAX_COUNTER_VALUE - value`.

kernel/eventfddev.c
```diff
@@ -0,0 +1,537 @@
...
+    for (;;)
+    {
+        if (myst_cond_wait(&eventfd->cond, &eventfd->mutex) != 0)
+        {
+            /* unexpected */
+            _unlock(eventfd);
+            ERAISE(-EPIPE);
+        }
+
+        if (eventfd->counter + value < MAX_COUNTER_VALUE)
```
jxyang, line 240: Same as above

kernel/eventfddev.c
```diff
@@ -0,0 +1,537 @@
...
+        {
+            eventfd->counter += value;
+            myst_cond_signal(&eventfd->cond);
+            _unlock(eventfd);
+            ret = sizeof(uint64_t);
+            goto done;
+        }
+    }
+
+    _unlock(eventfd);
```
jxyang, line 250: Move `_unlock` to the done section?

kernel/eventfddev.c
```diff
@@ -0,0 +1,537 @@
...
+
+    if (!eventfddev || !_valid_eventfd(eventfd) || !statbuf)
+        ERAISE(-EINVAL);
+
+    memset(&buf, 0, sizeof(buf));
+    buf.st_dev = 13; /* event fd */
+    buf.st_ino = (ino_t)eventfd;
+    buf.st_mode = 0600;
+    buf.st_nlink = 1;
+    buf.st_uid = MYST_DEFAULT_UID;
```
jxyang, line 315: We probably should use `thread->uid`. But I doubt it makes any difference to any application.

kernel/eventfddev.c
```diff
@@ -0,0 +1,537 @@
...
+    if (!_valid_eventfd(eventfd))
+    {
+        /* cannot unlock since mutex is no longer valid */
+        ERAISE(-EBADF);
+    }
+
+    /* if the counter is readable (non-zero) */
+    if (eventfd->counter != 0)
+    {
+        memcpy(buf, &eventfd->counter, sizeof(uint64_t));
```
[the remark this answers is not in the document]
mikbras: Unfortunately there is no guarantee that ``buf`` is 8-byte aligned.

kernel/eventfddev.c
mikbras: Good idea. Fixed.

kernel/eventfddev.c
mikbras: Good idea. Fixed.

kernel/eventfddev.c
mikbras: I replaced the constant with ``UINT64_MAX``. The manual page actually refers to the constant strangely.

kernel/eventfddev.c
mikbras: Nice idea. Fixed.

kernel/eventfddev.c
mikbras: Fixed.

kernel/eventfddev.c
mikbras: Fixed.

kernel/eventfddev.c
mikbras: Good catch! I verified that uid and gid for eventfd's is 0, so I changed both to zero.

jxyang (approved):

PR #355, opened 2021-05-12
Support glibc static mutex initializers

mikbras:
This PR fixes the hang that occurs in the following program on the second lock (when compiled with glibc headers).

```C++
#define _GNU_SOURCE
#include <stdio.h>
#include <unistd.h>
#include <syscall.h>
#include <pthread.h>

int main()
{
    static pthread_mutex_t m = PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
    pthread_mutex_lock(&m);
    pthread_mutex_lock(&m);

    return 0;
}
```
The glibc headers put the mutex type into the 5th word, whereas musl libc expects the type to be in the 1st word. This fix replaces the ``_m_type`` macro with ``_m_get_type()`` and ``_m_set_type()``. See the source comments for more details.

This fix handles the following types of static initializers.
- PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP
- PTHREAD_ERRORCHECK_MUTEX_INITIALIZER_NP
- PTHREAD_ADAPTIVE_MUTEX_INITIALIZER_NP -- not supported by musl libc so mapped to normal mutex

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -102,6 +102,7 @@ DIRS += devfs
 DIRS += callonce
 DIRS += syscall_exception
 DIRS += dotnet-proc-maps
+DIRS += mutex
 #DIRS += mprotect
 
 __tests:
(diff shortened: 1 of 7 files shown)


jxyang (approved):
Thanks for the careful approach.

PR #357, opened 2021-05-13
reenable and optimize ext2/plain test

mikbras:
This PR reenables the ext2/plain test (disabled as part of another PR). It also optimizes byte scanning operations with the new ``myst_memcchr`` function.

It also optimizes the kernel's ``memset`` function to be highly efficient.

### DIFF
--- a/ext2/ext2.c
+++ b/ext2/ext2.c
@@ -27,7 +27,7 @@
 #define EXT2_TRIPLE_INDIRECT_BLOCK 14
 
 /* limit the stack size of the functions below */
-#pragma GCC diagnostic error "-Wstack-usage=256"
+#pragma GCC diagnostic error "-Wstack-usage=512"
 
 #if 0
 #define CHECKS
@@ -162,24 +162,9 @@ static void _dirent_init(
     ent->rec_len = _next_mult(ent->rec_len, 4);
 }
 
-static bool _zero_filled_u32(const uint32_t* s, size_t n)
+MYST_INLINE bool _zero_filled_u32(const uint32_t* s, size_t n)
 {
-    const uint32_t* p = s;
-    const uint32_t* end = s + n;
-
-    while (n > 4 && !p[0] && !p[1] && !p[2] && !p[3])
-    {
-        n -= 4;
-        p += 4;
-    }
-
-    while (n > 0 && !p[0])
-    {
-        n--;
-        p++;
-    }
-
-    return p == end;
+    return myst_memcchr(s, 0, n * sizeof(uint32_t)) == NULL;
 }
 
 static ssize_t _read(myst_blkdev_t* dev, size_t offset, void* data, size_t size)
@@ -197,9 +182,6 @@ static ssize_t _read(myst_blkdev_t* dev, size_t offset, void* data, size_t size)
     struct locals* locals = NULL;
 
     if (!dev || !data)
-        goto done;
-
-    if (!(locals = malloc(sizeof(struct locals))))
         goto done;
 
     /* calculate the block number */
@@ -225,6 +207,9 @@ static ssize_t _read(myst_blkdev_t* dev, size_t offset, void* data, size_t size)
         {
             uint32_t off; /* offset into this block */
             uint32_t len; /* bytes to read from this block */
+
+            if (!(locals = malloc(sizeof(struct locals))))
+                goto done;
 
             if (dev->get(dev, i, locals->blk) != 0)
                 goto done;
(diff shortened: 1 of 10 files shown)


ext2/ext2.c
```diff
@@ -645,29 +630,15 @@ static int _get_blkno(ext2_t* ext2, uint32_t* blkno)
...
-                    p++;
-                    i--;
-                }
-
-                lblkno = (8 * sizeof(uint64_t)) * (n - i);
-            }
+            if (p)
+                lblkno = (p - locals->bitmap.data) * 8;
+            else
+                lblkno = locals->bitmap.size * 8;
```
jxyang, line 641: Shouldn't we break out here?

utils/Makefile
```diff
@@ -20,6 +20,7 @@ CFLAGS += -fdata-sections
 CFLAGS += -Wno-conversion
 CFLAGS += -Wno-parentheses
 CFLAGS += -Wstack-usage=512
+CFLAGS += -O3
```
jxyang, line 23: As discussed, we should not use `-O3` for debug builds.

tests/ext2/plain/ext2.c
```diff
@@ -273,8 +283,8 @@ int main(int argc, const char* argv[])
         assert(buf.st_size == sizeof(alpha));
         assert(buf.st_mode == mode);
         assert(buf.st_nlink == 1);
-        assert(buf.st_uid == 0);
-        assert(buf.st_gid == 0);
+        assert(buf.st_uid == myst_syscall_geteuid());
```
jxyang, line 286: Why don't we use `getuid` and `getgid` directly?

tests/ext2/plain/Makefile
```diff
@@ -34,11 +34,14 @@ EXT2FS_SIZE=134217728
 tests:
 	mkdir -p $(SUBOBJDIR)
 	sudo $(MYST) mkext2 --force --size=$(EXT2FS_SIZE) ext2dir $(EXT2FS)
+	chown $(USER).$(USER) $(EXT2FS)
 	$(PREFIX) $(RUNTEST) $(PREFIX) $(SUBBINDIR)/ext2 $(EXT2FS)
-	rm -rf $(EXT2FS)
 
 cg:
```
jxyang, line 40: Please use a name more descriptive than `cg`.

utils/Makefile
paulcallen: these kind of flags should also be in the main top-level options that other projects share. Having every project have its own options like this is a real pain!

paulcallen (approved):

ext2/ext2.c
mikbras: Fixed. It works either way but your way is easier to read.

utils/Makefile
mikbras: Removed. I left it in by accident.

tests/ext2/plain/Makefile
mikbras: Fixed. Changed to cachegrind.

tests/ext2/plain/ext2.c
mikbras: It will work either way but I switched to use to shorter forms.

utils/strings.c
```diff
@@ -328,3 +328,49 @@ bool myst_isspace(char c)
...
+    /* if more than 16 bytes and p is 16-byte aligned */
+    if (n > 16 && ((ptrdiff_t)p & 0x000000000000000f) == 0)
+    {
+        const __uint128_t* pp = (__uint128_t*)p;
+        size_t nn = n / 16;
+        const __uint128_t* end = pp + nn;
+        __uint128_t cc;
+        memset(&cc, c, sizeof(cc));
+
+        while (nn > 0 && *pp == cc)
```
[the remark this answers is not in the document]
mikbras: This is the BSD ``memcchr`` as explained by the header file. I added this to ``<myst/strings.h>``:

```
// BSD memcchr() function: return a pointer to the first byte that is not
// equal to c or null if not found.
```

I adopted the nomenclature and the exact signature of the function (which determines the variable names).

```
void* myst_memcchr(const void* b, int c, size_t len);
```

When implementing standard functions I stick with the naming introduced by the signature as much as possible. For example, considr:

```
void *memset(void *s, int c, size_t n);
```
This is the POSIX standard definition of ``memset`` (from The Open Group). I adhere to their naming.

I did rename variables like ``pp`` and ``cc`` to ``p16`` and ``c16`` to indicate that these are the 16-byte versions of ``p`` and ``c``.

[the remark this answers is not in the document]
jxyang: I think those names were poor choices and was only appropriate when the community was dominated by hackers. But I am fine with them for now.

jxyang (approved):

ext2/ext2.c
mikbras: It turns out is is a continue (to advance to the next group). :)

PR #358, opened 2021-05-14
add cpprestsdk test suite

praenubilus:
add CPPRESTSDK (https://github.com/microsoft/cpprestsdk) solution. Running the in stock test suites.
All six test suites will end with myst_run_thread_ecall(): failure. Requires further investigation. 

recipe `run-cpio` and `gdb` can run/debug single test suites out of the following 6 test suites:
- libhttpclient_test.so
- libhttplistener_test.so
- libjson_test.so
- libpplx_test.so
- libstreams_test.so
- liburi_test.so

### DIFF
--- a/solutions/Makefile
+++ b/solutions/Makefile
@@ -32,6 +32,7 @@ endif
 DIRS += nodejs
 DIRS += numpy_core_tests
 DIRS += redis
+DIRS += cpprestsdk
 
 .PHONY: $(DIRS)
 
--- a/solutions/cpprestsdk/Dockerfile
+++ b/solutions/cpprestsdk/Dockerfile
@@ -1,0 +1,22 @@
+# stage 1 build
+FROM alpine:3.12.3 AS cpprest-base-image
+
+RUN apk add --no-cache curl build-base bash git boost-dev cmake zlib-dev openssl-dev ninja
+
+RUN rm -rf /app;mkdir -p /app
+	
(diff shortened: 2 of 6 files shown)


praenubilus:
I have updated the cpprestsdk solution and applied patch to skip 4G read/write tests(ignored). However, with the latest main branch, I find some previously fixed problem appeared again. 
- libstreams_test.so: file_buffer_tests:SequentialReadWrite... https://github.com/deislabs/mystikos/pull/362
- libhttpclient_test.so: compression_tests:compress_client_server
- libhttplistener_test.so request_stream_tests:test_chunked_transfer will fail assertion randomly(not every run) https://github.com/deislabs/mystikos/pull/363

jxyang:
We have fixed many issues in the latest release. Could you try the test suite again? @praenubilus

praenubilus:
> We have fixed many issues in the latest release. Could you try the test suite again? @praenubilus

@jxyang I have done a full test, most of the test cases will fail with `Unhandled exception: Request canceled by user. FAILED`. The exception is thrown by GTEST, looks like user authorization related error happens. Is there any user settings update need to be applied since late May 2020(the last commit for this PR)?

bodzhang:
@praenubilus, the cpprestsdk test suite is expected to work with the Main branch now, can you rebase and check?

bodzhang (approved):

PR #359, opened 2021-05-14
Adding codeql code scan config

mrragava:
Workflow config to enable CodeQL code scan.
Ref: https://docs.github.com/en/code-security/secure-coding/setting-up-code-scanning-for-a-repository

Signed-off-by: Ragavan Dasarathan <mrragava@gmail.com>

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -12,11 +12,6 @@
 name: "CodeQL"
 
 on:
-  push:
-    branches: [ main ]
-  pull_request:
-    # The branches below must be a subset of the branches above
-    branches: [ main ]
   schedule:
     - cron: '0 0 * * *'
 
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,57 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-18.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    # Initializes the CodeQL tools for scanning.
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-18.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
```
jxyang, line 31: Is `cpp` the same thing as `c` in this regard? We are definitely using c not c++.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,57 @@
...
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
```
jxyang, line 19: Does it mean the scan job is triggered for every PR? How long does it take for a scan?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,57 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y runc containerd docker.io libmbedtls-dev
+        sudo systemctl start docker
+        sudo systemctl enable docker && sudo chmod 666 /var/run/docker.sock
+        sudo apt-get install python3-setuptools
+
+    - name: Building Mystikos
+      run: |
+        make MYST_PRODUCT_BUILD=1 -j4
```
jxyang, line 54: Why do we use PRODUCT_BUILD?

jxyang: I thought this tool scans the source code, not the binary builds?

.github/workflows/codeql-analysis.yml
mrragava: Yes, CodeQL refers 'cpp' for  both 'c' 'and 'c++'

.github/workflows/codeql-analysis.yml
mrragava: CodeQL scan job will be triggered for every PR and populates the detentions to "Security Alerts". It takes approximately 30mins for a scan. 
ACC security team will triage these alerts on a regular basis and triage the valid security issues by creating an GH issue.

.github/workflows/codeql-analysis.yml
mrragava: Yes CodeQL scans only the source code. 
I used MYST_PRODUCT_BUILD to reduce build time since it skips building test projects.

paulcallen:
Do we have existing failures that are reported on Github if this is enabled? 
Are there false-alarms/false-positively? 
What does the output look like?
If we do have errors then it would be good to fix them before we enable this on all runs.

.github/workflows/codeql-analysis.yml
jxyang: As discussed, can we remove the push and pr triggers and leave only the schedule trigger?

jxyang (approved):

jxyang:
Thanks for the contribution. Look forward to the 1st scan report.

PR #361, opened 2021-05-14
Add cpuinfo proc entry

vtikoo:
### Summary
This PR adds support for /proc/cpuinfo file. 
Two calls are made to the target, first one to get the size of the cpuinfo file on the host. The second one passes a buffer of that size, and the contents of /proc/cpuinfo on the host are copied over into the buffer.
The buffer is cached inside the kernel, so that subsequent requests can be serviced without target calls.

### DIFF
--- a/kernel/procfs.c
+++ b/kernel/procfs.c
@@ -15,6 +15,7 @@
 #include <myst/procfs.h>
 
 static myst_fs_t* _procfs;
+static char* _cpuinfo_buf = NULL;
 
 int procfs_setup()
 {
@@ -53,6 +54,9 @@ int procfs_teardown()
         myst_eprintf("failed to release procfs\n");
         return -1;
     }
+
+    /* free cached cpuinfo information */
+    free(_cpuinfo_buf);
 
     return 0;
 }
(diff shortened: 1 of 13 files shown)


kernel/procfs.c
```diff
@@ -189,6 +193,38 @@ static int _self_vcallback(myst_buf_t* vbuf)
     return ret;
 }
 
+static int _cpuinfo_vcallback(myst_buf_t* vbuf)
+{
+    int ret = 0;
+
+    if (!vbuf)
+        ERAISE(-EINVAL);
+
```
jxyang, line 202: Should we clear vbuf first?

target/shared/cpuinfo.c
```diff
@@ -0,0 +1,47 @@
...
+#include <myst/tcall.h>
+#include <fcntl.h>
+#include <string.h>
+
+int myst_tcall_cpuinfo_size()
+{
+    int fd, nbytes, size = 0;
+    char buf[1024];
+
+    fd = open("/proc/cpuinfo", O_RDONLY);
```
jxyang, line 14: You can use `ftell` to get the file size quicker:
```c
fp = fopen("/proc/cpuinfo", "r");
fseek(fp, 0L, SEEK_END);
sz = ftell(fp);
```

target/shared/cpuinfo.c
```diff
@@ -0,0 +1,47 @@
...
+{
+    int fd, nbytes, count = 0;
+    char tmp[1024];
+    char* p = buf;
+
+    fd = open("/proc/cpuinfo", O_RDONLY);
+    if (fd < 0)
+        return errno;
+
+    while ((nbytes = read(fd, tmp, sizeof(tmp))))
```
jxyang, line 35: You can read `size` bytes once without the need for a loop.

target/shared/cpuinfo.c
vtikoo: Tried this, didnt work. Digging further it looks like /proc files depending on the seq file implementation of lseek don't support `SEEK_END`.

kernel/procfs.c
vtikoo: It is cleared on L220.

target/shared/cpuinfo.c
vtikoo: Fixed.

jxyang (approved):

PR #363, opened 2021-05-14
cpprestsdk: fix for httplistener test

mikbras:
This PR fixes the following cpprestsdk tests:

```
libhttplistener_test.so: request_stream_tests:test_chunked_transfer
```

This test calls into the target (OE enclave) and fails to allocate enough memory to serialize (flatten) a gather-write buffer. 

**The fix flattens the gather-write buffer on the kernel side instead of the target side.**

### DIFF
--- a/tools/myst/enc/enc.c
+++ b/tools/myst/enc/enc.c
@@ -748,7 +748,6 @@ int myst_load_fssig(const char* path, myst_fssig_t* fssig)
 #define ENCLAVE_SECURITY_VERSION 1
 #define ENCLAVE_DEBUG true
 
-#define ENCLAVE_HEAP_SIZE 131072
 #define ENCLAVE_HEAP_SIZE (256 * 1024)
 
 #ifdef MYST_ENABLE_GCOV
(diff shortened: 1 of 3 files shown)


tools/myst/enc/enc.c
```diff
@@ -748,7 +748,7 @@ int myst_load_fssig(const char* path, myst_fssig_t* fssig)
 #define ENCLAVE_SECURITY_VERSION 1
 #define ENCLAVE_DEBUG true
 
-#define ENCLAVE_HEAP_SIZE 131072
+#define ENCLAVE_HEAP_SIZE (256 * 1024)
```
jxyang, line 751: We have tried very hard to keep the OE heap size small. This would throw away the work for one extreme case. Could this be avoided?

mikbras: I'm working on an alternative now. :)

mikbras: I posted a new fix that flattens the buffer on the kernel side.

jxyang (approved):

PR #364, opened 2021-05-15
Add locking wrapper for file systems

mikbras:


paulcallen (approved):
May want to look into read/write lock at some point through as there are generally more reads than writes

PR #365, opened 2021-05-17
[WIP] Mount single file using symlink

Francis-Liu:


kernel/mount.c
```diff
@@ -25,9 +27,11 @@
 #include <myst/verity.h>
 
 #define MOUNT_TABLE_SIZE 8
+#define AUTOMOUNT_DIR "/tmp/automounts"
```
mikbras, line 30: Maybe move this to ``/run/mystikos/automounts``.

kernel/mount.c
```diff
@@ -39,13 +43,17 @@ static size_t _mount_table_size = 0;
 static myst_spinlock_t _lock = MYST_SPINLOCK_INITIALIZER;
 
 static bool _installed_free_mount_table = false;
+static bool _created_automount_dir = false;
```
mikbras, line 46: maybe drop this flag and just check for the existence of the directory each time like this.

```
struct stat statbuf;

if (!(myst_syscall_stat(pathname, &statbuf) && S_ISDIR(statbuf.st_mode)))
{
}
```

kernel/mount.c
```diff
@@ -299,6 +310,114 @@ static const char* _find_arg(const char* args[], const char* name)
 }
 #endif /* MYST_ENABLE_EXT2FS */
 
+#ifdef MYST_ENABLE_HOSTFS
+static bool _find_mount_source(const char* source, char path[PATH_MAX])
```
mikbras, line 314: I think I would consider returning a const pointer to the path rather than making a copy.

kernel/mount.c
```diff
@@ -299,6 +310,114 @@ static const char* _find_arg(const char* args[], const char* name)
...
+    return found;
+}
+
+static long _new_automount_dir(char mountdir[PATH_MAX])
+{
+    long ret = -1;
+    int mount_num = 1;
+
+    myst_strlcpy(mountdir, AUTOMOUNT_DIR, PATH_MAX);
+    if (!_created_automount_dir)
```
mikbras, line 335: Again, I would just check for the existence of the directory as described above.

kernel/mount.c
```diff
@@ -299,6 +310,114 @@ static const char* _find_arg(const char* args[], const char* name)
...
+    {
+        ECHECK(myst_syscall_mkdir(mountdir, 0x777));
+        _created_automount_dir = true;
+    }
+
+    while (1)
+    {
+        snprintf(mountdir, PATH_MAX, "%s/%d", AUTOMOUNT_DIR, mount_num++);
+        if (myst_syscall_access(mountdir, F_OK) != 0)
```
mikbras, line 344: This loop seems only to terminate if the mountdir does not exist.

kernel/mount.c
```diff
@@ -299,6 +310,114 @@ static const char* _find_arg(const char* args[], const char* name)
...
+    }
+    return found;
+}
+
+static long _new_automount_dir(char mountdir[PATH_MAX])
+{
+    long ret = -1;
+    int mount_num = 1;
+
+    myst_strlcpy(mountdir, AUTOMOUNT_DIR, PATH_MAX);
```
mikbras, line 334: It looks like ``mountdir`` is overloaded (used for the base directory and the subdirectory). Can you use two variables? Maybe ``mountdir`` and ``mountsubdir``?

kernel/mount.c
```diff
@@ -299,6 +310,114 @@ static const char* _find_arg(const char* args[], const char* name)
...
+
+    return ret;
+}
+
+static int _mount_single_file(
+    myst_fs_t* fs,
+    const char* source,
+    const char* target)
+{
+    int ret = -1;
```
mikbras, line 366: Our convention is to assign ``ret`` equal to zero.

kernel/mount.c
```diff
@@ -299,6 +310,114 @@ static const char* _find_arg(const char* args[], const char* name)
...
+
+    /* create a symlink */
+    if (myst_syscall_access(target, F_OK) == 0 &&
+        myst_syscall_unlink(target) != 0)
+    {
+        myst_eprintf("kernel: unlink failed: %s\n", target);
+        ERAISE(-EINVAL);
+    }
+
+    myst_strlcpy(locals->mountpath + strlen(locals->mountpath), "/", PATH_MAX);
```
mikbras, line 395: You can just use ``myst_strcat`` like this:

```
myst_strlcat(locals->mountpath, "/", PATH_MAX);
```

kernel/mount.c
```diff
@@ -299,6 +310,114 @@ static const char* _find_arg(const char* args[], const char* name)
...
+    myst_strlcpy(
+        locals->mountpath + strlen(locals->mountpath),
+        locals->sourcebase,
+        PATH_MAX);
```
mikbras, line 399: Again, you can use ``myst_strlcat``.
```
myst_strlcat(locals->mountpath, locals->sourcebase, PATH_MAX);
```

tools/myst/host/exec.c
```diff
@@ -289,6 +303,22 @@ int exec_action(int argc, const char* argv[], const char* envp[])
          */
         cli_getopt(&argc, argv, "--app-config-path", &commandline_config);
 
+        /* Get --symlink options (there may be more than one) */
+        {
+            const char* arg = NULL;
+
+            symlinks[0] = NULL;
+
+            if ((cli_getopt(&argc, argv, "--symlink", &arg) == 0))
```
mikbras, line 312: Do we still need the ``--symlink`` option?

mikbras (commented):
Hi Francis. Thank you so much for doing this. This looks very clean. A few minor comments and question about there we still need the ``--sysmlink`` option.

kernel/mount.c
Francis-Liu: Good point!

kernel/mount.c
Francis-Liu: Yes, it is what is intended -- the code jumps out of the loop once it finds a non-existence mount folder, e.g. ${AUTOMOUNT_DIR}/1, ${AUTOMOUNT_DIR}/2, ...

kernel/mount.c
Francis-Liu: It turns out I can use `AUTOMOUNT_DIR` directly. So `mountdir` is the function parameter which is used to return the new mount target directory.

kernel/mount.c
Francis-Liu: Cool!

tools/myst/host/exec.c
Francis-Liu: I think we can hide that option since it is only used internally, and --symlink currently doesn't check if LINK is defined in config.json.

kernel/mount.c
Francis-Liu: Done.

kernel/mount.c
Francis-Liu: Done.

kernel/mount.c
Francis-Liu: Done.

kernel/mount.c
Francis-Liu: Done.

Francis-Liu:
@mikbras I have created a new issue #383 for documentation changing. Will work on it in a separate PR.

kernel/mount.c
mikbras: I'm sorry. I missed that. :)

mikbras (approved):
Looks really great Francis. Thank you so much for doing this.

Issue #372, opened 2021-05-18
Mystikos failed to support dlopen of .so built with initial-exec Thread Local Storage mode


bodzhang:
Mystiko's musl dlopen failed to load .so built with "initial-exec" TLS mode, showing "initial-exec TLS resolves to dynamic definition" error.

The failure can be reproduced with the following test case, attempting to dlopen the default libgomp.so build for Ubuntu18.04.

Dockerfile:

```c
FROM ubuntu:18.04

RUN apt-get update && apt-get install -y gcc libgomp1
   
RUN rm -rf /app;mkdir -p /app
ADD omp-dlopen /app

RUN cd /app && \
    gcc -o omp-dlopen -Wall -g omp-dlopen.c -ldl

CMD ["/bin/bash"]
```

omp_dlopen.c:
```c
#include <dlfcn.h>
#include <stdio.h>

int main()
{
    void* handle;

    handle = dlopen("libgomp.so.1.0.0", RTLD_GLOBAL);
    if (!handle)
    {
        printf("%s\n", dlerror());
        return (-1);
    }

    printf("dlopen libgomp.so succeed\n");

    dlclose(handle);
    return (0);
}
```
Makefile:
```c
ifdef STRACE
OPTS = --strace
endif

all: appdir rootfs

appdir:
    $(APPBUILDER) Dockerfile

rootfs: appdir
    sudo $(MYST) mkext2 appdir ext2rootfs
    sudo chown $(USER).$(USER) ext2rootfs
    $(MYST) fssig --roothash ext2rootfs > roothash
    # $(MYST) mkcpio appdir rootfs

run:
    $(MYST_EXEC) $(OPTS) --roothash=roothash ext2rootfs /app/omp-dlopen

gdb:
    $(MYST_GDB) --args $(MYST_EXEC) $(OPTS) --roothash=roothash ext2rootfs /app/openmp-test

clean:
    rm -rf rootfs ext2rootfs appdir roothash
```
From [All about thread-local storage | MaskRay](https://maskray.me/blog/2021-02-14-all-about-thread-local-storage):
"If you add the __attribute((tls_model("initial-exec"))) attribute, a thread-local variable can use this model in -fpic mode. If the object file is linked into an executable, everything is fine. If the object file is linked into a shared object, the shared object generally needs to be an immediately loaded shared object. The linker sets the DF_STATIC_TLS flag to annotate a shared object with initial-exec TLS relocations.
[...]

PR #375, opened 2021-05-19
Handle dlopened library with R_X86_64_TPOFF64 symbols

jxyang:

MUSL maintains a counter for the number of loaded dynamic libraries that contain `TLS` section (verifiable with `readelf -l <.so file path>`). When a `R_X86_64_TPOFF64` symbol needs to be relocated, it checks whether the `tls_id` of the library being relocated is smaller than the number of total number of loaded dynamic libraries that contain `TLS` section.

The bug is that musl failed to update the total number before the relocation for the dlopen case, causing an assertion failure.

Fix https://github.com/deislabs/mystikos/issues/372

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -59,6 +59,7 @@ DIRS += tlscert
 DIRS += tlscert2
 DIRS += wake_and_kill
 DIRS += cross_fs_symlinks
+DIRS += openmp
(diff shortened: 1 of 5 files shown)


mikbras (approved):
Nice work.

PR #378, opened 2021-05-20
Add --max-threads options

mikbras:
Add ``--max threads`` option to speed up the LTP tests (cutting the total execution time in half).

### DIFF
--- a/include/myst/file.h
+++ b/include/myst/file.h
@@ -24,7 +24,7 @@ ssize_t myst_writen(int fd, const void* data, size_t size);
 
 int myst_copy_file(const char* oldpath, const char* newpath);
 
-int myst_copy_file_fd(char* oldpath, int newfd);
+int myst_copy_file_fd(const char* oldpath, int newfd);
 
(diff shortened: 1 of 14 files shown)


mikbras:
> From the total pipeline run time, I do not see significant improvement. Could you apply the flag to libc and libcxx test suites?
> 
> With this option, we effectively bring `NumUserThreads` back (albeit with command line vs previously config.json). And it again will trip users who specify `--max-threads` and later code requires more threads than what was specified. That's the fallacy we try to avoid.
> 
> My vote is that unless `--max-threads` makes a huge performance difference, we should opt for usability.

I'm not sure why this option makes the LTP run twice as fast on my box but not in the pipeline. It may be the server class machines perform so well with so many processors that it doesn't matter. :)

mikbras:
I confirmed that this option does cut the LTP run time down from **3m15s** to **1m36s** (by taking the difference of the timestamps from the start to the end of the LTP tests in the main branch and this branch). The execution time of the pipeline is not constant since more than one pipeline VM can run on the same machine.

PR #385, opened 2021-05-21
Forward hardware exceptions as signals to kernel

jxyang:

This PR translate relevant hardware exceptions into signals and forward them from enclave to the kernel. If the user application has registered a handler, the kernel would further forward the signal to the custom handler.

With the forwarding, I have re-enabled tests/mprotect by providing a lenient handler that won't crash on SIGSEGV.

### DIFF
--- a/kernel/signal.c
+++ b/kernel/signal.c
@@ -128,7 +128,10 @@ static long _default_signal_handler(unsigned signum)
     return 0;
 }
 
-static long _handle_one_signal(unsigned signum, siginfo_t* siginfo)
+static long _handle_one_signal(
+    unsigned signum,
+    siginfo_t* siginfo,
+    mcontext_t* mcontext)
 {
     long ret = 0;
     ECHECK(_check_signum(signum));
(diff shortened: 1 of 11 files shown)


kernel/signal.c
```diff
@@ -298,3 +308,11 @@ long myst_signal_clone(myst_thread_t* parent, myst_thread_t* child)
 done:
     return ret;
 }
+
+long myst_handle_host_signal(unsigned signum, mcontext_t* mcontext)
+{
+    siginfo_t siginfo = {0};
+    siginfo.si_code = SI_KERNEL;
```
bodzhang, line 315: At least for certain SIGSEGV handler implementation, `siginfo->si_addr` is needed, for example, [.NET runtime's SIGSEGV handler](https://github.com/dotnet/runtime/blob/0761780988b93a4d1cc98dc5d4e2d84db29e33fd/src/coreclr/pal/src/exception/signal.cpp#L556). The .NET runtime's signal handlers or other signal handler implementations might require other information in `siginfo`. It's better to support signal/exception specific `siginfo` setting in the HW exception handling code, instead of only allow common `siginfo` setting here.

bodzhang (approved):

bodzhang:
OE is adding support for trust-worthy #PF information in SGX Enclave in [PR 3984](https://github.com/openenclave/openenclave/pull/3984)

PR #387, opened 2021-05-21
fix mmap bug

mikbras:
This fixes a bug in ``mman.c`` where ``memcpy`` copies onto a null buffer. There was a missing ``goto done`` but there were a couple of other places that did not set errors.


salsal97 (approved):
Works on my branch

jxyang (approved):

PR #389, opened 2021-05-22
Fix the bugs in CPUID instr. exception handler and in SYS_sched_getaffinity handler

bodzhang:
The CPUID instruction exception handler missed the required RIP advancement. The RAX, RBX, RCX, RDX value returned to the calling code are corrupted in many cases. The fix is simple, just advancing the RIP by 2 to resume execution on the instruction after the cpuid instruction. The test case under tests/cpuid is enhanced to query a handful of CPUID leaves. It can be further enhanced in the future.

The bug in SYS_sched_getaffinity erroneously over-writes the buffer provided. The fix corrects the implementation by honoring the  buffer size info in `cpusetsize`.

### DIFF
--- a/tools/myst/enc/enc.c
+++ b/tools/myst/enc/enc.c
@@ -116,6 +116,9 @@ static uint64_t _vectored_handler(oe_exception_record_t* er)
                 er->context->rcx = rcx;
                 er->context->rdx = rdx;
 
+                /* Skip over the illegal instruction. */
+                er->context->rip += 2;
+
                 return OE_EXCEPTION_CONTINUE_EXECUTION;
                 break;
             }
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -4537,8 +4537,8 @@ static long _syscall(void* args_)
             // for now, make all threads fixed to cpu 0.
             if (mask != NULL)
             {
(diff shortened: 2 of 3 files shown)


tools/myst/enc/enc.c
```diff
@@ -116,6 +116,9 @@ static uint64_t _vectored_handler(oe_exception_record_t* er)
                 er->context->rcx = rcx;
                 er->context->rdx = rdx;
 
+                /* Skip over the illegal instruction. */
+                er->context->rip += 2;
```
jxyang, line 120: I wonder why previously the `cpuid` test could pass without advancing the rip. Any explanation there?

bodzhang: When RIP was not advanced by mistake, the CPUID instruction will be executed again with RAX value returned by the myst_cpuid_ocall. If the RAX value returned is one of the value in [0,1,4,7,0x80000000,0x80000001], the resulting CPUID instruction exception will be handled by the OE code, instead of the code in tools/myst/enc/enc.c. The OE code advances the RIP after emulating the CPUID result for [0,1,4,7,0x80000000,0x80000001].

jxyang (approved):

PR #396, opened 2021-05-25
Deliver previously blocked signals when they are unmasked

### LINKED ISSUE #394: Hold a masked signal for delivery until it is unmasked
Currently, Mystikos drops a signal if it's masked as blocked for the target thread. Instead, we should save the signal somewhere as pending, and when later a request to remove the blockage on the target threads is received by the kernel, deliver the pending signal.  

From https://man7.org/linux/man-pages/man7/signal.7.html:
>  A signal may be blocked, which means that it will not be
       delivered until it is later unblocked.  Between the time when it
       is generated and when it is delivered a signal is said to be
       pending.

jxyang:
Fix #394

### DIFF
--- a/kernel/signal.c
+++ b/kernel/signal.c
@@ -2,6 +2,7 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <myst/atexit.h>
 #include <myst/eraise.h>
 #include <myst/fsgs.h>
 #include <myst/printf.h>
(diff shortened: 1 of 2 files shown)


paulcallen (commented):
Current implementation has a couple of issues I see:
1. when a thread goes away the pending items are not deleted until process exit. I suggest adding them to a thread pending delivery list (extend siginfos array to be a list of siginfos needing to be delivered)
2. signals are delivered out of order
I would suggest fixing this before proceeding

jxyang:
Superseded by #397

PR #400, opened 2021-05-26
support for OS signing's OpenSSL signing engine

paulcallen:
add parameters to package and sign commands to support the oesign tools'
engine parameters

Not sure how to test this though!

### DIFF
--- a/tools/myst/host/package.c
+++ b/tools/myst/host/package.c
@@ -107,6 +107,9 @@ int _package(int argc, const char* argv[])
     static const size_t max_roothashes = 128;
     const char* roothashes[max_roothashes];
     size_t num_roothashes = 0;
+    const char* signing_engine_key = NULL;
+    const char* signing_engine_name = NULL;
+    const char* signing_engine_path = NULL;
 
     /* Get --pubkey=filename and --roothash=filename options */
     get_archive_options(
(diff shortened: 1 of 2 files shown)


tools/myst/host/package.c
```diff
@@ -126,6 +129,21 @@ int _package(int argc, const char* argv[])
         goto done;
     }
 
+    cli_getopt(&argc, argv, "--signing-engine-name", &signing_engine_name);
+    cli_getopt(&argc, argv, "--signing-engine-path", &signing_engine_path);
+    cli_getopt(&argc, argv, "--signing-engine-key", &signing_engine_key);
+    // engine path optional
```
jxyang, line 135: Is this documented somewhere?

> // engine path optional

jxyang:
We are just a pass through to OE. https://github.com/openenclave/openenclave/tree/master/tests/tools/oesign/test-engine

tools/myst/host/package.c
paulcallen: are you referring to the new parameters, or the optional path parameter?
from oe-sign help: 
  -n, --engine             name of the crypto engine to use, for example
                           "pkcs-11".
  -i, --key-id             engine-specific ID string specifying the
                           desired enclave signing key from the engine.
  -p, --load-path          [optional] absolute path to the shared object
                           that implements the engine.

PR #411, opened 2021-05-28
Patch musl with missing glibc symbols

jxyang:

This PR patches the following missing symbols required by various Python packages and Azure SDK in C++ that were built targeting glibc:

**__mbrlen**
**__openat_2**
**_obstack_newchunk**
**argz_create_sep**
**clnt_create**
**clnt_spcreateerror**
**makecontext**
**mallopt**
**obstack_vprintf**
**register_printf_modifier**
**register_printf_specifier**
**register_printf_type**
**setcontext**
**xdr_enum**
**xdr_pointer**
**xdr_string**
**__res_iclose**
**__resolv_context_get_preinit**
**__resolv_context_get_override**
**__resolv_context_get**
**__resolv_context_put**
**inet_nsap_ntoa**
**reallocarray**
**__resp**
**__explicit_bzero_chk**
**__libc_alloca_cutoff**

For the above symbols, this PR doesn't provide the actual implementation. Any invocation from the user application should trigger assertion failures. But at least now we can launch the application.

sys_errlist
sys_nerr
_poll
__gmtime_r
__sendmmsg
errno
__h_errno
_stpncpy
__open_nocancel
__read_nocancel
__snprintf
__close_nocancel

For the above symbols, this PR provides the actual implementation by aliasing to or unhiding existing MUSL functions. Any invocation from the user application should succeed.

### DIFF
--- a/crt/glibc_stub.c
+++ b/crt/glibc_stub.c
@@ -10,6 +10,8 @@
 #define weak_alias(old, new) \
     extern __typeof(old) new __attribute__((__weak__, __alias__(#old)))
 
+#define ENOMEM 12
+
 __thread int errno;
 
 const char* const sys_errlist[] = {};
@@ -34,7 +36,7 @@ typedef struct __mbstate_t
 size_t __mbrlen(const char* s, size_t n, mbstate_t* ps)
 {
     assert(0);
-    return 0;
+    return -1;
 }
 weak_alias(__mbrlen, mbrlen);
 
@@ -56,7 +58,7 @@ void _obstack_newchunk(struct obstack* h, int length)
 int __argz_create_sep(const char* string, int delim, char** argz, size_t* len)
 {
     assert(0);
-    return 0;
+    return ENOMEM;
 }
 weak_alias(__argz_create_sep, argz_create_sep);
 
@@ -83,7 +85,7 @@ char* clnt_spcreateerror(const char* msg)
 int __libc_mallopt(int param_number, int value)
 {
     assert(0);
-    return -1;
+    return 0;
 }
 weak_alias(__libc_mallopt, mallopt);
 
@@ -144,7 +146,7 @@ typedef struct XDR
 int xdr_enum(XDR* xdrs, int* ep)
 {
     assert(0);
-    return -1;
+    return false;
 }
 
 typedef int (*xdrproc_t)(XDR*, void*, ...);
@@ -220,5 +222,5 @@ void __explicit_bzero_chk(void* __dest, size_t __len, size_t __destlen)
 int __libc_alloca_cutoff(size_t size)
 {
     assert(0);
-    return -1;
+    return false;
 }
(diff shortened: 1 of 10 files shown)


crt/glibc_stub.c
```diff
@@ -0,0 +1,224 @@
...
+char* clnt_spcreateerror(const char* msg)
+{
+    assert(0);
+    return NULL;
+}
+
+int __libc_mallopt(int param_number, int value)
+{
+    assert(0);
+    return -1;
```
bodzhang, line 86: https://man7.org/linux/man-pages/man3/mallopt.3.html - on error returns 0.

crt/glibc_stub.c
```diff
@@ -0,0 +1,224 @@
...
+
+void _obstack_newchunk(struct obstack* h, int length)
+{
+    assert(0);
+}
+
+int __argz_create_sep(const char* string, int delim, char** argz, size_t* len)
+{
+    assert(0);
+    return 0;
```
bodzhang, line 59: https://www.man7.org/linux/man-pages/man3/argz_create_sep.3.html - return ENOMEM on error

crt/glibc_stub.c
```diff
@@ -0,0 +1,224 @@
...
+}
+
+typedef struct __mbstate_t
+{
+    unsigned __opaque1, __opaque2;
+} mbstate_t;
+size_t __mbrlen(const char* s, size_t n, mbstate_t* ps)
+{
+    assert(0);
+    return 0;
```
bodzhang, line 37: https://www.man7.org/linux/man-pages/man3/mbrlen.3.html - 0 return value indicates a null wide character was recognized. -1 return value if if an invalid multibyte sequence was encountered. -1 is a better error return to prevent the calling function from continue execution.

crt/glibc_stub.c
```diff
@@ -0,0 +1,224 @@
...
+
+void __explicit_bzero_chk(void* __dest, size_t __len, size_t __destlen)
+{
+    assert(0);
+}
+
+int __libc_alloca_cutoff(size_t size)
+{
+    assert(0);
+    return -1;
```
bodzhang, line 223: The function return seems to be defined as a Boolean in glibC. Might not have a good way to set a "error" return to prevent the calling function to continue execution.

bodzhang (changes requested):

crt/glibc_stub.c
bodzhang: Latest PR returns 0/false. The calling code might still progress further. Should the stub implementation use abort().

crt/glibc_stub.c
bodzhang: I'm not sure whether the implementation should return -ENOMEM or ENOMEM. @mikbras, can you chime in?

bodzhang (approved):

PR #412, opened 2021-05-28
Support for lchown and fchownat

vtikoo:
### Summary
- Support for fchownat and lchown syscalls.
- symlink and lstat hostfs operations now propagate identities.
- Path to inode lookup in ramfs and ext2fs was returning ENOENT if an intermediate component is not a directory. Changed it to return ENOTDIR.


ext2/ext2.c
```diff
@@ -5324,6 +5329,62 @@ static int _ext2_fchown(
...
+        &tfs));
+    if (tfs)
+    {
+        /* delegate operation to target filesystem */
+        ECHECK((ret = tfs->fs_lchown(tfs, locals->suffix, owner, group)));
+        goto done;
+    }
+
+    if (!S_ISLNK(locals->inode.i_mode))
+        ERAISE(-ENOTDIR);
```
jxyang, line 5373: Does this mean we return -ENOTDIR when the file is not a soft link? I think in that case, we just treat it as a regular file.

kernel/ramfs.c
```diff
@@ -534,8 +534,10 @@ static int _path_to_inode_recursive(
     {
         for (size_t i = 0; i < ntoks; i++)
         {
-            inode_t* p;
+            if (!S_ISDIR(parent->mode))
```
jxyang, line 537: When would this happen?

tests/ltp/Makefile
```diff
@@ -118,22 +119,27 @@ group-file:
 	echo "grp701:x:701:user700" >> appdir/etc/group
 	echo "grp704:x:704:" >> appdir/etc/group
 	echo "grp705:x:705:" >> appdir/etc/group
+	echo "grp1000:x:1000:" >> appdir/etc/group
 
 add_host_users_and_groups:
 	sudo groupadd -g 100701 grp701
 	sudo groupadd -g 100704 grp704
 	sudo groupadd -g 100705 grp705
+	sudo groupadd -g 101000 grp1000
```
jxyang, line 128: Sometimes the group might already be there from previous build. We should either append ` | ture` or check whether the group/user exists before the add commands.

kernel/uid_gid.c
```diff
@@ -880,6 +881,26 @@ int check_thread_group_membership(gid_t group)
...
 }
 
+static int _path_checks(const char* pathname)
+{
+    int ret = 0;
+
+    if (!pathname)
+        return -EINVAL;
+
+    if (pathname == (void*)0xffffffffffffffff)
```
jxyang, line 891: Use we use `-1`?

kernel/uid_gid.c
```diff
@@ -1000,3 +1020,50 @@ long myst_syscall_fchown(int fd, uid_t owner, gid_t group)
 
     return ret;
 }
+
+long myst_syscall_lchown(const char* pathname, uid_t owner, gid_t group)
```
jxyang, line 1024: This, `myst_syscall_fchown`, `myst_syscall_chown`, and `myst_syscall_fchownat` should be put together in a new file `chown.c`.

kernel/syscall.c
```diff
@@ -1881,6 +1881,95 @@ long myst_syscall_fstatfs(int fd, struct statfs* buf)
...
+        ERAISE(-EINVAL);
+
+    if ((flags & ~(AT_EMPTY_PATH | AT_SYMLINK_NOFOLLOW)) != 0)
+        ERAISE(-EINVAL);
+
+    if (!(locals = malloc(sizeof(struct locals))))
+        ERAISE(-ENOMEM);
+
+    /* If pathname is absolute, then ignore dirfd */
+    if (*pathname == '/' || dirfd == AT_FDCWD)
```
jxyang, line 1909: I think all these cases can be consolidated with:
```c
   helper_function_to_get_real_path(dirfd, pathname, &locals->realpath);
   ret = (flags & AT_SYMLINK_NOFOLLOW) ? 
                myst_syscall_lchown(locals->realpath, owner, group) : 
                myst_syscall_chown(locals->realpath, owner, group);
```

kernel/ramfs.c
vtikoo: If any intermediate path component is a file. For eg: /reg-file/file2. On linux this returns on ENOTDIR, mystikos was returning an ENOENT.

kernel/uid_gid.c
vtikoo: This really should be a bad address check. One of the lchown ltp tests uses `0xffffffffffffffff` as a bad address test.

ext2/ext2.c
vtikoo: Thanks for pointing this out. Fixed.

tests/ltp/Makefile
vtikoo: Fixed.

kernel/syscall.c
vtikoo: Thanks for the suggestion, have incorporated it.

kernel/uid_gid.c
vtikoo: Done.

include/myst/syscall.h
```diff
@@ -317,5 +317,18 @@ long myst_syscall_getcpu(unsigned* cpu, unsigned* node);
...
 long myst_syscall_lchown(const char* pathname, uid_t owner, gid_t group);
+long myst_syscall_fchownat(
+    int dirfd,
+    const char* pathname,
+    uid_t owner,
+    gid_t group,
+    int flags);
+
+/* Used by XXXXXat() syscalls */
+long get_absolute_path_from_dirfd(
```
jxyang, line 328: I think the convention is to prefix non-static function names with "myst_".

vtikoo: Renamed.

jxyang (approved):

PR #414, opened 2021-05-29
Paulcallen/fork

paulcallen:


crt/exit.c
```diff
@@ -0,0 +1,32 @@
+#include <myst/syscallext.h>
+#include <stdio.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include "syscall.h"
+
+void do_crt_exit(int code);
+
+/*
+ * - If we were cloned with CLONE_VM | CLOINE_VFORK we are sharing the process
```
vtikoo, line 10:
```suggestion
 * - If we were cloned with CLONE_VM | CLONE_VFORK we are sharing the process
```
spelling nit

crt/fork.c
```diff
@@ -0,0 +1,569 @@
...
+        *stack = (uint8_t*)self->stack - self->stack_size;
+        *stack_size = self->stack_size;
+    }
+    else
+    {
+        const long n = SYS_myst_get_process_stack;
+
+        if (syscall(n, stack, stack_size) != 0)
+        {
+            fprintf(stderr, "cannot retrieve parent stack\n");
```
jxyang, line 151: Do you mean "the process thread of the parent?" My understanding is that the thread that calls "fork" is the parent thread, but it is not necessarily the process thread.

crt/fork.c
```diff
@@ -0,0 +1,569 @@
...
+
+/* ATTN: arrange for this function to be called on exit */
+__attribute__((__unused__)) static void _thread_args_free(void* arg)
+{
+    struct thread_args* args = (struct thread_args*)arg;
+    /* ATTN: release child stack parent is not process thread */
+    free(args);
+}
+
+static int _valid_stack_address(const void* sp, const void* p)
```
jxyang, line 263: Seems this function is replaced by "_within".

crt/fork.c
```diff
@@ -0,0 +1,569 @@
...
+{
+    const uint8_t* start = data;
+    const uint8_t* end = start + size;
+    const uint8_t* p = ptr;
+    bool flag = p >= start && p < end;
+    return flag;
+}
+
+static int _fixup_frame_pointers(
+    const void* parent_sp,
```
jxyang, line 283: We could use a lot of comments in this function.

crt/fork.c
```diff
@@ -0,0 +1,569 @@
...
+** Or:
+**     [ guard ]                td->guard_size | __default_guardsize
+**     [ stack ]                td->stack_size | __default_stacksize
+**     [ tls area | tcb ]       __libc.tls_size
+**     [ tsd area ]             __pthread_tsd_size
+**
+**==============================================================================
+*/
+
+struct pthread* _create_child_pthread()
```
jxyang, line 176: This function create child process stack and initialize the pthread structure. The function name is confusing.

crt/fork.c
```diff
@@ -0,0 +1,569 @@
...
+            return -ENOMEM;
+
+        stack = (uint8_t*)child_pthread->stack - child_pthread->stack_size;
+        stack_size = child_pthread->stack_size;
+
+        _get_parent_stack(&parent_stack, &parent_stack_size);
+
+        assert(stack_size == parent_stack_size);
+
+        if (_fixup_frame_pointers(
```
jxyang, line 413: Do we need to copy parent's stack?

crt/fork.c
```diff
@@ -0,0 +1,569 @@
...
+    else /* child */
+    {
+        pid = 0;
+    }
+
+    return pid;
+}
+
+#if 0
+long _vfork_parent(void* _args)
```
jxyang, line 481: Why do we keep this?

jxyang:
Half way through my review. In general, I wish there were more comments to make the code review and future maintenance easier.

crt/Makefile
```diff
@@ -17,7 +17,12 @@ ifdef MYST_ENABLE_GCOV
 DEFINES += -DMYST_ENABLE_GCOV
 endif
 
-SOURCES = $(wildcard *.c) $(wildcard $(TOP)/third_party/gcompat/gcompat/libgcompat/*.c)
+SOURCES =
+SOURCES += $(wildcard *.c)
+SOURCES += $(wildcard $(TOP)/third_party/gcompat/gcompat/libgcompat/*.c)
+SOURCES += $(wildcard ../kernel/setjmp.s)
```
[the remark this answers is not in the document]
paulcallen: these are stand-alone assembly files. It makes no sense to just copy them so it is either a case of pulling from here, or just move them to a source library directory that both the kernel and the crt pull from. Leaving them where they were seemed like the simplest solution. You want a separate directory that both projects pull in from?

crt/enter.c
```diff
@@ -45,6 +49,14 @@ long myst_syscall(long n, long params[6])
     if (n == SYS_setitimer)
         pthread_once(&_once, _create_itimer_thread);
 
+    if (n == SYS_fork)
```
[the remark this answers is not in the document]
paulcallen: actually we can implement it in the kernel, but we can also implement it in application space also.

crt/fork.c
paulcallen: the stack is part of the pthread structure!

crt/fork.c
```diff
@@ -0,0 +1,569 @@
...
+#include <unistd.h>
+
+#include <myst/backtrace.h>
+#include <myst/lfence.h>
+#include <myst/round.h>
+#include <myst/setjmp.h>
+#include <myst/syscallext.h>
+#include <myst/thread.h>
+
+#define DEFAULT_STACK_SIZE (4 * 1024 * 1024)
```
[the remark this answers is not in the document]
paulcallen: turns out the only place this was used was in a function that is no longer used. Removed this and the function

crt/fork.c
```diff
@@ -0,0 +1,569 @@
...
+extern hidden struct __libc __libc;
+
+enum
+{
+    DT_EXITING = 0,
+    DT_JOINABLE,
+    DT_DETACHED,
+};
+
+struct pthread
```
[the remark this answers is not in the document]
paulcallen: I think I have managed to pull in the pthread_impl.h which has this definition. If so I will remove these definitions.

crt/fork.c
```diff
@@ -0,0 +1,569 @@
...
+    volatile int timer_id;
+    locale_t locale;
+    volatile int killlock[1];
+    char* dlerror_buf;
+    void* stdio_locks;
+    uintptr_t canary_at_end;
+    uintptr_t* dtv_copy;
+};
+
+int myst_round_up(uint64_t x, uint64_t m, uint64_t* r)
```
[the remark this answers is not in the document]
paulcallen: This is just a function that was needed and we cannot pull in other libraries so I cut and pasted it. Keeping the same name seemed logical, but I can rename it easily enough and make it static.

crt/fork.c
paulcallen: removed

crt/fork.c
paulcallen: or the pthread, stack, td. All in a single continuous thunk of memory

crt/fork.c
paulcallen: by the time we have got here we have created and copied the stack, and this is fixing up all the rbp pointers that allow the stack to be walked

crt/fork.c
paulcallen: removed

crt/Makefile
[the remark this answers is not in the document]
mikbras: Maybe we can move them to a shared directory (maybe utils)?

kernel/thread.c
```diff
@@ -259,53 +263,121 @@ void myst_zombify_thread(myst_thread_t* thread)
...
-        }
-        else
-        {
             _zombies_count++;
         }
+        myst_mutex_unlock(&_zombies_mutex);
+    }
+    else
+    {
+        thread->status = MYST_ZOMBIE;
```
jxyang, line 303: When do we free the threads that are not put into the zombie list?

kernel/thread.c
```diff
@@ -237,6 +239,7 @@ static void _free_zombies(void* arg)
 
 void myst_zombify_thread(myst_thread_t* thread)
 {
+#if 0
```
jxyang, line 242: We don't need these clean ups anymore?

kernel/thread.c
```diff
@@ -348,40 +425,173 @@ long myst_syscall_wait4(
     for (;;)
     {
         /* search the zombie list for a process thread */
-        for (myst_thread_t* p = _zombies_head; p; p = p->znext)
+        for (p = _zombies_head; p; p = p->znext)
         {
-            bool match = false;
-
             if (!myst_is_process_thread(p))
```
jxyang, line 430: Every thread in the zombie list is a process thread, right? Do we still need the check?

kernel/thread.c
```diff
@@ -348,40 +425,173 @@ long myst_syscall_wait4(
...
+        {
+            printf(
+                "Waiting for new zombification of child whose pgid=%d from our "
+                "process %u\n",
+                -pid,
+                process->pid);
+        }
+#endif
+        myst_mutex_unlock(&_zombies_mutex);
+        myst_sleep_msec(100);
```
jxyang, line 590: Ideally we should use conditional variables to wake up the waiter.

tests/fork/fork.c
```diff
@@ -0,0 +1,246 @@
...
+{
+    pid_t pid = vfork();
+    if (pid < 0)
+    {
+        fprintf(stderr, "%s: fork() failed: %d\n", argv[0], pid);
+        exit(1);
+    }
+    else if (pid == 0)
+    {
+        //        _printf("*** inside child\n");
```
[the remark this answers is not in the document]
paulcallen: vfork is not implemented yet

crt/Makefile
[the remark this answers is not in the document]
paulcallen: moved them into a new asm directory and pulled them all from there

crt/enter.c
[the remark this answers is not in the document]
paulcallen: implementing in the kernel is now a little more complicated due to the switching stack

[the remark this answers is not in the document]
paulcallen: regarding the locks, well there used to be something in there, but apparently not any more so the lock is gone.

crt/exit.c
```diff
@@ -0,0 +1,32 @@
...
+ *   heap and so also global variables. If this is so we cannot do clean
+ *   shutdown, instead we do a terminate instead.
+ * - If this process is created from vfork the calling thread in the parent
+ *   process is blocked doing a read on this handle to determine when the
```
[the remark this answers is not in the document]
paulcallen: looks like this comment is out of date... updating

crt/fork.c
paulcallen: hmm... The only thread we cannot get the stack from pthread is the main process thread. All other threads are created through pthread_create. 
If this is called from a non-main-process-thread the pthread stack is sufficient, but it is a little confusing. 
I will change the name, and will fail if not called on the main process thread as if we were to return the main process stack from a non-main process thread it would cause all sorts of problems.

tests/fork/fork.c
```diff
@@ -0,0 +1,246 @@
...
+    if (pid < 0)
+    {
+        fprintf(stderr, "%s: fork() failed: %d\n", argv[0], pid);
+        exit(3);
+    }
+    else if (pid == 0)
+    {
+        _printf("*** inside child.. sleeping and waiting for signal\n");
+        sleep(1);
+        _printf("Shutting down child\n");
```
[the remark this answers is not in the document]
paulcallen: the test runs very quickly, so this is sufficient

tests/fork/fork.c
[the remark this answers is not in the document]
paulcallen: I've removed for now

tools/myst/host/exec_linux.c
```diff
@@ -172,6 +175,28 @@ static void _get_options(int* argc, const char* argv[], struct options* opts)
     cli_getopt(argc, argv, "--app-config-path", &opts->app_config_path);
 }
 
+myst_kernel_args_t kernel_args;
+
+static void sigaction_handler(int sig, siginfo_t* si, void* context)
```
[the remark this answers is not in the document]
paulcallen: fixed

tools/myst/host/exec_linux.c
```diff
@@ -172,6 +175,28 @@ static void _get_options(int* argc, const char* argv[], struct options* opts)
...
+{
+    mcontext_t* mcontext = (mcontext_t*)context;
+    kernel_args.myst_handle_host_signal(si, mcontext);
+}
+
+void install_signal_handlers()
```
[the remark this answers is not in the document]
paulcallen: fixed

crt/fork.c
paulcallen: added big comment

kernel/thread.c
paulcallen: I am not sure condition variables work here anyway which is why things broke in the first place. What would the condition variable be? We have zombies? Well we have multiple waits going on here. When it is signaled, do we catch it? Did we miss it? There are very major problems with any kind of signaling. We would have to have a condition variable connected to every process, so if any process is woken we wait on our own and then we know if it is signaled. In that case a zombie would need to signal all running processes.

jxyang:
Finished reviewing half of the 2nd iteration. Stopped at poll.c.

crt/fork.c
```diff
@@ -0,0 +1,420 @@
...
+
+#include <myst/backtrace.h>
+#include <myst/lfence.h>
+#include <myst/round.h>
+#include <myst/setjmp.h>
+#include <myst/syscallext.h>
+#include <myst/thread.h>
+
+static int _round_up(uint64_t x, uint64_t m, uint64_t* r)
+{
```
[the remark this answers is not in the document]
paulcallen: this is in the musl shared library so it is not available here, right?

kernel/process.c
```diff
@@ -1,4 +1,46 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
-
 #include <myst/process.h>
+#include <myst/syscall.h>
+
+long myst_syscall_setpgid(pid_t pid, pid_t pgid, myst_thread_t* thread)
```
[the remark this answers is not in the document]
paulcallen: I was going to keep this similar to thread.c, but for process type functions

ext2/ext2.c
```diff
@@ -4768,27 +4773,15 @@ static int _ext2_dup(
...
 
     if (!_ext2_valid(ext2) || !_file_valid(file) || !file_out)
         ERAISE(-EINVAL);
 
-    /* create the new file object */
-    {
-        if (!(new_file = calloc(1, sizeof(myst_file_t))))
-            ERAISE(-ENOMEM);
-
-        *new_file = *file;
```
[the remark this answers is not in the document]
paulcallen: oh. That is tricky. That is going to need to restructure myst_file_t and all usage of it, having a shared part of the fields, with the necessary flags only present in the outer wrapper, and shared part being refcounted

include/myst/thread.h
```diff
@@ -102,6 +102,9 @@ struct myst_thread
     /* The exit status passed to SYS_exit */
     int exit_status;
 
+    /* Terminating signal value */
+    unsigned signal_termination;
```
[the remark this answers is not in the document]
paulcallen: done

include/myst/thread.h
```diff
@@ -169,6 +173,17 @@ struct myst_thread
         mode_t umask;
         myst_spinlock_t umask_lock;
 
+        /* the process group ID */
+        pid_t pgid;
+
+#define MYST_MAX_MUNNAP_ON_EXIT 5
```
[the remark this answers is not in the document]
paulcallen: done

ext2/ext2.c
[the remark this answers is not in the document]
paulcallen: I think we need to defer that change until after this pull request goes in. Then we rework the file handle for each of the filesystems.

crt/Makefile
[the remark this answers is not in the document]
paulcallen: done

crt/fork.c
paulcallen: renamed

crt/fork.c
```diff
@@ -0,0 +1,569 @@
...
+
+extern hidden void* __copy_tls(unsigned char*);
+
+static void _get_parent_stack(void** stack, size_t* stack_size)
+{
+    struct pthread* self = __pthread_self();
+
+    if (self->stack && self->stack_size)
+    {
+        *stack = (uint8_t*)self->stack - self->stack_size;
```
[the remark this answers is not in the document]
jxyang: This function actually gets the stack of `self`. The call site just uses this function to get the stack of current thread before it spawns a child.

crt/fork.c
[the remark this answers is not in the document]
jxyang: From Mike's comment above, it seems files under `/utils/` are available to CRT, kernel, and enclave.

ext2/ext2.c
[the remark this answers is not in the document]
jxyang: Sounds good.

crt/fork.c
```diff
@@ -0,0 +1,420 @@
...
+        _get_parent_stack(&parent_stack, &parent_stack_size);
+
+        if (!(child_pthread =
+                  _create_child_pthread(parent_stack, parent_stack_size)))
+            return -ENOMEM;
+
+        stack = (uint8_t*)child_pthread->stack - child_pthread->stack_size;
+        stack_size = child_pthread->stack_size;
+
+        _get_parent_stack(&parent_stack, &parent_stack_size);
```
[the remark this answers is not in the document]
jxyang: I meant we just called `_get_parent_stack` a few lines above (line 344). Does the size actually change between those two lines?

kernel/poll.c
```diff
@@ -142,38 +150,72 @@ static long _syscall_poll(struct pollfd* fds, nfds_t nfds, int timeout)
         }
     }
 
-    /* pre-poll for kernel events */
+    myst_syscall_clock_gettime(CLOCK_MONOTONIC, &start);
+    if ((original_timeout > 500) || (original_timeout < 0))
+        timeout = 500;
```
jxyang, line 155: Where is 500 coming from? Is this a Linux convention?

kernel/signal.c
```diff
@@ -213,9 +246,20 @@ static long _handle_one_signal(
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wstack-usage="
+
+int myst_signal_has_active_signals(myst_thread_t* thread)
+{
+    uint64_t unblocked = (~thread->signal.mask) |
+                         ((uint64_t)1 << (SIGKILL - 1)) |
+                         ((uint64_t)1 << (SIGSTOP - 1));
```
jxyang, line 254: Seems `((uint64_t)1 << (SIGKILL - 1)) | ((uint64_t)1 << (SIGSTOP - 1))` is used a lot. Maybe worth a `#define`

kernel/poll.c
paulcallen: it is a time that I thought would be acceptable!

solutions/dotnet/config.json
```diff
@@ -4,7 +4,7 @@
     // Whether we are running myst+OE+app in debug mode
     "Debug": 1,
     // The stack size of each ethread
-    "StackMemSize": "400k",
+    "StackMemSize": "800k",
```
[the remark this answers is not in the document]
paulcallen: done

kernel/signal.c
paulcallen: done

kernel/syscall.c
```diff
@@ -3210,19 +3232,37 @@ static long _syscall(void* args_)
...
+            int ret = 0;
 
-            _strace(n, "addr=%p length=%zu", addr, length);
+            _strace(n, "ptr=%p, size=%zu", ptr, size);
 
-            myst_assume(self->unmap_on_exit_addr == NULL);
-            myst_assume(self->unmap_on_exit_length == 0);
+            if (i >= MYST_MAX_MUNNAP_ON_EXIT)
+            {
+                i--;
```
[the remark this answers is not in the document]
paulcallen: good catch

crt/fork.c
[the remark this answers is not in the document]
paulcallen: renamed to _get_current_stack

kernel/thread.c
```diff
@@ -348,40 +410,172 @@ long myst_syscall_wait4(
...
+                        "supported): %s\n",
+                        WIFCONTINUED(*wstatus) ? "true" : "false");
+#endif
+                }
                 ret = p->pid;
+
+                // remove from zombie list
+                if (p->zprev == NULL)
+                {
+                    _zombies_head = p->znext;
```
jxyang, line 472: Consider using "myst_list_t" to manage zombies. We can replace line 470-484 with a simple `myst_list_remove`.

kernel/thread.c
```diff
@@ -348,40 +410,172 @@ long myst_syscall_wait4(
...
+                if (p->znext != NULL)
+                    p->znext->zprev = p->zprev;
+
+                if (_zombies_tail == p)
+                    _zombies_tail = p->group_prev;
+
+                _zombies_count--;
+
+                // remove from process list
+                myst_spin_lock(&myst_process_list_lock);
```
jxyang, line 487: Possible dead lock due to acquiring a second lock while holding `_zombies_mutex`. Shouldn't the process be removed from the process list during zombify?

kernel/thread.c
```diff
@@ -348,40 +410,172 @@ long myst_syscall_wait4(
...
+
+        myst_spin_unlock(&myst_process_list_lock);
+
         if ((options & WNOHANG))
         {
             ret = 0;
             goto done;
         }
 
         /* wait for signal from myst_zombify_thread() */
```
jxyang, line 539: The comment should be removed also. We are not using conditional variables now.

kernel/thread.c
```diff
@@ -581,29 +801,36 @@ static long _run_thread(void* arg_)
...
             procfs_pid_cleanup(thread->pid);
         }
 
-        myst_zombify_thread(thread);
-
         /* Send a SIGCHLD to the parent process */
-        myst_syscall_kill(thread->ppid, SIGCHLD);
+        if (!is_child_thread)
+        {
+            myst_syscall_kill(thread->ppid, SIGCHLD);
```
jxyang, line 824: If this is scoped to a process thread, it should move into the previous `if (!is_child_thread)` branch, i.e., after line 818.

kernel/syscall.c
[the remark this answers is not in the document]
jxyang: We should add the unmap address and length to this array for SYS_unmap when the address is determined within the calling thread's stack range.

kernel/thread.c
```diff
@@ -898,53 +1131,132 @@ size_t myst_get_num_threads(void)
...
+        if ((p->ppid == pid) && (p->clone.flags & CLONE_VFORK))
+        {
+            ret = 1;
+            break;
+        }
+        p = p->main.prev_process_thread;
+    }
+
+    if (!p)
+    {
```
jxyang, line 1154: It would be nice to have an abstraction for operations applied to a process/thread list:

```c
bool predicate(myst_thread_t*);
void action(myst_thread_t*);
map2(myst_thread_t* current_thread, predicate, action);
```
Both myst_have_child_forked_processes and kill_child_fork_processes can be implemented on top of `map2`.

kernel/poll.c
```diff
@@ -142,38 +150,72 @@ static long _syscall_poll(struct pollfd* fds, nfds_t nfds, int timeout)
...
+            break;
+
+        has_signals = myst_signal_has_active_signals(myst_thread_self());
+        if (has_signals)
+        {
+            ret = -EINTR;
+            goto done;
+        }
+
+        if (tevents == -EINTR)
```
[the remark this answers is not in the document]
paulcallen: was leftover from a previous attempt! Removed

kernel/poll.c
```diff
@@ -142,38 +150,72 @@ static long _syscall_poll(struct pollfd* fds, nfds_t nfds, int timeout)
...
-        ECHECK((tevents = myst_tcall_poll(NULL, tnfds, timeout)));
-    }
+        /* poll for target events */
+        if (tnfds && tfds)
+        {
+            tevents = myst_tcall_poll(tfds, tnfds, timeout);
+        }
+        else
+        {
+            tevents = myst_tcall_poll(NULL, tnfds, timeout);
```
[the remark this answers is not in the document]
paulcallen: there is a wake_all_polls tcall that gets called and this catches that. As it waits on that kernel object, it acts as a sleep, but interruptible by the wake.

kernel/thread.c
```diff
@@ -259,53 +248,121 @@ void myst_zombify_thread(myst_thread_t* thread)
         myst_munmap(thread->unmapself_addr, thread->unmapself_length);
```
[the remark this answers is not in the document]
paulcallen: done

kernel/syscall.c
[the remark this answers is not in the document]
paulcallen: done

kernel/thread.c
paulcallen: we can consider that for a future update.

kernel/thread.c
paulcallen: hmm... not sure I like this type of approach

kernel/thread.c
paulcallen: The impact at this point is unknown for changing the lists threads are in. Moving to use the same lock for zombies and processes. Moving an item from one list to the other with two locks has the possibility of another thread thinking the process thread does not exist.

kernel/thread.c
paulcallen: done

kernel/thread.c
paulcallen: done

kernel/thread.c
jxyang: It's a way to avoid repeating writing tedious and error-prone code to traverse a doubly linked list.

jxyang:
Please create issues for the comments that isn't addressed yet. I don't have more requests.

@mikbras  It's your turn now.

crt/enter.c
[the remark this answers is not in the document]
mikbras: We implemented fork in the kernel because there is no way to create a thread inside the kernel today and implementing it would be very complex. So we implement fork inside the CRT by calling ``clone``, which enters the kernel. This is how ``posix_spawn`` is implemented in musl libc (it calls clone and all synchronization occurs in the CRT).

crt/exit.c
```diff
@@ -0,0 +1,71 @@
...
+ * All global variables the CRT uses are shared between the two processes.
+ * If either process shuts down through exit() the CRT will clean up the process
+ * by deleting memory, deallocating locks and various other things. In exit() we
+ * call into the kernel to determine if we are a parent of a fork or a child
+ * fork process. If we are a child fork we skip the cleanup and let the parent
+ * process do that. If we are the parent process of forks we will terminate all
+ * child forks, wait for them to exit, then proceed to cleanup If we are neither
+ * a child forked process or parent of one we cleanup as normal.
+ */
+void exit(int code)
```
mikbras, line 21: Is this a strong form of ``exit`` where there is a weak from someplace. If so, can you add a short comment?

crt/fork.c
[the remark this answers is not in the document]
mikbras: I suggest moving the body of ``myst_round_up`` into ``<myst/round.h>``. I overserved some performance problems with calling this function over and over in other places.

crt/fork.c
```diff
@@ -0,0 +1,427 @@
...
+    if (__builtin_add_overflow(x, m - 1, &t))
+        return -ERANGE;
+
+    if (__builtin_mul_overflow(t / m, m, r))
+        return -ERANGE;
+
+    return 0;
+}
+
+/* ATTN: does not work for Linux yet: SYS_arch_prctl */
```
mikbras, line 47: Is the comment correct? (that it does not work on Linux yet)? If so, we can copy and modify this function from ``kernel/fsbase.c``.

```C++
void myst_set_fsbase(void* p)
{
    if (__options.have_syscall_instruction)
    {
        const long n = SYS_arch_prctl;
        myst_syscall2(n, ARCH_SET_FS, (long)p);
    }
    else
    {
        __asm__ volatile("wrfsbase %0" ::"r"(p));
    }
}
```

We will need a new system call to tell us that it is safe to use the **SYSCALL** instruction (e.g., ``SYS_have_syscall_instruction``).

mikbras: Also we should discuss with @vtikoo who is working on ``arch_prctl`` to add ``SET_FSBASE`` possible in the future.

crt/fork.c
```diff
@@ -0,0 +1,427 @@
...
+static void _set_fsbase(void* p)
+{
+    __asm__ volatile("wrfsbase %0" ::"r"(p));
+}
+
+extern hidden unsigned __default_guardsize;
+extern hidden unsigned __default_stacksize;
+extern hidden volatile size_t __pthread_tsd_size;
```
mikbras, line 55: These three symbols are declared in ``./musl/src/internal/pthread_impl.h``. Is there some way to include it?

mikbras (commented):
Thank you so much for doing this Paul. Great effort. I will start a continuing review tomorrow morning.

crt/enter.c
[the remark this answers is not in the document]
jxyang: We can make ocall to create threads. I think we just need to copy some code  in `pthread_create` to ensure the thread data conforms to what MUSL expects.

crt/fork.c
vtikoo: Can `SYS_set_thread_area` be used here? Thats the syscall the musl loader is calling for setting fsbase for the main thread.

crt/Makefile
```diff
@@ -11,17 +11,30 @@ endif
 
 CFLAGS += $(OPTIMIZATION_CFLAGS)
 
-INCLUDES = -I$(INCDIR) -I$(TOP)/build/musl/include
+MUSLSRC=$(TOP)/third_party/musl/crt/musl
+
+INCLUDES = -I$(INCDIR) 
+INCLUDES += -I$(TOP)/build/musl/include
+INCLUDES += -I$(MUSLSRC)/src/internal
```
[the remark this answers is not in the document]
paulcallen: I cannot. It is used to pull in the pthread structures

.azure_pipelines/ci-pipeline-makefile.yml
```diff
@@ -25,6 +25,7 @@ pr:
 jobs:
   - job: BuildAndTest
     displayName: 'Building and Testing'
+    timeoutInMinutes: 60
```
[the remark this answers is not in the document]
paulcallen: 60 is default. I bumped it up at one point, but it does seem to finish with enough time to spare. We can bump later if we do run out of time.

crt/Makefile
```diff
@@ -11,17 +11,30 @@ endif
...
+INCLUDES += -I$(MUSLSRC)/arch/x86_64/
```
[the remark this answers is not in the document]
paulcallen: I cannot. It is used to pull in the pthread structures

crt/exit.c
paulcallen: yes, this is the strong version  that overwrites the CRT weak version

crt/fork.c
paulcallen: We do actually include that header now. I'll delete

crt/fork.c
paulcallen: you added this comment :)
that said I think we are going to have that syscall, but only on a read-basis. Therefore a have_syscall type call may be problematic if we only support part.

crt/Makefile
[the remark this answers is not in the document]
paulcallen: actually, this one can go. Removed.

crt/Makefile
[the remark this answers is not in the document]
mikbras: :)

crt/enter.c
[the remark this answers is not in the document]
mikbras: The must libc pthread structures are opaque to the kernel. We will need some design discussions later.

kernel/fdtable.c
```diff
@@ -178,6 +178,29 @@ int myst_fdtable_free(myst_fdtable_t* fdtable)
...
+
+    if (!fdtable)
+        ERAISE(-EINVAL);
+
+    /* Close all objects */
+    for (int i = 0; i < MYST_FDTABLE_SIZE; i++)
+    {
+        myst_fdtable_entry_t* entry = &fdtable->entries[i];
+
+        if (entry->type == MYST_FDTABLE_TYPE_PIPE)
```
mikbras, line 193: Eventfds may be interruptible too.

kernel/pipedev.c
```diff
@@ -592,6 +594,24 @@ static int _pd_dup(
     return ret;
 }
 
+static int _pd_interrupt(myst_pipedev_t* pipedev, myst_pipe_t* pipe)
```
mikbras, line 597: All types need this methods, whether they can be interrupted or not.

mikbras (changes requested):
Looks great. Thank you for doing all this. 

Just one request. Can you please see if kernel stack size max can be restored? This will really impact startup. If it crashes with this enabled, I can help figure out why.

```
#define MYST_KSTACK_SIZE (128 * 1024)
```

include/myst/kstack.h
```diff
@@ -9,7 +9,7 @@
 #include <myst/defs.h>
 
 #define MYST_MAX_KSTACKS 1024
-#define MYST_KSTACK_SIZE (64 * 1024)
+#define MYST_KSTACK_SIZE (128 * 1024)
```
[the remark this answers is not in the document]
paulcallen: reverted

tests/fork/fork.c
```diff
@@ -0,0 +1,198 @@
...
+    if (argc < 2)
+    {
+        fprintf(stderr, "Not enough args... which test?\n");
+        return -1;
+    }
+    if (strcmp(argv[1], "fork") == 0)
+    {
+        assert(test_fork1(argc, argv) == 0);
+        assert(test_fork2(argc, argv) == 0);
+        assert(test_fork3(argc, argv) == 0);
```
jxyang, line 195: Can we add a test case that print out the stack in the child? And ensure the write to the stack by the child is invisible to the parent, and vice versa?

paulcallen: added 2 tests.
* the child overwrites a variable on the stack and it is unchanged in the parent. This proves the child and parent stacks are different.
* added a fork/exec test

jxyang (approved):

mikbras (approved):

kernel/signal.c
```diff
@@ -106,12 +109,42 @@ long myst_signal_sigprocmask(int how, const sigset_t* set, sigset_t* oldset)
     return ret;
 }
 
+void myst_signal_free_siginfos(myst_thread_t* thread)
+{
+    uint64_t unblocked = MYST_SIG_UNBLOCKED(thread->signal.mask);
+    uint64_t active_signals = thread->signal.pending & unblocked;
```
vtikoo, line 115: @paulcallen There might be siginfos related to blocked signals. #420 is such a case.

paulcallen: I am wondering if there are some race conditions that add after we finish freeing or something. I got a bunch of the cleanup where it was not so good before, but there seemed to be occasional times there was still one left behind.

paulcallen: oh, blocked code came in after I implemented most of this! Now I understand!

PR #428, opened 2021-06-02
EXPERIMENTAL ONLY: mystikos.v4

mikbras:
This PR is a test of the new openenclave/mystikos.v4 branch, which upgrades to the latest official version of openenclave.


mingweishih:
It would be good to rebase against the latest OE master, which supports the #PF/#GP information delivery for SGX2 (just got merge yesterday).

third_party/openenclave/Makefile
```diff
@@ -6,7 +6,7 @@ BUILDDIR=${TOP}/build
 
 URL=https://github.com/openenclave/openenclave
 
-BRANCH=mystikos.v2
+BRANCH=mystikos.v4
```
mingweishih, line 9: Removing the use of submodule means that any future changes to the OE mystikos branch will be automatically picked up by the mystikos build. In other word, there will be no way to lock down the OE side change before mystikos uses it unless we prevent any further update to the branch (say `mystikos.v4`). Any update on the OE will only be applied when mystikos migrate to another branch (say `mystikos.v5`). Is this the plan?

jxyang: I think we can fix to a certain commit hash if wanted. Note the `HASH` variable in the Makefile.

mingweishih: Hard-code the commit should work although it sounds like mimic the submodule behavior. I don't have strong opinion on either option.

mikbras: Working with submodules has been unmanageable so we decided to experiment with working directly with a branch. This way, we can switch between branches at will. With submodules, we would see changes to the submodule but not the branch so that the branch and latest hash were inconsistent. 

Our general approach is to maintain a branch and create PR's against that branch.

mingweishih: How can we test the PR to the mystik OE branch against the mystikos?
Ideally we want:

Submit a PR to the mystikos-oe branch -> Test the mystikos against the updated mystikos-oe branch -> Merge the PR.

With the submodule, the process looks like:

Submit a PR to the mystikos-oe branch -> Merge the PR into the mystikos-oe branch -> Submit a PR to mystikos that updates the submodule -> Test the mystikos against the updated mystikos-oe branch -> Merge the PR

This can be mimic with @jxyang's suggestion: (Hard-code the hash to the commit in mystikos)

Submit a PR to the mystikos-oe branch -> Merge the PR into the mystikos-oe branch -> Submit a PR to mystikos that updates the hash -> Test the mystikos against the updated mystikos-oe branch -> Merge the PR

mikbras:
> It would be good to rebase against the latest OE master, which supports the #PF/#GP information delivery for SGX2 (just got merge yesterday).

It would be interesting to see what works and what doesn't but generally we have to depend on stable releases rather than experimental development branch.

mingweishih:
> > It would be good to rebase against the latest OE master, which supports the #PF/#GP information delivery for SGX2 (just got merge yesterday).
> 
> It would be interesting to see what works and what doesn't but generally we have to depend on stable releases rather than experimental development branch.

Currently only this PR is needed: https://github.com/openenclave/openenclave/pull/4068.

mikbras:
This PR is breaking the pipeline because it uses the latest OE release, whose ansible scripts mess up the pipeline systems for other PRs that use the old OE releases, that use older ansible scripts. We are going to have to build OE in a container, so until then I am closing this PR.

PR #432, opened 2021-06-04
Sanitize reporting of xsave buffer area

vtikoo:

### DIFF
--- a/tools/myst/enc/enc.c
+++ b/tools/myst/enc/enc.c
@@ -156,12 +156,15 @@ int myst_setup_clock(struct clock_ctrl*);
 static void _sanitize_xsave_area_fields(uint64_t* rbx, uint64_t* rcx)
 {
     assert(rbx && rcx);
-    /* replace XSAVE/XRSTOR save area size with fixed large value,
+    /* replace XSAVE/XRSTOR save area size with fixed large value of 4096,
     to protect against spoofing attacks from untrusted host.
     If host returns smaller xsave area than required, this can cause a buffer
     overflow at context switch time.
-    We believe value of 4096 should be sufficient for forseeable future */
-    *rbx = *rcx = 4096;
+    We believe value of 4096 should be sufficient for forseeable future. */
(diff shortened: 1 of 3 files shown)


tools/myst/enc/enc.c
```diff
@@ -153,6 +153,17 @@ static size_t _get_num_tcs(void)
...
 
+static void _sanitize_xsave_area_fields(uint64_t* rbx, uint64_t* rcx)
+{
+    assert(rbx && rcx);
+    /* replace XSAVE/XRSTOR save area size with fixed large value,
+    to protect against spoofing attacks from untrusted host.
+    If host returns smaller xsave area than required, this can cause a buffer
+    overflow at context switch time.
+    We believe value of 4096 should be sufficient for forseeable future */
+    *rbx = *rcx = 4096;
```
bodzhang, line 164: Add sanity check to only override rbx, rcx, if the original value is less than 4096. This way, in case the code runs on some future CPU that reruns larger value, the implementation itself will not override the buffer size to be a value smaller than expected.

bodzhang (changes requested):

tools/myst/enc/enc.c
vtikoo: Fixed.

bodzhang (approved):

tools/myst/enc/enc.c
```diff
@@ -153,18 +156,102 @@ static size_t _get_num_tcs(void)
...
+
+    return "unknown";
+}
+
+__attribute__((format(printf, 2, 3))) static void _exception_handler_strace(
+    long n,
+    const char* fmt,
+    ...)
+{
+    if (_trace_syscalls)
```
bodzhang, line 211: The release_build PR tries to consolidate strace switches. @mikbras , please review this code.

vtikoo: Vectored exception handler tracing is now consistent with syscall tracing, its turned off if `tee_debug_mode` is not set.

PR #436, opened 2021-06-08
swallow userdel errors

salsal97:


tests/ltp/Makefile
```diff
@@ -128,12 +128,12 @@ add_host_users_and_groups:
...
 del_host_users_and_groups:
-	sudo userdel user700
-	sudo userdel user702
-	sudo userdel user703
-	sudo groupdel grp701
-	sudo groupdel grp704
-	sudo groupdel grp705
+	sudo userdel user700 || /bin/true
+	sudo userdel user702 || /bin/true
+	sudo userdel user703 || /bin/true
```
jxyang, line 133: Can we just use `sudo userdel user700 || true`?

jxyang (approved):

tests/ltp/Makefile
salsal97: Hi Xuejun, would it ever act differently?

jxyang: I think it's the same. We have a few Makefiles use `... || true`

salsal97: Gotcha, if they act the same I think its okay to keep this here, what do you think?

jxyang: Yes since this is a blocking issue. Feel free to merge after the pipeline passes.

jxyang:
I have a more complete pipeline fix PR #437

salsal97:
Thanks Xuejun, I'll close this one

PR #448, opened 2021-06-09
SOS debugger extension support

vtikoo:
### Summary
- Preserve file names when copying symbol files to host. SOS looks for the libcoreclr module by name.
- Add cmdline option to report host thread ids for SYS_gettid.
- Partial support for /proc/[pid]/status.
- Patch addresses in mmapped dynamic section in the musl loader.
- Test for dotnet sos integration.

### DIFF
--- a/.azure_pipelines/ci-pipeline-makefile.yml
+++ b/.azure_pipelines/ci-pipeline-makefile.yml
@@ -48,7 +48,7 @@ jobs:
           # remove all untracked files and directories in the git repository
           sudo rm -rf `git ls-files --others --directory`
           make distclean
-          sudo rm -rf third_party/openenclave/openenclave
+          make distclean -C third_party/openenclave/
           make -j
         displayName: 'build repo source'
         workingDirectory: $(Build.SourcesDirectory)
--- a/.azure_pipelines/ci-pipeline-makefile.yml
+++ b/.azure_pipelines/ci-pipeline-makefile.yml
@@ -48,7 +48,7 @@ jobs:
           # remove all untracked files and directories in the git repository
           sudo rm -rf `git ls-files --others --directory`
           make distclean
-          make distclean -C third_party/openenclave/
+          sudo rm -rf third_party/openenclave/openenclave
           make -j
         displayName: 'build repo source'
         workingDirectory: $(Build.SourcesDirectory)
--- a/.azure_pipelines/ci-pipeline-makefile.yml
+++ b/.azure_pipelines/ci-pipeline-makefile.yml
@@ -39,7 +39,7 @@ jobs:
           sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
           sudo apt-get update
           while sudo lsof /var/lib/dpkg/lock-frontend | grep dpkg; do sleep 3; done
-          sudo apt-get install build-essential python3-setuptools libmbedtls-dev docker-ce -y 
+          sudo apt-get install build-essential python3-setuptools libmbedtls-dev docker-ce lldb-10 -y
           sudo chmod 666 /var/run/docker.sock
         displayName: 'minimum init config'
 
--- a/.azure_pipelines/ci-pipeline-release.yml
+++ b/.azure_pipelines/ci-pipeline-release.yml
@@ -30,7 +30,7 @@ jobs:
           curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
           sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
           sudo apt-get update
-          sudo apt-get install build-essential python3-setuptools libmbedtls-dev docker-ce -y 
+          sudo apt-get install build-essential python3-setuptools libmbedtls-dev docker-ce lldb-10 -y 
           sudo chmod 666 /var/run/docker.sock
         displayName: 'minimum init config'
 
--- a/kernel/procfs.c
+++ b/kernel/procfs.c
@@ -1,5 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
+
+#include <assert.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
@@ -13,6 +15,9 @@
 #include <myst/printf.h>
 #include <myst/process.h>
 #include <myst/procfs.h>
+#include <myst/strings.h>
+
+static int _status_vcallback(myst_buf_t* vbuf);
 
 static myst_fs_t* _procfs;
 static char* _cpuinfo_buf = NULL;
@@ -68,6 +73,7 @@ int procfs_pid_setup(pid_t pid)
     {
         char fdpath[PATH_MAX];
         char mapspath[PATH_MAX];
+        char statuspath[PATH_MAX];
     };
     struct locals* locals = NULL;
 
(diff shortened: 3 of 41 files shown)


.azure_pipelines/ci-pipeline-makefile.yml
```diff
@@ -39,7 +39,8 @@ jobs:
           sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
           sudo apt-get update
           while sudo lsof /var/lib/dpkg/lock-frontend | grep dpkg; do sleep 3; done
-          sudo apt-get install build-essential python3-setuptools libmbedtls-dev docker-ce -y 
+          sudo apt-get install build-essential python3-setuptools libmbedtls-dev docker-ce lldb-8 -y
+          sudo ln -s /usr/lib/llvm-8/bin/lldb /usr/bin/lldb
```
anakrish, line 43: The /usr/bin/lldb symlink is not needed. oelldb figures out the right lldb executable:
https://github.com/openenclave/openenclave/blob/ee0032655e5e8b6f847cd2a3a0008d1464254e3c/debugger/oelldb#L25-L32

.azure_pipelines/ci-pipeline-release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
           sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
           sudo apt-get update
-          sudo apt-get install build-essential python3-setuptools libmbedtls-dev docker-ce -y 
+          sudo apt-get install build-essential python3-setuptools libmbedtls-dev docker-ce lldb-8 -y
```
anakrish, line 33: If available, use lldb-10 since OE has switched to clang-10 as the compiler.

kernel/procfs.c
```diff
@@ -225,6 +245,132 @@ static int _cpuinfo_vcallback(myst_buf_t* vbuf)
...
+        locals->curr_thread->gid,
+        locals->curr_thread->egid,
+        locals->curr_thread->savgid,
+        locals->curr_thread->fsgid);
+    ECHECK(myst_buf_append(vbuf, tmp, strlen(tmp)));
+
+    /* TODO: memory, signal, capability and cpu related fields*/
+
+done:
+    if (locals->_host_status_buf)
```
anakrish, line 365: locals can be NULL here

tests/dotnet-sos/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+```
+make
+make run-ext2-lldb
+```
+
+
+Dotnet runtime issues a lot of cpuid instructions, which generate a SIGILL inside the enclave. To handle these run after the mystikos process is launched -
+```
+pro hand -p true -s false -n false SIGILL
```
anakrish, line 31: Try to see if we can do this automatically as part of `oelldb` initialization.

It would be ideal if we can automatically ignore only the CPUID related SIGILL.

tests/dotnet-sos/sos_test.py
```diff
@@ -0,0 +1,58 @@
...
+    assert(res.Succeeded())
+    process.Continue()
+
+    # Should've hit managed breakpoint
+    assert(process.GetState() == lldb.eStateStopped)
+    assert(process.GetSelectedThread().GetStopReason() == lldb.eStopReasonBreakpoint)
+
+    print("\nRun clrstack at breakpoint hit: ")
+    print("=================================")
+    ci.HandleCommand("clrstack", res)
```
anakrish, line 35: If we can capture the output of HandleCommand, then we can add assertions for frames

anakrish (commented):
Looks great!

kernel/procfs.c
vtikoo: Thanks for catching this!

tests/dotnet-sos/sos_test.py
vtikoo: For SOS commands, couldn't capture output via `res.GetOutput`, the command output goes directly to stdout. I have added grep assertions on the stdout as an alternative.

tests/dotnet-sos/README.md
vtikoo: `pro hand` way of handling SIGILL requires the debuggee process to exist, so can't do it oelldb init time.
Have added a python script `ignore_sigill.py` which does the equivalent and can be imported at init. However, it ignores SIGILL completely, couldn't find a way to make it selective.

tests/dotnet-sos/README.md
```diff
@@ -0,0 +1,52 @@
+### Debugging managed dotnet code
+
+## Prerequisites
+1. lldb:
+```
+sudo apt install lldb-8
```
anakrish, line 6:
```suggestion
sudo apt install lldb-10
```

kernel/procfs.c
```diff
@@ -225,6 +245,132 @@ static int _cpuinfo_vcallback(myst_buf_t* vbuf)
...
+        ERAISE(-ENOMEM);
+
+    ECHECK(myst_tcall_read_file(
+        locals->status_path, locals->_host_status_buf, size));
+    locals->_host_status_buf[size] = 0;
+
+    myst_buf_clear(vbuf);
+    char tmp[128];
+
+    snprintf(tmp, sizeof(tmp), "Name:\t%s\n", locals->curr_thread->name);
```
jxyang, line 318: Check the return value to see if there is overflow. Do this for all `snprintf`.

tests/dotnet-sos/README.md
```diff
@@ -0,0 +1,52 @@
...
+2. dotnet-sos:
+```
+dotnet tool install -g dotnet-sos
+# Installs the SOS extension, and
+# updates .lldbinit file to load SOS on lldb startup
+dotnet-sos install
+```
+3. oelldb, SGX enclave extension for lldb:
+```
+git clone --branch working https://github.com/vtikoo/oelldb
```
jxyang, line 17: Is there a plan to merge oelldb into OE master or Mystikos's OE branch?

third_party/musl/crt/patch.diff
```diff
@@ -211,7 +211,53 @@ index afec985a..5435b9b5 100644
  #define MAXP2(a,b) (-(-(a)&-(b)))
  #define ALIGN(x,y) ((x)+(y)-1 & -(y))
  
-@@ -1043,38 +1050,7 @@ static struct dso *load_library(const char *name, struct dso *needed_by)
+@@ -217,6 +224,14 @@ static int search_vec(size_t *v, size_t *r, size_t key)
+ 	return 1;
+ }
+ 
++static size_t* search_vec2(size_t *v, size_t *r, size_t key)
```
jxyang, line 218: Suggest renaming to `find_dynv_value_addr_by_key`

tests/dotnet-sos/README.md
vtikoo: The lldb extension is in OE master, but not in any release branch yet. Using the OE master version would also require mystikos to adopt the multi module debugger support in OE which @mikbras is already handling in his OE 0.16 rebase PR. Once thats in, maybe we can pull the lldb extension changes from master into Mystikos's OE branch.

third_party/musl/crt/patch.diff
vtikoo: Renamed.

kernel/procfs.c
vtikoo: Replaced all `snprintf` occurences in this file with `myst_snprintf`. `myst_snprintf` is just a wrapper around the existing snprintf implementation with additional check for overflow.

anakrish (approved):
LGTM. Thanks for making this happen!

It would be good ensure that the debugger related code within the enclave is not executed if the enclave is not debuggable ie signed with Debug = False.

PR #449, opened 2021-06-10
Auto-mount host /etc/resolv.conf

Francis-Liu:
- mirror docker container's behavior, copy host /etc/resolv.conf into each enclave, ramfs
  - if /etc/resolv.conf already exist, delete it
  - if /etc folder doesn't exist, create it with permission 0755
  - if /etc is not a folder, delete it, and create the /etc folder
  - set /etc/resolv.conf permission 0644
- delete corresponding code from appbuilder


kernel/enter.c
```diff
@@ -564,6 +564,27 @@ static int _mount_rootfs(myst_kernel_args_t* args, myst_fstype_t fstype)
     return ret;
 }
 
+#if defined(MYST_ENABLE_HOSTFS)
+#define ETC_RESOLV_CONF "/etc/resolv.conf"
+static int _mount_host_resolv_conf()
```
bodzhang, line 569: Auto mount `/etc/resolv.conf` without opt-in effectively overrides the hardcoded value inserted by `AppBuilder` (also no opt-in). Is that the design intention?

Francis-Liu: yes

kernel/enter.c
```diff
@@ -95,6 +95,83 @@ static int _process_mount_configuration(myst_mounts_config_t* mounts)
...
+        {
+            if ((myst_mkdirhier("/etc", 0755)) != 0)
+            {
+                myst_eprintf("kernel: failed to mkdir /etc\n");
+                ERAISE(-EINVAL);
+            }
+        }
+        else if (!S_ISDIR(statbuf.st_mode))
+        {
+            if ((myst_syscall_unlink("/etc")) < 0)
```
jxyang, line 141: Same here. Error out if /etc is not a folder.

kernel/enter.c
```diff
@@ -95,6 +95,83 @@ static int _process_mount_configuration(myst_mounts_config_t* mounts)
...
+                ERAISE(-EINVAL);
+            }
+        }
+    }
+    if ((fd = open(resolv_file, O_WRONLY | O_CREAT | O_TRUNC, 0644)) < 0)
+    {
+        myst_eprintf("kernel: failed to open file %s\n", resolv_file);
+        ERAISE(-EINVAL);
+    }
+    if ((myst_writen(fd, buf, size)) < 0)
```
jxyang, line 158: Rename the function to `myst_write_file_fd`.

tests/curl/Makefile
```diff
@@ -25,10 +18,10 @@ all:
 	$(MAKE) rootfs
 
 tests:
-	$(RUNTEST) $(MYST_EXEC) $(OPTS) --mount /etc/resolv.conf=/etc/resolv.conf rootfs --app-config-path config1.json /curl https://www.microsoft.com 2>&1 > /dev/null
+	$(RUNTEST) $(MYST_EXEC) $(OPTS) rootfs --app-config-path config1.json /curl https://www.microsoft.com 2>&1 > /dev/null
```
jxyang, line 21: I don't think we need `--app-config-path config1.json` anymore.

tests/myst/auto-mount/main.c
```diff
@@ -34,35 +34,18 @@ int test_auto_mount(int argc, const char* argv[])
...
+    assert(stat("/run/mystikos/automounts/1/testfile1", &details) == 0);
     assert(S_ISREG(details.st_mode));
     assert(stat("/targetfile2", &details) == 0);
     assert(S_ISREG(details.st_mode));
-    assert(stat("/run/mystikos/automounts/2/testfile2", &details) == 0);
+    assert(stat("/run/mystikos/automounts/1/testfile2", &details) == 0);
     assert(S_ISREG(details.st_mode));
 
-    fp = fopen("/etc/resolv.conf", "r");
-    assert(fp != NULL);
```
jxyang, line 56: We should still verify that `/etc/resolv.conf` is there, but not necessarily the content.

kernel/enter.c
```diff
@@ -709,6 +786,8 @@ int myst_enter_kernel(myst_kernel_args_t* args)
 
     ECHECK(_process_mount_configuration(args->mounts));
 
+    ECHECK(_copy_host_etc_resolv());
```
jxyang, line 789: Does this also cover ext2fs?

kernel/enter.c
Francis-Liu: Changed.

kernel/enter.c
```diff
@@ -95,6 +95,83 @@ static int _process_mount_configuration(myst_mounts_config_t* mounts)
     return ret;
 }
 
+static int _copy_host_etc_resolv()
+{
```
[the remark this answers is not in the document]
Francis-Liu: Changed to *_files

kernel/enter.c
```diff
@@ -95,6 +95,83 @@ static int _process_mount_configuration(myst_mounts_config_t* mounts)
...
+        {
+            myst_eprintf("kernel: failed to unlink file %s\n", resolv_file);
+            ERAISE(-EINVAL);
+        }
+    }
+    else
+    {
+        if (stat("/etc", &statbuf) == -1)
+        {
+            if ((myst_mkdirhier("/etc", 0755)) != 0)
```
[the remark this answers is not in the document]
Francis-Liu: It turns out this is the only place we manipulate /etc.

kernel/enter.c
Francis-Liu: Changed.

tests/curl/Makefile
Francis-Liu: Right, changed.

tests/myst/auto-mount/main.c
Francis-Liu: Changed.

kernel/enter.c
```diff
@@ -95,6 +95,83 @@ static int _process_mount_configuration(myst_mounts_config_t* mounts)
...
+                ERAISE(-EINVAL);
+            }
+            if ((myst_mkdirhier("/etc", 0755)) != 0)
+            {
+                myst_eprintf("kernel: failed to mkdir /etc\n");
+                ERAISE(-EINVAL);
+            }
+        }
+    }
+    if ((fd = open(resolv_file, O_WRONLY | O_CREAT | O_TRUNC, 0644)) < 0)
```
vtikoo, line 153: nit: this is equivalent to  `creat(resolv_file, 0644)`

kernel/enter.c
```diff
@@ -95,6 +95,83 @@ static int _process_mount_configuration(myst_mounts_config_t* mounts)
...
+
+    if (!(buf = malloc(size + 1)))
+        ERAISE(-ENOMEM);
+
+    if ((myst_tcall_read_file(resolv_file, buf, size)) < 0)
+    {
+        myst_eprintf("kernel: failed to read file %s\n", resolv_file);
+        ERAISE(-EINVAL);
+    }
+    buf[size] = 0;
```
vtikoo, line 119: I had this additional null termination for /proc files copied from the host, as linux doesn't terminate them properly. Not sure if thats the case with /etc/resolv.conf.

kernel/enter.c
Francis-Liu: Added ext2rootfs to test/auto-mount.

kernel/enter.c
Francis-Liu: Good point! Changed.

kernel/enter.c
Francis-Liu: Good point! I thought we need the trailing null to print buf as a string. Changed since we don't need it.

kernel/enter.c
```diff
@@ -95,6 +95,82 @@ static int _process_mount_configuration(myst_mounts_config_t* mounts)
     return ret;
 }
```
bodzhang, line 97: Does the `/etc` directory created need to have execution permission? I prefer least privilege approach.

If '/etc' directory already exist in the CPIO ramfs or the dm_verify ext2fs, is it possible that the code fails to create the `reslov.conf` file in that directory due to permission limitation?

In debug mode, Mystikos supports app rootfs on the hostfs (for testing/debugging purpose). Is the implementation compatible with that configuration?

Francis-Liu: - The execution bit of a directory means one can access (read, write, execute) items within the directory. Without the execution permission of `/etc`, one cannot read `resolv.conf` under.
- If `/etc` directory doesn't have write permission, then creat() will fail. Currently, myst "don't check permissions fully yet", so even if the `/etc` directory doesn't have write permission, we can still creat resolv file under it. Also, in case rootfs already contains `/etc/resolv.conf` and it's readonly, we first unlink it (see https://github.com/deislabs/mystikos/pull/449/files#diff-e71a12ac51b1964e823bd4db8de920dcff6f742035713f1a76a53457977a91daR122). This is because "If you have write + execute permissions on a directory, you can {delete,rename} items living within even if you don't have write perimission on those items. (use sticky bit to prevent this)" https://unix.stackexchange.com/questions/21251/execute-vs-read-bit-how-do-directory-permissions-in-linux-work
- Good catch! The unlink() call failed. This was not exposed in ltp test because its appdir/ doesn't have appdir/etc/resolv.conf
```
$ myst exec-sgx  appdir /curl https://www.microsoft.com
kernel: failed to unlink file /etc/resolv.conf
```

jxyang: When we use hostfs as rootfs, we usually point the root directory to a folder on the host (which should be different from `/`), I think it's OK as long as we give a general warning about the danger of persisted file modification for users of hostfs.

utils/file.c
```diff
@@ -77,27 +77,32 @@ int myst_load_file(const char* path, void** data_out, size_t* size_out)
...
-        ssize_t n = write(fd, p, r);
-
-        if (n == 0)
+        if ((n = write(fd, p, r)) == 0)
             break;
-        else if (n < 0)
-            ERAISE(n);
+
+        if (n < 0)
+            ERAISE((int)-n);
```
jxyang, line 96: Why change from `n` to `-n`?

utils/file.c
```diff
@@ -77,27 +77,32 @@ int myst_load_file(const char* path, void** data_out, size_t* size_out)
...
-ssize_t myst_writen(int fd, const void* data, size_t size)
+int myst_write_file_fd(int fd, const void* data, size_t size)
 {
-    ssize_t ret = 0;
+    int ret = 0;
     const uint8_t* p = (const uint8_t*)data;
     size_t r = size;
+    ssize_t n;
+
+    if (fd < 0 || !data)
```
jxyang, line 87: I think it's ok for data to be NULL, as long as `size` is 0.

utils/file.c
```diff
@@ -77,27 +77,32 @@ int myst_load_file(const char* path, void** data_out, size_t* size_out)
...
+        if (n < 0)
+            ERAISE((int)-n);
 
         p += n;
         r -= (size_t)n;
     }
 
-done:
+    if (r != 0)
+        ERAISE(-EIO);
```
jxyang, line 103: Generally we don't change the error code in the done section.

utils/file.c
Francis-Liu: Changed.

utils/file.c
Francis-Liu: This is above the done section.

utils/file.c
Francis-Liu: Changed.

jxyang (approved):

PR #452, opened 2021-06-10
Run passing openmp tests in /samples

Sahakait:
Manually compile openmp test cases from gcc and run all passing tests with `make` and `make tests`.


samples/openmp_testsuite/Makefile
```diff
@@ -0,0 +1,46 @@
...
+	sudo truncate --size=-4096 rootfs
+
+appdir:
+	docker build -t $(CONTAINER_NAME) . \
+	--build-arg IMAGE_FLAVOR=$(CONTAINER_IMAGE_FLAVOR) \
+	--build-arg IMAGE_VERSION=$(CONTAINER_IMAGE_VERSION) \
+	--build-arg GCC_VERSION=$(GCC_VERSION)
+	$(MAKE) export
+
+export:
```
jxyang, line 26: We have a tool `appbuilder` does exactly the same things.

samples/openmp_testsuite/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN apk add --quiet --no-cache bash build-base dejagnu isl-dev make mpc1-dev mpfr-dev texinfo zlib-dev
+RUN wget -q https://ftp.gnu.org/gnu/gcc/gcc-${GCC_VERSION}/gcc-${GCC_VERSION}.tar.gz && \
+    tar -xzf gcc-${GCC_VERSION}.tar.gz && \
+    rm -f gcc-${GCC_VERSION}.tar.gz
+
+RUN echo '----GCC compiling----'
+WORKDIR /gcc-${GCC_VERSION}/libgomp/testsuite
+COPY tests.mak .
+RUN make -f tests.mak
+
```
jxyang, line 16: Use a multi-stage docker file to minimize the docker image.

samples/openmp_testsuite/Makefile
```diff
@@ -0,0 +1,46 @@
...
+
+define NL
+
+
+endef
+
+OPTS = --memory-size=32m
+
+tests:
+	$(foreach i, $(TESTS), $(MYST_EXEC) rootfs $(i) $(OPTS) $(NL) )
```
jxyang, line 43: This is inefficient. Please see [tests/libc](https://github.com/deislabs/mystikos/tree/main/tests/libc) for possible optimization.

jxyang (changes requested):

jxyang:
I prefer this to be added under `tests` instead of `samples`.

samples/openmp_testsuite/Makefile
Sahakait: Changed custom build logic to use  `appbuilder` instead.

samples/openmp_testsuite/Makefile
Sahakait: Test cases currently fail when using the `run.c` file linked, will keep this as a sample for now.

samples/openmp_testsuite/Dockerfile
Sahakait: Changed to multi-stage docker file.

jxyang: The `repo` stage is unnecessary. But we do need another stage after building the test suite. So the final image would contain the test executables but not `build-base` and other packages required for build.

samples/openmp_testsuite/Makefile
```diff
@@ -2,43 +2,38 @@ TOP=$(abspath ../..)
...
 
-export:
-	docker stop $(ID)
-	docker export $(ID) -o image.tar
-	rm -rf appdir
-	mkdir -p appdir
-	tar xf image.tar -C appdir
+$(ROOTFS):
+	sudo $(MYST) mkext2 --force --size $(EXT2FS_SIZE) $(APPDIR) $(ROOTFS)
+	sudo chown $(USER).$(USER) $(ROOTFS)
```
jxyang, line 24: Line 24 is not needed anymore.

mkext2 is unnecessary, we can run the tests as ramfs. Check for `mkcpio` in other samples.

samples/openmp_testsuite/Makefile
jxyang: The optimization is blocked by #498. Drop the request.

samples/openmp_testsuite/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN wget -q https://ftp.gnu.org/gnu/gcc/gcc-7.5.0/gcc-7.5.0.tar.gz && \
+    tar -xzf gcc-7.5.0.tar.gz && \
+    rm -f gcc-7.5.0.tar.gz
+RUN apk add --quiet --no-cache bash build-base dejagnu isl-dev make mpc1-dev mpfr-dev texinfo zlib-dev
+WORKDIR /gcc-7.5.0/libgomp/testsuite
+COPY tests.mak .
+RUN make -f tests.mak
+
+FROM ${IMAGE_FLAVOR}:${IMAGE_VERSION}
+RUN apk add --quiet --no-cache bash build-base
```
jxyang, line 14: Is there a reason why build-base is still needed in the final image?

samples/openmp_testsuite/Makefile
```diff
@@ -0,0 +1,37 @@
...
+$(ROOTFS):
+	sudo $(MYST) mkcpio $(APPDIR) $(ROOTFS)
+
+$(APPDIR):
+	myst-appbuilder Dockerfile
+
+TESTS=$(shell cat $(TEST_FILE))
+
+define NL
+
```
jxyang, line 29: What's line 28-31 for?

samples/openmp_testsuite/Dockerfile
Sahakait: image
Some of the test cases will fail with this error.

samples/openmp_testsuite/Makefile
Sahakait: It adds a new line so this:
image
becomes this:
image

jxyang (approved):

jxyang:
Thanks for the contribution!

PR #454, opened 2021-06-11
remove mmap11 since it fails gcov

salsal97:


salsal97:
Will be going with a different approach for this fix

PR #461, opened 2021-06-15
myst_tcall_wake() fix

mikbras:
Fix incorrect handling of `myst_tcall_wake()`, which returns the number of threads awoken. A caller in one place assumed that the function returns zero or -errno (instead of a positive number or -errno).

Also reverses positive errno returns in (cond and mutex) to negative errnos (to be consistent with the rest of the kernel).


bodzhang (commented):
Futex.c line 370, can it be simply a ECHECK(...)?

paulcallen (approved):

jxyang (approved):

PR #465, opened 2021-06-16
Release mode work

mikbras:
This PR:

- enables release mode (must be disabled again before check-in)
- adds many ``MYST_RELEASE`` checks in make files and source files
- fixes ``__unmap_self`` which crashes due to ``-O3`` optimization
- relaxes stack-usage check in ``kernel/signal.c:_handle_one_signal()``
- removes export-file feature (dead code)
- adds auto detection of debug symbols option (checks ``/proc/self/status:TracerPid``)


kernel/syscall.c
```diff
@@ -3145,6 +3155,7 @@ static long _syscall(void* args_)
         }
         case SYS_myst_add_symbol_file:
         {
+#ifndef MYST_RELEASE
```
bodzhang, line 3158: Just realized that a developer might still want to use myst-gdb to debug the app code inside Mystikos, when using the prebuilt Mystikos binaries (Mystikos "release build"). Should the switch for debug symbol related APIs be based on the debug flag in config.json, instead of the compiling switch? 

Alternatively, we can have two sets of binaries of Mystikos in a release, one for "release build" app, where all debug capability is disabled, one for "debug build" app, where the Mystikos binaries themselves are "debug build".

config.mak
```diff
@@ -5,4 +5,4 @@ MYST_PREFIX=/opt/mystikos
 MYST_ENABLE_HOSTFS=1
 MYST_ENABLE_EXT2FS=1
 #export MYST_MEMCHECK=1
-#export MYST_RELEASE=1
+export MYST_RELEASE=1
```
bodzhang, line 8: Is the intent to build Mystikos as "release build" by default?

kernel/syscall.c
mikbras: Oh good point. Can we use the SGX debug attribute instead? We fetch it on enclave startup like this:

```C++
static int _test_oe_debug_mode(void)
{
    int ret = -1;
    uint8_t* buf = NULL;
    size_t buf_size;
    oe_report_t report;

    if (oe_get_report_v2(0, NULL, 0, NULL, 0, &buf, &buf_size) != OE_OK)
        goto done;

    if (oe_parse_report(buf, buf_size, &report) != OE_OK)
        goto done;

    if (!(report.identity.attributes & OE_REPORT_ATTRIBUTES_DEBUG))
        goto done;

    ret = 0;

done:

    if (buf)
        oe_free_report(buf);

    return ret;
}
```

config.mak
mikbras: No, this should be removed before we merge the PR (this is just for testing the PR in the pipeline).

kernel/syscall.c
bodzhang: Yes, SGX debug attribute decides whether the symbols are accessible by GDB.

mikbras: So is it an acceptable solution?

bodzhang: For developer app debugging support in this PR, the solution is OK. We can discuss the "two sets of binaries" idea in a different thread.

mikbras: Okay, I will allow debugging if the attestation of the SGX properties passes. Also I will add a ``-no-debug-symbols`` option to we can selectively speed up tests, solutions, and demos. :)

mikbras:
The PR now enabled debug only if the SGX debug attribute is set and if so, then it only adds symbols if the process is being debugged.  This is determined with a new ``process_is_being_traced()`` function. See below.

```
// Return 1 if the process is being traced by a debugger, 0 if not being
// traced, and errno on failure. This function searches for the "TracerPid"
// field in the "/proc/<pid>/status" file. The process is being traced if
// the "TracerPid" field has non-zero process id value.
int process_is_being_traced(void);
```

tools/myst/host/process.c
```diff
@@ -0,0 +1,60 @@
...
+int process_is_being_traced(void)
+{
+    int ret = 0;
+    FILE* is = NULL;
+    char line[1024];
+
+    /* open "/proc/<pid>/status" for read */
+    {
+        char path[PATH_MAX];
+        snprintf(path, sizeof(path), "/proc/%u/status", getpid());
```
vtikoo, line 17: nit: can also be `/proc/self/status` instead.

mikbras: Oh, nice simplification, thanks. Fixed.

mikbras:
I removed the no-longer-used export-file feature (was used by old version of gcov).

kernel/syscall.c
mikbras: Finished. We enable debug symbols when two conditions are satisfied:
- The SGX debug attribute is on
- ``/proc/self/status:TracerPid`` is non-zero

kernel/enter.c
```diff
@@ -657,21 +651,26 @@ int myst_enter_kernel(myst_kernel_args_t* args)
...
 
-    /* Save the aguments */
+    /* make a copy of the input kernel aguments and reassign args */
     __myst_kernel_args = *args;
+    args = &__myst_kernel_args;
+
+    /* turn off various options when TEE is not in debug mode */
+    if (!args->tee_debug_mode)
+    {
+        args->trace_errors = false;
```
bodzhang, line 661: `myst_eprintf()` used in a lot of code, but it doesn't seem to be under `trace_errors` control.

tests/Makefile
```diff
@@ -34,7 +34,11 @@ DIRS += json
 DIRS += conf
 DIRS += nbio
 DIRS += thread
+
+ifndef MYST_RELEASE
```
bodzhang, line 38: Now `MYST_RELEASE` build supports debugging if the the TEE-debug-mode is true. Should gdb and dotnet-sos tests be enabled with `MYST_RELEASE` build?

mikbras: Good find. I reenabled those tests now that we have auto-detection of debug mode.

kernel/enter.c
mikbras: This is intentional since users still need to know about certain error conditions in release mode (e.g., failure to mount a file system because the root hash is wrong). I did a pass over all the calls to ``myst_eprintf`` I cannot see any case where secrets would be leak (except for the cases that are already disabled by various trace options).

bodzhang: Saw the latest change to guard backtrace with tee_debug_mode flag. All other existing myst_eprintf() usages don't expose any info about the app.

bodzhang (approved):

PR #467, opened 2021-06-16
Run sos test inside docker and fix dotnet code viewer

vtikoo:
Signed-off-by: Vikas Tikoo <vitikoo@microsoft.com>

PR #470, opened 2021-06-16
Disable dotnet sos test

vtikoo:

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -106,7 +106,7 @@ DIRS += mutex
 DIRS += mprotect
 DIRS += eventfd
 DIRS += polleventfd
-DIRS += dotnet-sos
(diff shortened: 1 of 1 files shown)


jxyang (approved):

anakrish:
@vtikoo In case you are having trouble running oelldb within containers, it could be due to the `personality` syscall not being supported within the compiler. By default,  LLDB tries to disable ASLR using the personality syscall. This can be fixed by
https://github.com/openenclave/openenclave/blob/783fd05984a9baa3202d0cd39983c0ca3be8f68b/tests/debugger/commands.py#L203-L207

vtikoo:
Thank you for the suggestion @anakrish. I had gotten around this by running the container as privileged, but this approach is much better. Have incorporated your fix in #467.

PR #472, opened 2021-06-17
update auto-mount design doc...

paulcallen:
specify that the target mount-point needs to exist

### DIFF
--- a/doc/design/mount-config-design.md
+++ b/doc/design/mount-config-design.md
@@ -58,7 +58,7 @@ mount: [
 
 | Name | Value |
 | -- | -- |
-| target | This is the target mount point within the TEE. |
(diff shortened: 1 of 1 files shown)


doc/design/mount-config-design.md
```diff
@@ -58,7 +58,7 @@ mount: [
 
 | Name | Value |
 | -- | -- |
-| target | This is the target mount point within the TEE. |
+| target | This is the target mount point within the TEE. The target mount point needs to already exist in the target filesystem for the mount to succeed. |
```
jxyang, line 61: This is not clear to me. Do you mean the **path** of the mount point needs to already exist?

paulcallen:
pulling this into another mount-related pull request

PR #483, opened 2021-06-17
Enhance boundary condition check, group the OCALLs into categories.

bodzhang:
- Add boundary condition check for several input controlled by the host-side;
- Add more comments in myst_cond_timedwait();
- Group the OCALLs in the EDL into categories, preparing for future enhancement of excluding OCALL proxy code when the relevant feature is to be excluded in certain build configuration.

Signed-off-by: Bo Zhang (ACC) zhanb@microsoft.com

### DIFF
--- a/tools/myst/enc/enc.c
+++ b/tools/myst/enc/enc.c
@@ -516,7 +516,11 @@ static long _enter(void* arg_)
         /* rootfs buffer content set by the host side. Max length of the string
          * is PATH_MAX-1. Enforce NULL terminator at the end of the buffer.
          */
-        options->rootfs[PATH_MAX - 1] = '\0';
+        if (strnlen(options->rootfs, PATH_MAX) == PATH_MAX)
+        {
+            fprintf(stderr, "rootfs path too long (> %u)\n", PATH_MAX);
+            goto done;
+        }
 
         rootfs = options->rootfs;
     }
--- a/tools/myst/enc/enc.c
+++ b/tools/myst/enc/enc.c
@@ -513,11 +513,10 @@ static long _enter(void* arg_)
         report_native_tids = options->report_native_tids;
         max_affinity_cpus = options->max_affinity_cpus;
 
-        if (strlen(options->rootfs) >= PATH_MAX)
-        {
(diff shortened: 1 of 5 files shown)


tools/myst/enc/enc.c
```diff
@@ -513,11 +513,10 @@ static long _enter(void* arg_)
...
 
-        if (strlen(options->rootfs) >= PATH_MAX)
-        {
-            fprintf(stderr, "rootfs path too long (> %u)\n", PATH_MAX);
-            goto done;
-        }
+        /* rootfs buffer content set by the host side. Max length of the string
+         * is PATH_MAX-1. Enforce NULL terminator at the end of the buffer.
+         */
+        options->rootfs[PATH_MAX - 1] = '\0';
```
mikbras, line 519: It would be nice to retain the error condition above if possible rather than silently truncating the buffer (in the case where the buffer has no null terminators). Maybe this:

```
        if (strnlen(options->rootfs, PATH_MAX) == PATH_MAX)
```

tools/myst/enc/syscall.c
```diff
@@ -819,10 +822,12 @@ static long _poll(struct pollfd* fds, nfds_t nfds, int timeout)
...
     }
 
-    /* copy back the revents field */
+    /* copy back the revents field and sanitize the value provided by the
+     * host-side to be a subset of the events mask.
+     */
     for (nfds_t i = 0; i < nfds; i++)
     {
-        fds[i].revents = copy[i].revents;
+        fds[i].revents = copy[i].revents & fds[i].events;
```
mikbras, line 830: I missed this before: according the the man page "The  bits  returned  in revents can include any of those specified in events, **or one of the values POLLERR, POLLHUP,  or POLLNVAL**.". So maybe:

```
const short int mask = POLLERR | POLLHUP | POLLNVAL;
fds[i].revents = copy[i].revents & (fds[i].events | mask);
```

mikbras:
I added a couple of comments to pushed some minor formatting changes to the .edl file (no functional changes).

mikbras (commented):
Thank you for doing this. Look good but a couple of minor comments.

mikbras (approved):

PR #491, opened 2021-06-21
Fix ltp Makefile useradd user1000

Francis-Liu:
Fixing:
~~grep file pattern~~
grep pattern file

1. call `groupadd -f`

    This option causes the command to simply exit with success status if the specified group already exists.

2. call `id -u`


vtikoo (approved):
Thanks for making this more robust!

PR #495, opened 2021-06-22
Sockperf

mikbras:
@salsal97 and I created this fix for ``./tests/sockperf``, which ensures that the server is listening on port ``11111`` before the clients attempt to connect.


salsal97 (approved):
lgtm

PR #496, opened 2021-06-22
[****HI-PRIORITY****]: Enforce EXT2 roothash in release mode,

mikbras:
Mainly this PR enforces the presence of an EXT2 roothash in release mode, but it also:

- automatically extracts the roothash from the file-system image for the roothash and auto-mount file systems (eliminating the need to specify --roothash, although that option is still needed for application-mounted file systems).
- add the ``--myst-rootfs=<filename>`` option for packaged mode, in addition to the MYST_ROOTFS_PATH environment variable (the option takes precedence).


tests/ext2/crypt/Makefile
```diff
@@ -54,13 +54,10 @@ MKEXT2_OPTS += --force
...
 MKEXT2_OPTS += -p=passphrase12345
 
-OUTFILE=$(SUBOBJDIR)/outfile
-
 $(IMAGE):
 	mkdir -p $(SUBOBJDIR)
 	sudo $(MYST) mkext2 $(MKEXT2_OPTS) ext2dir $(IMAGE)
-	$(MYST) fssig $(IMAGE) > $(OUTFILE)
-	@ grep root_hash= $(OUTFILE) | cut -d= -f 2 > $(ROOTHASH)
+	$(MYST) fssig --roothash $(IMAGE) > $(ROOTHASH)
```
jxyang, line 60: I thought we don't need the roothash in a separate file anymore.

tools/myst/host/package.c
```diff
@@ -603,6 +603,30 @@ int _exec_package(
...
+    /* Get --myst-rootfs=<path> option if any  */
+    {
+        const char* arg = NULL;
+
+        if ((cli_getopt(&argc, argv, "--myst-rootfs", &arg) == 0))
+        {
+            if (access(arg, R_OK) != 0)
+            {
+                fprintf(
+                    stderr, "%s: bad --myst-rootfs option: %s", argv[0], arg);
```
jxyang, line 615: Be more specific on the error.

tools/myst/host/exec.c
```diff
@@ -376,6 +378,31 @@ int exec_action(int argc, const char* argv[], const char* envp[])
 
     const char* rootfs = argv[2];
     const char* program = argv[3];
+    static const size_t extra_roothashes_size = 64;
+    char* extra_roothashes[max_roothashes];
+    ssize_t num_extra_roothashes = 0;
+
+    num_extra_roothashes = create_roothashes(
```
jxyang, line 385: We are not using roothash from the command line anymore, why don't we use the `roothashes` array directly?

jxyang: Why don't we make  `roothashes` an array of root hashes instead of file names containing the root hashes.

tools/myst/host/package.c
```diff
@@ -767,28 +791,31 @@ int _exec_package(
...
-        char* env;
-
-        if (!(env = getenv("MYST_ROOTFS_PATH")))
+        /* if rootfs has not already been specified by --myst-rootfs=<path> */
+        if (*options.rootfs == '\0')
         {
-            fprintf(stderr, "MYST_ROOTFS_PATH is undefined\n");
-            goto done;
-        }
+            char* env;
```
jxyang, line 797: Suggest renaming to "roothash_path"

tools/myst/host/utils.c
```diff
@@ -349,3 +350,148 @@ void free_mount_mapping_opts(myst_mount_mapping_t* mappings)
         free(mappings->mounts);
     }
 }
+
+int create_roothash_file(const char* path, char filename[PATH_MAX])
```
jxyang, line 354: Suggest renaming to `read_roothash_from_ext2_rootfs`. Again, can we skip saving the roothash to a file?

tools/myst/host/utils.c
```diff
@@ -349,3 +350,148 @@ void free_mount_mapping_opts(myst_mount_mapping_t* mappings)
...
+
+            if (!(eq = strchr(path, '=')))
+            {
+                ret = -EINVAL;
+                goto done;
+            }
+
+            *eq = '\0';
+
+            int ret = create_roothash_file(path, filename);
```
jxyang, line 465: It's not clear to me why the mounted hostfs path would contain a roothash section? Why do we need to create root hash for mounted host ext2 fs? Is it verified during mounting or file accessing?

tests/ext2/crypt/Makefile
mikbras: We still need it when the app is explicitly mounting the ext2 file system (as in this case). In the rootfs case and the --mount case we can pull it from the image itself and put it in the measured archive region. But when the app mounts a file system, the myst command cannot know about it ahead of time (it has to be known when the image is created).

mikbras:
> With the change to automatically extracts the roothash from the file-system image, does the package-sgx command still support --roothash option?

For file systems that are explicit arguments to the ``myst`` command, the roothash is obtained from the ext2 image. For file systems that may be mounted by the app (by calling the ``mount`` function) it is still necessary to provide the ``--roothash`` option. In both cases, the roothash is placed in the archive region and measured during enclave loading.

tools/myst/host/package.c
mikbras: Fixed. I turned it into a "file not found" error that mentions the option name.

tools/myst/host/exec.c
mikbras: We still use ``--roothash=<filename>`` when the app itself mounts the EXT2 fiel system (because the path to the image is not known by the ``myst`` command).

mikbras: We could make ``roothashes`` an array of hashes rather than filenames, but the archive packager expects files and files are still needed for the ``--roothash`` option.

tools/myst/host/package.c
mikbras: This environment variable really does refer to the rootfs path (not a roothash). This is what allows packaging mode to find an ext2 rootfs.

tools/myst/host/exec.c
jxyang: Can you elaborate when would a user want to mount an ext2 file system in their code? And what are the steps required for them to do so?

mikbras: Yes, that will be part of the ext2 doc. In short, a user can pass the path of the ext2 image to the app and then call ``mount(source, target, "ext2")``.

mikbras: I removed the creation of these roothash temporary files by reworking the implementation to use a new "roothashes" region (and .mystroothashes section).

tools/myst/host/utils.c
mikbras: I eliminated saving of roothashes to file in favor of a new "roothashes" binary region, so that we no longer need the temporary file to do CPIO packaging when creating the "archive" region.

tools/myst/host/utils.c
mikbras: The roothash region (part of the measured image) contains the roothashes that are trusted. When an app attempts to mount an ext2 file system, the ext2 image itself contains a self-proclaimed roothash (in the fssig section). The mounter verifies that the self-proclaimed roothash is trusted (i.e., that is is contained in the roothash region). Then during mounting, the hash tree is recomputed to ensure that the roothash is correct.

mikbras:
I separated the root hashes into a new measured region called "roothashes". This region contains zero or more binary SHA-256 hashes. This eliminates the need to create temporary root hash files for the purposes of CPIO packaging (when the root hashes were part of the "archive" CPIO region).

The archive region now just contains public keys. This region will eventually be eliminated and replaced with a "pubkeys" region that will contain zero or more trusted public keys.

mikbras:
I reworked to PR to eliminate the "archive" region, a CPIO archive that contained "/pubkeys" and "/roothashes" directories. The new approach creates two new regions: "roothashes" and "pubkeys". The "roothashes" region contains zero or more binary SHA-256 root hashes. The "pubkeys" region contains public keys in PEM format (each with a binary header). The format is:

```
([header][PEM-public-key])*
```

I also updated the ELF library to allow addition of zero-sized sections (since either of these can be empty).

include/myst/pubkey.h
```diff
@@ -6,9 +6,21 @@
 
 #include <stdint.h>
 
+#define MYST_PUBKEY_NAME_SIZE 256
+#define MYST_PUBKEY_MAGIC 0x6f14783e80394815
+
+typedef struct myst_pubkey_header
+{
+    uint64_t magic; /* should be MYST_PUBKEY_MAGIC */
```
jxyang, line 14: I think we can get away without a public key header. Just treat public keys as `char* []`.

kernel/shell.c
```diff
@@ -231,8 +231,11 @@ static void _mem_command(int argc, char** argv)
     n = __myst_kernel_args.crt_size;
     printf("crt size     =%11zu (%zumb)\n", n, n / mb);
 
-    n = __myst_kernel_args.archive_size;
-    printf("archive size =%11zu (%zumb)\n", n, n / mb);
+    n = __myst_kernel_args.pubkeys_size;
+    printf("pubkeys size =%11zu (%zumb)\n", n, n / mb);
```
jxyang, line 235: I think printing the number of public keys / root hashes, rather than the total size, is more meaningful to the user.

tools/myst/host/package.c
jxyang: I think the consensus is to group Mystikos-specific command line options before the application specific command line options, but keep the option name of the former without `myst-`.

tests/libcxx/Makefile
```diff
@@ -56,13 +54,9 @@ $(APPDIR):
...
 	$(MYST) mkext2 $(APPDIR) $(ROOTFS)
-	# run with posix spawn
-
-$(ROOTHASH): run.c
-	mkdir -p $(SUBOBJDIR)
-	$(MYST) fssig --roothash $(ROOTFS) > $(ROOTHASH)
-
-OPTS += --roothash=$(ROOTHASH)
+ifndef MYST_RELEASE
+	truncate --size=-4096 mkext2
```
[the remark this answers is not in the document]
mikbras: That was a merging error. I removed the truncate.

Yes to the second question.

kernel/shell.c
mikbras: Perhaps; we should reconsider when we decide whether to enhance the shell feature or take it out.

include/myst/pubkey.h
mikbras: Fixed.

include/myst/pubkey.h
```diff
@@ -9,15 +9,6 @@
 #define MYST_PUBKEY_NAME_SIZE 256
 #define MYST_PUBKEY_MAGIC 0x6f14783e80394815
```
jxyang, line 10: The above two definitions are not used anymore, right?

mikbras: Good catch! Fixed and pushed.

jxyang (approved):

PR #497, opened 2021-06-23
improved script for libcxx tests

salsal97:


tests/libcxx/run_tests_locally.sh
```diff
@@ -16,29 +16,18 @@ FILE=$1
 while read test; do
   echo "$test"
   OUTPUT=$(2>&1 timeout 10 make one TEST=$test )
+  RETURN_VAL=$?
   echo $OUTPUT >> temp_$FILE.output
-  HAS_UNHANDLED_SYSCALL=$(2>&1 timeout 10 make one TEST=$test | grep "unhandled")
-  if [ -z "$HAS_UNHANDLED_SYSCALL" ]
+  TERMINATED=$(echo "$OUTPUT" | grep -F 'terminate called') # check for termination due to exceptions
```
vtikoo, line 21: Could you clarify what case this covers? Can the return code ever be zero in this case?

tests/libcxx/run_tests_locally.sh
```diff
@@ -16,29 +16,18 @@ FILE=$1
...
-    if [[ -z $FAILED && -z $FAIL_ENCLAVE && -z $TIMED_OUT && -z $TERMINATED ]]  
-    then
-      echo $test >> temp_passed.output
-    else
-      echo $test >> temp_other_errors.output
-    fi
+    echo $test >> temp_passed.output
   else
-    echo "$test: $HAS_UNHANDLED_SYSCALL" >> temp_unhandled_syscalls.output
+    echo $test >> temp_other_errors.output
```
vtikoo, line 26:
```suggestion
    echo $test >> temp_failed.output
```

tests/libcxx2/run_tests_locally.sh
```diff
@@ -16,32 +16,21 @@ FILE=$1
...
-    if [[ -z $FAILED && -z $FAIL_ENCLAVE && -z $TIMED_OUT && -z $TERMINATED ]]  
-    then
-      echo $test >> temp_passed.output
-    else
-      echo $test >> temp_other_errors.output
-    fi
+    echo $test >> temp_passed.output
   else
-    echo "$test: $HAS_UNHANDLED_SYSCALL" >> temp_unhandled_syscalls.output
+    echo $test >> temp_other_errors.output
```
vtikoo, line 27:
```suggestion
    echo $test >> temp_failed.output
```

tests/libcxx/run_tests_locally.sh
salsal97: This covers the case in which libcxx generates an exception and calls terminate on it. mystikos takes the exception literally, so it seems to have failed, but the error code is still 0. Until we fix the cpp exception handling, these tests would belong in the failed category

tests/libcxx/run_tests_locally.sh
salsal97: I named this other_errors since we put half of them in unhandled syscalls, but I have thought about making this change before - since we dont have any unhandled syscall errors in libcxx. Will add.

tests/libcxx/run_tests_locally.sh
salsal97: eg: 
```
/app/llvm-project/libcxx/test/libcxx/containers/associative/map/at.abort.pass.cpp.exe terminate called after throwing an instance of 'std::out_of_range' what(): map::at
```

vtikoo: Digging further, it seems to me this test is doing the right thing. 
Looks like its built with no-exceptions - https://github.com/salsal97/llvm-project/blob/main/libcxx/test/libcxx/containers/associative/map/at.abort.pass.cpp#L18
According to [this SO answer](https://stackoverflow.com/a/7249442/4307215), with no-exceptions, throws are converted into `abort()` calls. (The test checks for this in the SIGABRT handler - `exit_success`)

salsal97: So should this be considered as a passing test? Its not explicitly built with no-exceptions [here](https://github.com/deislabs/mystikos/blob/main/tests/libcxx/Dockerfile#L21) but it might be the default behavior. Is there a flag that'll enable graceful exception handling in that case?

vtikoo: > So should this be considered as a passing test? 
I think so. One way to confirm would be to compare with running on Alpine Linux.

vtikoo: From https://www.cs.colostate.edu/~cs253/Spring19/Lecture/ExceptionHandling -

If something is thrown but never caught, then the special function terminate() is called, which complains in an implementation-defined manner, and stops the program by calling abort().

Looks like the `no-exceptions` line in the source file is a red-herring.

>  Is there a flag that'll enable graceful exception handling in that case?

In this case, I think the exception is handled the right way. The throwing of the exception(due to L34) causes an SIGABRT, which is handled by the `exit_success` handler, and returns exit code 0.

salsal97: You're right, I understand now. Expecting no output from this test is not the desirable result. I've updated the PR to remove this line from the script and added the exception tests to the pipeline

tests/libcxx/run_tests_locally.sh
```diff
@@ -16,36 +16,24 @@ FILE=$1
 while read test; do
   echo "$test"
   OUTPUT=$(2>&1 timeout 10 make one TEST=$test )
+  RETURN_VAL=$?
   echo $OUTPUT >> temp_$FILE.output
-  HAS_UNHANDLED_SYSCALL=$(2>&1 timeout 10 make one TEST=$test | grep "unhandled")
-  if [ -z "$HAS_UNHANDLED_SYSCALL" ]
+  if [[ "$RETURN_VAL" -eq "0" && -z $TERMINATED ]]
```
vtikoo, line 21: Looks like a use of the `$TERMINATED` var is still lurking around.

tests/libcxx2/run_tests_locally.sh
```diff
@@ -16,37 +16,25 @@ FILE=$1
 
 while read test; do
   echo "$test"
-  OUTPUT=$(2>&1 timeout 3 make one TEST=$test )
+  OUTPUT=$(2>&1 timeout 10 make one TEST=$test )
+  RETURN_VAL=$?
   echo $OUTPUT >> temp_$FILE.output
-  HAS_UNHANDLED_SYSCALL=$(2>&1 timeout 3 make one TEST=$test | grep "unhandled")
-  if [ -z "$HAS_UNHANDLED_SYSCALL" ]
+  if [[ "$RETURN_VAL" -eq "0" && -z $TERMINATED ]]
```
vtikoo, line 22: Same.

tests/libcxx/run_tests_locally.sh
salsal97: My bad! fixed

tests/libcxx2/run_tests_locally.sh
salsal97: Fixed

vtikoo (approved):
LGTM

PR #500, opened 2021-06-23
Forward divide by zero exception code to kernel

vtikoo:

### DIFF
--- a/tools/myst/enc/enc.c
+++ b/tools/myst/enc/enc.c
@@ -136,13 +136,26 @@ static uint64_t _forward_exception_as_signal_to_kernel(
     }
     if (oe_exception_code == OE_EXCEPTION_X87_FLOAT_POINT)
     {
+        // ATTN: Consider implementing accurate si-code for
+        // OE_EXCEPTION_X87_FLOAT_POINT
         siginfo.si_code = SI_KERNEL;
         siginfo.si_signo = SIGFPE;
         (*_kargs.myst_handle_host_signal)(&siginfo, &_mcontext);
         _mcontext_to_oe_context(&_mcontext, oe_context);
         return OE_EXCEPTION_CONTINUE_EXECUTION;
     }
+    if (oe_exception_code == OE_EXCEPTION_DIVIDE_BY_ZERO)
+    {
+        siginfo.si_code = FPE_INTDIV;
+        siginfo.si_signo = SIGFPE;
+        (*_kargs.myst_handle_host_signal)(&siginfo, &_mcontext);
+        _mcontext_to_oe_context(&_mcontext, oe_context);
+        return OE_EXCEPTION_CONTINUE_EXECUTION;
+    }
 
+    // ATTN: Consider forwarding OE_EXCEPTION_BOUND_OUT_OF_RANGE,
+    // OE_EXCEPTION_ACCESS_VIOLATION, OE_EXCEPTION_MISALIGNMENT,
+    // OE_EXCEPTION_SIMD_FLOAT_POINT as signal.
     // Delegate unhandled hardware exceptions to other vector handlers.
     return OE_EXCEPTION_CONTINUE_SEARCH;
 }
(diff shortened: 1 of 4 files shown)


tools/myst/enc/enc.c
```diff
@@ -139,6 +139,15 @@ static uint64_t _forward_exception_as_signal_to_kernel(
         _mcontext_to_oe_context(&_mcontext, oe_context);
         return OE_EXCEPTION_CONTINUE_EXECUTION;
     }
+    if (oe_exception_code == OE_EXCEPTION_DIVIDE_BY_ZERO)
+    {
+        // ATTN: Are other si_code's possible?
+        siginfo.si_code = FPE_INTDIV;
```
jxyang, line 145: The exit info doesn't indicate whether this is a integer divide-by-0 or floating point divide-by-0. I think `FPE_INTDIV` is good enough here. Please update the comment.

jxyang (approved):

jxyang:
For the failed test, we will need Paul's pseudo fork PR to forward the signals on Linux target.

tools/myst/enc/enc.c
vtikoo: Thanks for clarifying. What about other SIGFPE cases like overflow, inexact result, out of range etc., will they generate a different `oe_exception_code`?

vtikoo:
> For the failed test, we will need Paul's pseudo fork PR to forward the signals on Linux target.

Okay, I'll hold off until #414 is merged.

tools/myst/enc/enc.c
bodzhang: OE code maps x86 #DE (Divide Error Exception) to `OE_EXCEPTION_DIVIDE_BY_ZERO`. #DE is integer divide-by-zero exception. Floating point divided-by-zero exception is part of x86 #MF (`OE_EXCEPTION_X87_FLOAT_POINT`) or #XM (`OE_EXCEPTION_SIMD_FLOAT_POINT`). For accurate [si_code](https://www.man7.org/linux/man-pages/man0/signal.h.0p.html) for SIGFPE, the implementation likely needs to parse the FPU status register or MXCSR register (for SIMD). Please take this opportunity to add "ATTN" comments to the code. 

ATTN: Consider implementing accurate si-code for OE_EXCEPTION_X87_FLOAT_POINT

ATTN: Consider forwarding OE_EXCEPTION_BOUND_OUT_OF_RANGE, OE_EXCEPTION_ACCESS_VIOLATION, OE_EXCEPTION_MISALIGNMENT, OE_EXCEPTION_SIMD_FLOAT_POINT as signal.

vtikoo: > ATTN: Consider implementing accurate si-code for OE_EXCEPTION_X87_FLOAT_POINT

Can we replace the si_code to FPE_FLTDIV?

vtikoo: > Please take this opportunity to add "ATTN" comments to the code.

Added.

jxyang: @bodzhang Yes we should fill in the accurate `si_code` as long as OE provides the details.

mingweishih: @vtikoo 
`FPE_FLTINV`: Invalid floating-point operation
may be a more general option to put for now.
As @bodzhang mentions, we need some work to determine the accurate code, which is captured by the ATTN comment.

bodzhang: I think for OE_EXCEPTION_DIVIDE_BY_ZERO, FPE_INTDIV is the right si_code.

mingweishih: Yes, I was referring to the OE_EXCEPTION_X87_FLOAT_POINT case

tools/myst/enc/enc.c
```diff
@@ -133,13 +133,24 @@ static uint64_t _forward_exception_as_signal_to_kernel(
     }
     if (oe_exception_code == OE_EXCEPTION_X87_FLOAT_POINT)
     {
-        siginfo.si_code = SI_KERNEL;
+        siginfo.si_code = FPE_FLTINV;
```
mingweishih, line 136: I would suggest still keep the following comment so that we can improve this later.

ATTN: Consider implementing accurate si-code for OE_EXCEPTION_X87_FLOAT_POINT

vtikoo: Makes sense, added the comment back.

mingweishih (approved):

PR #504, opened 2021-06-24
enhanced debug documentation

salsal97:
Final view here - https://github.com/salsal97/mystikos/blob/docs/doc/dev-jumpstart.md#debugging


doc/dev-jumpstart.md
```diff
@@ -123,12 +123,29 @@ customized Dockerfile
...
+
+O Call -> Calls into the host (untrusted environment)
+
 1. For most applications under `tests`, we can launch the debugger with
 command `make tests GDB=1`. For example:
     ```
     cd Mystikos/tests/hello
     make && make tests GDB=1
     ```
+    For solutions that are run in [package mode](../solutions/dotnet/Makefile#23), ensure that the field `"Debug":1` is set in the `config.json` file, and the debugger can be launched using the run command:
```
jxyang, line 144: `solutions` -> `applications`. We could have tests in package mode later.

jxyang (approved):

doc/dev-jumpstart.md
salsal97: Sounds right, made the change

mikbras (approved):
Thank you for doing this Saloni.

PR #509, opened 2021-06-25
Add openssl to third-party

mikbras:
Remove dependency on system openssl by adding openssl to third-party.

### DIFF
--- a/third_party/openssl/Makefile
+++ b/third_party/openssl/Makefile
@@ -18,8 +18,11 @@ all: .install
 	touch .build
 
 .configure: .openssl
-	( cd openssl; ./config --prefix=$(INSTALL_PREFIX) )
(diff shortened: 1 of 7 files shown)


Francis-Liu (approved):

third_party/openssl/Makefile
```diff
@@ -0,0 +1,35 @@
+TOP=$(abspath $(CURDIR)/../..)
+include $(TOP)/defs.mak
+
+URL=https://github.com/openssl/openssl
+
+BRANCH=OpenSSL_1_1_1-stable
```
mingweishih, line 6: Another option is depending on a specific release tag (e.g., https://github.com/openssl/openssl/tree/OpenSSL_1_1_1k).

mikbras: Good idea! Done. :)

mingweishih (approved):

PR #510, opened 2021-06-26
Add symlinkat syscall support

Francis-Liu:

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -797,56 +797,107 @@ done:
     return ret;
 }
 
+// Caller should check *abspath_out before free:
+//
+//   if (*abspath_out != pathname)
+//        free(*abspath_out)
 long myst_get_absolute_path_from_dirfd(
     int dirfd,
-    const char* filename,
-    char* abspath_out,
-    size_t size)
+    const char* pathname,
+    int flags,
+    char** abspath_out)
 {
     long ret = 0;
+    char* path_out = NULL;
     struct locals
     {
         char dirname[PATH_MAX];
     }* locals = NULL;
 
-    myst_fdtable_t* fdtable = myst_fdtable_current();
-    myst_fdtable_type_t type;
-    void* device = NULL;
-    void* object = NULL;
-    myst_fs_t* fs;
-    myst_file_t* file;
+    if (!pathname || !abspath_out)
+        ERAISE(-EINVAL);
 
-    if (dirfd < 0)
-        ERAISE(-EBADF);
-
-    if (!(locals = malloc(sizeof(struct locals))))
-        ERAISE(-ENOMEM);
-
-    /* first check dirfd is of file type, e.g. not tty */
-    ECHECK(myst_fdtable_get_any(fdtable, dirfd, &type, &device, &object));
-    if (type != MYST_FDTABLE_TYPE_FILE)
-        ERAISE(-ENOTDIR);
-
-    /* get the file object for the dirfd */
-    ECHECK(myst_fdtable_get_file(fdtable, dirfd, &fs, &file));
-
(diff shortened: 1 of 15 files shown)


kernel/syscall.c
```diff
@@ -1785,6 +1793,41 @@ long myst_syscall_symlink(const char* target, const char* linkpath)
...
+    struct locals
+    {
+        char abspath[PATH_MAX];
+    };
+    struct locals* locals = NULL;
+
+    /* If linkpath is absolute, then ignore dirfd */
+    if (*linkpath == '/' || newdirfd == AT_FDCWD)
+    {
+        ret = myst_syscall_symlink(target, linkpath);
```
jxyang, line 1811: This pattern is repeated enough times, we check should `dirfd` be ignored or not and call `myst_get_absolute_path_from_dirfd` is not. I wonder if we can consolidate and simply the logic as:

```c
myst_get_absolute_path_from_dirfd(dirfd, char* path_in, char** path_out)
{
    if (dirfd == AT_FDCWD || *path_in == '/')
    {
        *path_out = path_in;
    }
   else 
  {
      *path_out = malloc(....)
       ... resolve and copy the path_out ...
  }
}
```
And then in the caller's clean up code, it determines if there is heap variable needs to free with:
```c
  if (*path_out != path_in)
      free(*path_out);
```

vtikoo: [resolve_at_path()](https://github.com/deislabs/mystikos/blob/main/kernel/chown.c#L204) in kernel/chown.c does something similar. Maybe that can be reused here.

Francis-Liu: @jxyang Good idea! But there seems to be a small issue that `pathname` (`path_in` in your sample code) is of type `const char*`. Is type conversion OK per our standard?

Francis-Liu: > resolve_at_path

@vtikoo Thanks for pointing it out. Do you mean to reuse `resolve_at_path` in `myst_get_absolute_path_from_dirfd`? Or to reuse `myst_get_absolute_path_from_dirfd` in `resolve_at_path`? Or to merge them into one function?

vtikoo: `resolve_at_path` is at a higher level than `myst_get_absolute_path_from_dirfd`. You can check `myst_syscall_fchownat` in kernel/chown.c as a reference usage of `resolve_at_path`.

Francis-Liu: Oh, I got what you mean. Would you mind I move `resolve_at_path` into syscall.c?

vtikoo: Sure, go for it 👍

jxyang: Note `resolve_at_path` expects the caller to allocate memory, while my proposed `myst_get_absolute_path_from_dirfd` expects the callee to allocate memory (because it has better idea about how long the resolved path should be).

@Francis-Liu  You can cast away the `const` qualifier.

kernel/syscall.c
```diff
@@ -797,49 +797,108 @@ long myst_syscall_open(const char* pathname, int flags, mode_t mode)
     return ret;
 }
 
+// Caller should check *abspath_out before free:
+//
+//   if (*abspath_out != pathname)
+//        free(*abspath_out)
```
vtikoo, line 803: Some comments here would help I think. Especially about the null string return.

vtikoo (approved):
LGTM. Single comment about adding more function documentation for `myst_get_absolute_path_from_dirfd`.

kernel/syscall.c
Francis-Liu: I will add some comments in another PR. Merged this PR first to unblock #542

PR #516, opened 2021-06-29
HIGH PRIORITY: Fix EXT2 directory corruption bug

mikbras:
This PR fixes a bug in the EXT2 file system in which ``_find_dirent`` failed because it was relying on a stale inode. This occurred when an entry was removed from the directory (``_remove_dirent``), making it one block smaller and the subsequent call to ``_find_dirent`` attempted to read the directory using the old size (from the stale inode). The fix involved modifying the signature of ``_remove_dirent`` to accept a writable inode parameter (which updates the formerly stale inode).


vtikoo (approved):
LGTM

PR #519, opened 2021-06-29
Paulcallen/mount buffers

paulcallen:
packages up the command line arguments for mount mapping into a single buffer to marshal into the enclave.

### DIFF
--- a/include/myst/options.h
+++ b/include/myst/options.h
@@ -9,12 +9,6 @@
 #include <myst/kernel.h>
 #include <myst/types.h>
 
-typedef struct _myst_mount_mapping
-{
-    char** mounts; /* array of source=target */
-    size_t mounts_count;
-} myst_mount_mapping_t;
-
 typedef struct myst_options
 {
     bool trace_errors;
@@ -28,7 +22,6 @@ typedef struct myst_options
     char rootfs[PATH_MAX];
 
     myst_host_enc_uid_gid_mappings host_enc_uid_gid_mappings;
-    myst_mount_mapping_t mount_mapping;
 } myst_options_t;
 
 extern myst_options_t __options;
--- a/tools/myst/common.c
+++ b/tools/myst/common.c
@@ -1,11 +1,12 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-#include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include "shared.h"
+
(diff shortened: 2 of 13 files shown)


include/myst/options.h
```diff
@@ -28,7 +22,6 @@ typedef struct myst_options
     char rootfs[PATH_MAX];
 
     myst_host_enc_uid_gid_mappings host_enc_uid_gid_mappings;
-    myst_mount_mapping_t mount_mapping;
```
mikbras, line 31: I was confused because it was not called ``mount_mappings`` (plural).

tools/myst/common.c
```diff
@@ -40,16 +42,17 @@ int myst_expand_size_string_to_ulong(const char* size_string, size_t* size)
...
 
-    for (size_t i = 0; i < mount_mapping->mounts_count; i++)
+    for (size_t i = 0; i < mount_mapping->size; i++)
     {
         char* source;
         char* target;
         bool found = false;
-        source = strtok(mount_mapping->mounts[i], "=");
+
+        source = strtok((char*)(mount_mapping->data[i]), "=");
```
mikbras, line 55: It probably doesn't matter because this is a single threaded program, but I would probably use the thread-safe form: ``strtok_r``.

mikbras (approved):
Looks great! Thanks for doing this.

Minor: in a few places ``myst_mapping`` refers to multiple mappings I think.

Minor: maybe consider using ``strtok_r`` (string-safe form of ``strtok``).

include/myst/options.h
```diff
@@ -9,12 +9,6 @@
 #include <myst/kernel.h>
 #include <myst/types.h>
 
-typedef struct _myst_mount_mapping
-{
-    char** mounts; /* array of source=target */
-    size_t mounts_count;
-} myst_mount_mapping_t;
-
 typedef struct myst_options
```
mikbras, line 12: This structure is passed to ``myst_enter_ecall()`` but it contains ``mount_mapping``, which contains strings, which is a security concern if the pointers were actually used.

tools/myst/common.c
paulcallen: fixed

include/myst/options.h
paulcallen: this is deleted code!

include/myst/options.h
paulcallen: This is exactly what this PR is addressing :)

PR #526, opened 2021-07-01
Fix SYS_prlimit64 for RLIMIT_STACK

rads-1996:
Return static value of 1MB for stack limit.

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -2908,9 +2908,8 @@ long myst_syscall_prlimit64(
     {
         if (old_rlim)
         {
-            // TODO: get actual stack limits
-            old_rlim->rlim_cur = 1024 * 1024 * 1024;
-            old_rlim->rlim_max = 1024 * 1024 * 1024;
+            old_rlim->rlim_cur = MYST_PROCESS_INIT_STACK_SIZE;
+            old_rlim->rlim_max = MYST_PROCESS_MAX_STACK_SIZE;
         }
     }
     else
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -2890,20 +2890,32 @@ long myst_syscall_prlimit64(
     if (pid)
         return -EINVAL;
 
-    // Only support resource NOFILE
-    if (resource != RLIMIT_NOFILE)
-        return -EINVAL;
-
-    if (old_rlim)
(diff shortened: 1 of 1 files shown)


kernel/syscall.c
```diff
@@ -2815,20 +2815,32 @@ long myst_syscall_prlimit64(
...
+        }
+    }
+    else if (resource == RLIMIT_STACK)
     {
-        // ATTN: return currently effective RLIMIT_NOFILE settings
-        old_rlim->rlim_cur = 65536;
-        old_rlim->rlim_max = 65536;
+        if (old_rlim)
+        {
+            // TODO: get actual stack limits
```
bodzhang, line 2836: Mystikos process' main thread is initialized with a default stack size in [kernel/exec.c](https://github.com/deislabs/mystikos/blob/87c7a2ec370115ac35452fa1dd3fffe0c544d044/kernel/exec.c#L838). Probably should define a DEFAULT_PROCESS_STACK_SIZE and use here and in kernel/exec.c. 

@mikbras , any suggestion on which header file to include the DEFAULT_PROCESS_STACK_SIZE definition?

bodzhang (changes requested):

kernel/syscall.c
bodzhang:
```suggestion
```

kernel/syscall.c
```diff
@@ -2815,20 +2815,32 @@ long myst_syscall_prlimit64(
...
+            old_rlim->rlim_cur = 1024 * 1024 * 1024;
```
bodzhang, line 2837:
```suggestion
            old_rlim->rlim_cur = MYST_PROCESS_INIT_STACK_SIZE;
```

kernel/syscall.c
```diff
@@ -2815,20 +2815,32 @@ long myst_syscall_prlimit64(
...
+            old_rlim->rlim_max = 1024 * 1024 * 1024;
```
bodzhang, line 2838:
```suggestion
            old_rlim->rlim_max = MYST_PROCESS_INIT_STACK_SIZE;
```
MYST_PROCESS_INIT_STACK_SIZE is defined in the main branch, please rebase.

bodzhang (changes requested):

bodzhang (approved):

PR #528, opened 2021-07-01
Address release build failure of mman.c

bodzhang:
Compiler warning on mman.c were treated as error in MYST_RELEASE=1 build. Updated the code to avoid the warning.


paulcallen (approved):

salsal97 (approved):
lgtm, passes release build stage

PR #533, opened 2021-07-01
Workaround for pipeline errors cause by different OE versions

mikbras:
Please try this one line fix to workaround errors in the pipeline caused by different versions of OE.


vtikoo (approved):
LGTM

PR #537, opened 2021-07-06
Pipefix july

jxyang:
This PR fixes hangs in the event of shutdown for two cases:

* dotnet-proc-maps
* libcxx when a child thread fails an assertion

In both cases, there were thread waiting for a futex with infinite timeout, thus caused the process not able to shutdown.  This PR attempts to wake up the futex while sending the kill signal to the thread.

### DIFF
--- a/kernel/thread.c
+++ b/kernel/thread.c
@@ -1210,14 +1210,19 @@ size_t myst_kill_thread_group()
     myst_spin_lock(process->thread_lock);
     for (t = tail; t != NULL; t = t->group_prev)
     {
-        //        if (!myst_is_process_thread(t) && t->status == MYST_RUNNING)
         if (t != thread && t->status == MYST_RUNNING)
         {
             count++;
             myst_spin_unlock(process->thread_lock);
             myst_signal_deliver(t, SIGKILL, 0);
+
             // Wake up the thread from futex_wait if necessary.
-            myst_cond_signal_thread(t->signal.cond_wait, t);
+            if (t->signal.waiting_on_event)
+            {
+                myst_tcall_wake(t->event);
+                t->signal.waiting_on_event = false;
+            }
+
             myst_spin_lock(process->thread_lock);
         }
     }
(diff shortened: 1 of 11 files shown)


kernel/thread.c
```diff
@@ -1242,7 +1247,14 @@ size_t myst_kill_thread_group()
         for (t = tail; t != NULL; t = t->group_prev)
         {
             if (t != process && t != thread && t->status != MYST_ZOMBIE)
+            {
```
paulcallen, line 1250: may wont to delete the printf!
In my PR I did another WAKE here as there is a race condition between the signal on the killing and the waiting. The code may be in a transitional state where it has not quite made it to the wait and so the previous wake may not be enough. Adding it here also gets that other race.

kernel/thread.c
```diff
@@ -1210,14 +1210,19 @@ size_t myst_kill_thread_group()
...
-        //        if (!myst_is_process_thread(t) && t->status == MYST_RUNNING)
         if (t != thread && t->status == MYST_RUNNING)
         {
             count++;
             myst_spin_unlock(process->thread_lock);
             myst_signal_deliver(t, SIGKILL, 0);
+
             // Wake up the thread from futex_wait if necessary.
-            myst_cond_signal_thread(t->signal.cond_wait, t);
+            if (t->signal.waiting_on_event)
```
vtikoo, line 1220: Can we move this to `myst_signal_deliver`? That would cover waking futexes in non-shutdown scenarios, like musl SIGSYNCCALL.
However, in the non-shutdown case there is also the issue of restarting syscalls if SA_RESTART is set.

kernel/thread.c
jxyang: Will add this to the following loop. Thanks.

kernel/thread.c
jxyang: The last time I tried that, dotnet-proc-maps hangs for some reasons. I will give it another try.

I am not `SA_RESTART` applies when a kill signal interrupts the operation.

paulcallen (approved):

PR #541, opened 2021-07-06
DO NOT MERGE: CPIO DEFLATE EXPERIMENT

mikbras:


mikbras:
This is faster but slower than ext2.

PR #545, opened 2021-07-07
Support mmap() request with preferred address

bodzhang:
For mmap(addr, ...) request with non-zero addr, if the requested mapping is not part of an existing mapping, and MAP_FIXED flag is not set, treat the request as if addr=0.

### DIFF
--- a/kernel/mman.c
+++ b/kernel/mman.c
@@ -755,13 +755,6 @@ static int _mmap(
             ret = -EINVAL;
             goto done;
         }
-
-        if (flags & MYST_MAP_FIXED)
-        {
-            _mman_set_err(mman, "bad flags parameter: remove MYST_MAP_FIXED");
-            ret = -EINVAL;
-            goto done;
-        }
     }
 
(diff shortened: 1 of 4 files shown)


kernel/mman.c
```diff
@@ -789,21 +782,27 @@ static int _mmap(
...
+            *ptr_out = addr;
+            goto done;
+        }
+        else if (flags & MYST_MAP_FIXED)
         {
             _mman_set_err(
                 mman,
                 "bad addr parameter: "
-                "must be null or part of an existing mapping");
+                "MAP_FIXED request, but or part of an existing mapping");
```
mikbras, line 798: Typo?

mikbras (commented):
Looks good but small typo.

kernel/mman.c
bodzhang: @mikbras, thanks for spotting this. Fixed.

mikbras (approved):

Issue #371, opened 2021-05-18
Avoid calling fini functions before atexit functions


salsal97:
The code coverage exposed that we are not calling __free_futex__ function in kernel/futex.c at all, but it is being initialized as a atexit function in the same file. This leads us to believe that we are calling our global destructor functions (fini functions) before we are calling the atexit functions, since gcov is run with global destructor functions.

The ideal flow would be - contructors -> functions -> atexit functions -> destructor

We are currently interchanging the position of the atexit function and the destructor. Code coverage is linked here, look under `kernel/futex.c`  - https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=7330&view=codecoverage-tab

salsal97:
https://github.com/deislabs/mystikos/pull/546 fixed it

PR #546, opened 2021-07-07
Invoke atexit functions before fini functions

mikbras:
This PR fixes the wrong ordering of ``atexit`` and ``fini`` functions, where the ``fini`` functions should be called last.


salsal97:
This PR fixes https://github.com/deislabs/mystikos/issues/371

salsal97 (approved):
lgtm. Fixes issue that incorrectly reported that the _free_futex_ function was never called, ref: https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=8884&view=codecoverage-tab

PR #548, opened 2021-07-07
Separate pipeline for dotnet 5 test suite

vtikoo:

### DIFF
--- a/.azure_pipelines/ci-pipeline-dotnet-test-suite.yml
+++ b/.azure_pipelines/ci-pipeline-dotnet-test-suite.yml
@@ -1,0 +1,66 @@
+variables:
+  - group: myst-pipeline
+  - name: SCRIPTS_ROOT
+    value: $(Build.SourcesDirectory)/.azure_pipelines/scripts
+
+# dotnet tests pipeline will be triggered in pipeline schedule settings.
+trigger: none
+pr: none
+
+jobs:
+  - job: BuildAndTest
+    displayName: 'Building and Testing'
+    # 10m build
+    # ~3h for dotnet 5 test suite
+    timeoutInMinutes: 300
+    pool: '1804DC4CCagentpool'
+
+    steps:
+      # perform cleanup before starting pipeline
+      - script: |
+          sudo rm -rf /tmp/myst*
+          df
+        displayName: 'cleanup tmpfiles'
+      # install minimum required package to run pipeline
+      - script: |
+          curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+          sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+          sudo apt-get update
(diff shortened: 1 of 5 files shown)


.azure_pipelines/ci-pipeline-dotnet-test-suite.yml
```diff
@@ -0,0 +1,66 @@
...
+  - group: myst-pipeline
+  - name: SCRIPTS_ROOT
+    value: $(Build.SourcesDirectory)/.azure_pipelines/scripts
+
+# dotnet tests pipeline will be triggered in pipeline schedule settings.
+trigger: none
+pr: none
+
+jobs:
+  - job: BuildAndTest
```
salsal97, line 11: Theres an extra space here before the '-' that az pipelines might complain about

salsal97: The entire thing would accordingly go forward by 1 space.. (this is the syntax mistake I always make too)

vtikoo: The indentation here is similar to ci-pipelines-makefile.yml, which indents the jobs list by 2 spaces.

salsal97: Gotcha, I was looking at [this](https://github.com/deislabs/mystikos/blob/main/.azure_pipelines/ci-pipeline-makefile-nightly.yml). Should work then.

solutions/coreclr/Makefile
```diff
@@ -33,7 +33,7 @@ ext2fs: appdir
 	$(MYST) fssig --roothash ext2fs > roothash
 
 run:
-	./test-runner.sh $(MYST_EXEC) ext2 $(TEST_FILE)
+	COMPlus_EnableDiagnostics=0 COMPlus_EnableAlternateStackCheck=1 ./test-runner.sh $(MYST_EXEC) ext2 $(TEST_FILE)
```
jxyang, line 36: I thought we block the environment variables from host by default, unless we explicitly allow it in config.json. Did we change the behavior recently?

vtikoo: I tested with a simple program. It does look like passing host env variables is the default.

jxyang: I think that's a regression then. In the [document](https://github.com/deislabs/mystikos/blob/main/doc/sign-package.md) it says:

>  HostEnvironmentVariables | A list of environment variables that can be imported from the insecure host

Without an explicit entry in config.json that's measured, we shouldn't rely on any environment variable from the untrusted host, even though that's convenient. 

The remedy here is to create a config.json and put these two variables in there and run myst with `--app-config-path`

jxyang: Please create an issue for the regression.

vtikoo: > The remedy here is to create a config.json and put these two variables in there and run myst with --app-config-path

Fixed.

jxyang (approved):

PR #549, opened 2021-07-07
Optionally support memcheck on process mappings

mikbras:
Process mappings are excluded from the debug-malloc leak checker (memcheck) by default. Provide a way to enable leaks checks on these memory objects.


vtikoo (approved):
LGTM.

PR #551, opened 2021-07-07
Add some comments for myst_get_absolute_path_from_dirfd

Francis-Liu:

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -835,10 +835,28 @@ done:
     return ret;
 }
 
-// Caller should check *abspath_out before free:
+// Given a dir fd and a pathname, return a concatenated absolute path.
+// This function allocates buffer for the concatenated abspath. The
+// caller doesn't need to preallocate buffer, but should free the
+// abspath buffer after using it.
 //
-//   if (*abspath_out != pathname)
-//        free(*abspath_out)
+// Caveats:
+//
+// - Sometimes abspath is set to equal to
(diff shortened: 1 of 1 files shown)


kernel/syscall.c
```diff
@@ -835,10 +835,19 @@ long myst_syscall_open(const char* pathname, int flags, mode_t mode)
     return ret;
 }
 
-// Caller should check *abspath_out before free:
+// Given a dir fd and a pathname, return an concatenated absolute path.
```
vtikoo, line 838:
```suggestion
// Given a dir fd and a pathname, return a concatenated absolute path.
```

kernel/syscall.c
```diff
@@ -835,10 +835,19 @@ long myst_syscall_open(const char* pathname, int flags, mode_t mode)
...
+// *abspath_out before free:
+// e.g.
+//     if (*abspath_out != pathname)
+//         free(*abspath_out)
 //
-//   if (*abspath_out != pathname)
-//        free(*abspath_out)
+// If this function hits an error and doesn't return SUCCESS, it will
+// free the abspath buffer by itself. Also, *abspath_out == NULL.
+// Thus, the check above is still valid.
```
vtikoo, line 850: Could you also add a line about the case where *abspath_out is '\0'. Caller should differentiate this case from other cases and apply the fd version of the syscall using dirfd.

vtikoo (approved):
Looks great. Thanks for improving the documentation!

PR #558, opened 2021-07-09
Implement pseudo fork-exec

paulcallen:
If in fork-exec mode the caller thread of fork will wait for the child
forked process to exit or call exec

### DIFF
--- a/crt/fork.c
+++ b/crt/fork.c
@@ -385,8 +385,8 @@ myst_fork(void)
             return tmp_ret;
         }
 
-        /* wait for child to set args->pid */
         {
+            /* wait for child to set args->pid */
             struct timespec req;
             req.tv_sec = 0;
             req.tv_nsec = 1000;
@@ -394,6 +394,12 @@ myst_fork(void)
                 nanosleep(&req, NULL);
 
             pid = args->pid;
+
+            /* Wait if fork mode requires it */
+            if (arg.fork_mode == myst_fork_pseudo_wait_for_exit_exec)
+            {
+                syscall(SYS_fork_wait_exec_exit);
+            }
         }
     }
     else /* child */
--- a/kernel/exec.c
+++ b/kernel/exec.c
@@ -968,11 +968,6 @@ int myst_exec(
     thread->main.exec_crt_data = crt_data;
     thread->main.exec_crt_size = crt_size;
 
-    /* remove the thread clone flag CLONE_VFORK flag  as this is used to
-     * determine if we do CRT cleanup. As we are re-initializing the CRT it is
-     * safe to re-enable the cleanup. */
-    thread->clone.flags &= ~CLONE_VFORK;
-
     /* close file descriptors with FD_CLOEXEC flag */
     {
         myst_fdtable_t* fdtable = myst_fdtable_current();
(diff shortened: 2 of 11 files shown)


kernel/exec.c
```diff
@@ -994,6 +989,26 @@ int myst_exec(
...
+     * At the same time we may need to wake the parent processes thread that
+     * launched us in the case we were created in the fork/exec wait mode. */
+    if (thread->clone.flags & CLONE_VFORK)
+    {
+        myst_thread_t* parent_launch_thread = myst_find_process_specific_thread(
+            thread->clone.vfork_parent_pid, thread->clone.vfork_parent_tid);
+        if (parent_launch_thread)
+        {
+            __sync_val_compare_and_swap(
+                &parent_launch_thread->fork_exec_futex_wait, 0, 1);
```
jxyang, line 1004: Do we need to update the futex value to wake up the waiter?

jxyang: Why don't we put the futex variable in the child?

kernel/thread.c
```diff
@@ -605,12 +606,49 @@ myst_thread_t* myst_find_thread(int tid)
...
+    }
+    myst_spin_unlock(&myst_process_list_lock);
+
+    if (find_thread == NULL)
+        goto done;
+
+    myst_spin_lock(find_thread->thread_lock);
+    while (find_thread && find_thread->tid != tid)
+    {
+        find_thread = find_thread->group_next;
```
jxyang, line 643: Are we sure the process thread is always the first in the thread group list?

crt/fork.c
```diff
@@ -385,15 +385,21 @@ myst_fork(void)
...
             req.tv_nsec = 1000;
             while (args->pid == 0)
                 nanosleep(&req, NULL);
 
             pid = args->pid;
+
+            /* Wait if fork mode requires it */
+            if (arg.fork_mode == myst_fork_pseudo_wait_for_exit_exec)
+            {
+                syscall(SYS_fork_wait_exec_exit, args->pid);
```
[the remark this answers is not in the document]
paulcallen: `The problem is in where the futex is kept and managed. The futex absolutely needs to be managed in the kernel

kernel/thread.c
paulcallen: yes

paulcallen: there are a number of assumptions made elsewhere of the fact

kernel/exec.c
paulcallen: We set the futex value to 1 and trigger

crt/fork.c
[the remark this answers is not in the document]
paulcallen: we are in the CRT here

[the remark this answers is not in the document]
jxyang: We could use the libc `futex` wait call here, and use `myst_futex_wake` in the kernel.

kernel/exec.c
jxyang: I think setting the value to 1 is to avoid future futex waiters on the same address to get stuck. It doesn't make a difference in our simple scenario here.

jxyang: If we put the futex variable in the child, we don't have to search for the parent.

crt/fork.c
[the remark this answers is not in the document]
paulcallen: we dont have access to the right address for the futex, right? or is this a usage of futex I am not familiar with?

kernel/exec.c
paulcallen: setting it to one is to make sure it has not already been triggered before we hit the wait

crt/fork.c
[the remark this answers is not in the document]
jxyang: We should avoid introducing more custom syscalls unless absolutely necessary, as it introduces incompatibility with the Linux kernel and hampers the future upstreaming of MUSL. Here we can make the parent and the child operate on an address that both have access to.

kernel/exec.c
jxyang: There is just 1 waiter and it only waits once. And when it waits, it sets the variable to 0 first.

crt/fork.c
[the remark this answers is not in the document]
paulcallen: Our fork implementation will never be upstreamed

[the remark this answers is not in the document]
jxyang: You never know...

jxyang (approved):

PR #561, opened 2021-07-09
Report correct exit code on default signal handling

jxyang:

### DIFF
--- a/kernel/signal.c
+++ b/kernel/signal.c
@@ -164,7 +164,7 @@ static long _default_signal_handler(unsigned signum)
         myst_fork_exec_futex_wake(process_thread);
     }
 
-    thread->exit_status = 0;
+    thread->exit_status = 128 + signum;
     thread->status = MYST_KILLED;
     thread->terminating_signum = signum;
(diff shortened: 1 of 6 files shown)


salsal97 (approved):
lgtm

kernel/signal.c
```diff
@@ -155,7 +155,7 @@ static long _default_signal_handler(unsigned signum)
             myst_tcall_wake(process_thread->event);
         }
     }
-    thread->exit_status = 0;
+    thread->exit_status = 128 + signum;
```
mikbras, line 158: What is the significance of 128?

mikbras (approved):

mikbras:
I tried this and it resolves https://github.com/deislabs/mystikos/issues/567.

kernel/signal.c
mikbras: Can you use _NSIG here?

jxyang: The magic number is from here: https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html. There seems to be no macro for it.

PR #569, opened 2021-07-11
Optimize kstack usage

mikbras:
Currently Mystikos adds a **kernel stacks region** to the image during loading. This region contains **1024**, **64-kilobyte** stacks. The cost is incurred by any application, regardless of how many stacks are actually needed (one stack is required per thread). Loading this region during startup roughly doubles the boot time for the typical application.

This PR removes the **kernel stacks region** and instead allocates each kernel stack on demand using ``mmap``. Kernel stacks are cached and re-used and obtaining a kernel stack to service a syscall is an O(1) operation. This approach was made possible thanks to the ``mprotect`` feature, which is needed to set permissions for a guard page (at the top of the stack).

This change reduces the image size, decreases the startup time, and reduces the code size.

### DIFF
--- a/kernel/kstack.c
+++ b/kernel/kstack.c
@@ -1,34 +1,43 @@
+#include <sys/mman.h>
+
 #include <myst/kernel.h>
 #include <myst/kstack.h>
+#include <myst/mmanutils.h>
 #include <myst/spinlock.h>
+#include <myst/thread.h>
 #include <myst/time.h>
 
-static int _initialized;
 static myst_kstack_t* _head;
 static myst_spinlock_t _lock;
 
-void myst_init_kstacks(void)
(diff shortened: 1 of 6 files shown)


kernel/kstack.c
```diff
@@ -1,43 +1,62 @@
...
+        {
+            /* use the first kstack on the list (likely case) */
+            if ((kstack = _head))
+                _head = _head->u.next;
+        }
+        else
+        {
+            /* allocate a new kstack (unlikely case) */
+            uint8_t* sp = _stack + sizeof(_stack);
+            kstack = (myst_kstack_t*)myst_call_on_stack(sp, _new_kstack, NULL);
```
jxyang, line 58: Is there a possible race condition on `_stack`?

jxyang (approved):

PR #577, opened 2021-07-13
increase release pipeline timeout

salsal97:


jxyang:
Why is release pipeline taking longer. I am scratching my head because the release build with compile with more aggressive compiler optimizations.

salsal97:
If this pipeline succeeds, this PR may not be needed 

https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=9078&view=logs&j=fc362045-4e14-5eb3-5f80-de5683113e49&t=0d6dce9c-5b37-5616-e3a8-7458a3ff5fcc

PR #579, opened 2021-07-13
gcov updates

salsal97:
added missing gcov coverage
remove gcov exception tests
overridden settings for lcov
removed optimizations for gcov
added branch coverage option to genhtml


salsal97:
https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=9081&view=results

defs.mak
```diff
@@ -61,6 +61,10 @@ ifeq ($(MYST_RELEASE),1)
 OPTIMIZATION_CFLAGS += -O3
 endif
 
+ifdef MYST_ENABLE_GCOV
+GCOV_CFLAGS += -O0 --coverage -fprofile-arcs -ftest-coverage
```
mingweishih, line 65: Isn't `--coverage` the same as `-fprofile-arcs -ftest-coverage`?

salsal97: I believe so, thanks for pointing it out. I was trying something out, now removed it.

defs.mak
```diff
@@ -61,6 +61,10 @@ ifeq ($(MYST_RELEASE),1)
 OPTIMIZATION_CFLAGS += -O3
 endif
 
+ifdef MYST_ENABLE_GCOV
```
jxyang, line 64: Why do we need to enclose `GCOV_CFLAGS` in `ifdef MYST_ENABLE_GCOV`?

scripts/myst_cc
```diff
@@ -12,16 +12,30 @@
...
+cp -r "${MYST_ROOT}"/build/obj/utils/* -t "${MYST_ROOT}"/utils
+cp -r "${MYST_ROOT}"/build/obj/hostfs/* -t "${MYST_ROOT}"/hostfs
+cp -r "${MYST_ROOT}"/build/obj/host/* -t "${MYST_ROOT}"/host
+cp -r "${MYST_ROOT}"/build/obj/ext2/* -t "${MYST_ROOT}"/ext2
 cp -r "${MYST_ROOT}"/build/obj/kernel/* -t "${MYST_ROOT}"/kernel
-cp -r "${MYST_ROOT}"/build/obj/crt/"${MYST_ROOT}"/third_party/* -t "${MYST_ROOT}"/third_party
+cp -r "${MYST_ROOT}"/build/obj/oe/* -t "${MYST_ROOT}"/oe
+cp -r "${MYST_ROOT}"/build/obj/json/* -t "${MYST_ROOT}"/json
+cp -r "${MYST_ROOT}"/build/obj/target/* -t "${MYST_ROOT}"/target
+# cp -r "${MYST_ROOT}"/build/obj/crt/"${MYST_ROOT}"/third_party/* -t "${MYST_ROOT}"/third_party
```
jxyang, line 23: Do we want code coverage for MUSL?

tests/elf/Makefile
```diff
@@ -9,9 +9,9 @@ SOURCES = $(wildcard *.c)
 
 INCLUDES = -I$(SUBOBJDIR) -I$(INCDIR)
 
-CFLAGS = $(OEHOST_CFLAGS)
+CFLAGS = $(OEHOST_CFLAGS) $(GCOV_CFLAGS)
 
-LDFLAGS = $(OEHOST_LDFLAGS)
+LDFLAGS = $(OEHOST_LDFLAGS) $(GCOV_LDFLAGS)
```
jxyang, line 14: I don't think `GCOV_LDFLAGS` is ever defined.

tests/elf/Makefile
```diff
@@ -9,9 +9,9 @@ SOURCES = $(wildcard *.c)
 
 INCLUDES = -I$(SUBOBJDIR) -I$(INCDIR)
 
-CFLAGS = $(OEHOST_CFLAGS)
+CFLAGS = $(OEHOST_CFLAGS) $(GCOV_CFLAGS)
```
jxyang, line 12: Should protect ` $(GCOV_CFLAGS)` with `ifdef MYST_ENABLE_GCOV`

tests/run/Makefile
```diff
@@ -10,9 +10,9 @@ SOURCES = $(SUBOBJDIR)/run_u.c host.c
 
 INCLUDES = $(OEHOST_INCLUDES) -I$(SUBOBJDIR)
 
-CFLAGS = $(OEHOST_CFLAGS)
+CFLAGS = $(OEHOST_CFLAGS) $(GCOV_CFLAGS)
```
jxyang, line 13: Should protect $(GCOV_CFLAGS) with ifdef MYST_ENABLE_GCOV

scripts/myst_cc
salsal97: Not at the moment, we discussed putting in a version that would identify the files that we change using the patch and produce coverage results only for those files.

defs.mak
salsal97: We don't really need to, I kept it there to enforce this since we want to make sure that gcov flags are only used when it is enabled. Its being extra sure that future changes in the code don't accidentally use the GCOV flags, plus we had some instances in this revision as well but I am removing them

tests/elf/Makefile
salsal97: Removed it entirely, seems like it was used in the previous revision

tests/elf/Makefile
salsal97: Added this

tests/run/Makefile
salsal97: Added this

jxyang: Actually, I think tests/run belongs to a test framework that we no longer used. Can we remove the folder?

jxyang (approved):

PR #586, opened 2021-07-14
Always return read and write events for ramfs

rads-1996:
NodeJS 9 is using poll on /dev/random.

### DIFF
--- a/kernel/ramfs.c
+++ b/kernel/ramfs.c
@@ -2292,7 +2292,9 @@ static int _fs_get_events(myst_fs_t* fs, myst_file_t* file)
     if (!_ramfs_valid(ramfs) || !_file_valid(file))
         ERAISE(-EINVAL);
 
-    ret = -ENOTSUP;
+    /* Regular files always poll TRUE for reads and writes */
+    ret |= POLLIN;
+    ret |= POLLOUT;
 
 done:
     return ret;
--- a/tests/devfs/devfs.c
+++ b/tests/devfs/devfs.c
@@ -6,6 +6,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
+#include <poll.h>
 #include <stdio.h>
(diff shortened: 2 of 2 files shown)


kernel/ramfs.c
```diff
@@ -2292,7 +2292,9 @@ static int _fs_get_events(myst_fs_t* fs, myst_file_t* file)
     if (!_ramfs_valid(ramfs) || !_file_valid(file))
         ERAISE(-EINVAL);
 
-    ret = -ENOTSUP;
+    /* Regular files always poll TRUE for reads and writes */
```
vtikoo, line 2295: @mikbras does it make sense to add this to ext2 as well?

According to https://man7.org/linux/man-pages/man3/poll.3p.html -

>   Regular files shall always poll TRUE for reading and writing.

jxyang: Based on the pipeline run, it seems the changes are at least harmless. https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=9302&view=logs&j=ab477ad6-9781-5b54-82ca-a0141b26726e&t=2d9b6f4f-266f-5cad-a875-41dc232b2771

mikbras:
Hi Vikas. Yes, I think we want to do the same for EXT2. Do we have any tests for polling on files?

vtikoo:
@mikbras we don't have a test yet.

@rads-1996 could you add a test case for the nodejs 9 usage to the [existing /dev/null tests](https://github.com/deislabs/mystikos/blob/main/tests/devfs/devfs.c#L66-L88).

rads-1996:
> @rads-1996 could you add a test case for the nodejs 9 usage to the [existing /dev/random tests](https://github.com/deislabs/mystikos/blob/main/tests/devfs/devfs.c).

I have added the test case.

bodzhang (approved):

PR #590, opened 2021-07-14
Fix process-mappings bug

mikbras:
This PR is a fix for the following process-mappings bug described below.

The Bug
---------

1. Process 2 calls mmap and registers a process-mapping.
2. Process 2 calls munmap but leaves the corresponding process-mapping.
3. Process 1 calls mmap and obtains memory covered by that process-mapping.
4. Process 2 exits and calls munmap on the stale process-mapping (releasing memory now held by process 1)
5. Process 1 crashes when accessing memory within that mapping.

The Fix
--------

This PR is a half measure to resolve the bug. It introduces an **mman pids vector** region that has an element for every page in the mman region. So **pids[page_number]** either contains zero (the page is not currently mapped by any application process) or it contains the process id of the application that owns that mapped the page. The process-mapping logic is left in place for now but ``myst_release_process_mappings()`` now consults the pids vector and only releases memory if the calling process in fact owns the mapping.

The pids vector is only modified by the following syscalls: ``SYS_mmap``, ``SYS_mremap``, ``SYS_munmap``. Hence it is never modified when the kernel maps or unmaps memory for its own purposes. So if a page is mapped to the kernel, then **pids[page_number]** is zero. This is consistent with how process mappings worked before (they apply to application mappings not kernel mappings).

One remaining bug
----------------------

The current ``/proc/self/maps`` implementation can report inaccurate mappings because the process-mappings themselves get out of sync with the actual memory mappings as described above. Also process mappings do not cover ``mremap`` operations (although fd-mapped is rarely remapped and the semantics would be problematic even for POSIX implementations).

To resolve this problem, an **mman fds vector** could be introduced that keeps track of page-to-fd mappings, where **fds[page_number]** contains the page number that owns a given page. The the virtual ``/proc/self/maps`` file would scan this vector to obtain the same information.

Alternatively, the ``/proc/self/maps`` implementation could rely on the **msync mappings** instead.

Going further
---------------

The next step is to remove the process-mappings logic completely and use the **mman pids vector** to release a process's memory on exit. But a solution will be needed for the ``/proc/self/maps`` problem first (like using msync mappings instead).

The current memory manager was designed for single-process usage. The ``mmanutils`` layer attempts to augment the memory manager to accommodate multiple processes. An elegant solution will require reworking of the memory manager or adopting cross-process fork.


kernel/syscall.c
```diff
@@ -3715,7 +3717,18 @@ static long _syscall(void* args_)
                 }
             }
 
-            BREAK(_return(n, (long)myst_munmap(addr, length)));
+            long ret = (long)myst_munmap(addr, length);
```
paulcallen, line 3720: we really need to remove it from the process mapping structure. If it is a partial unmap then the process mappings need to be updated accordingly

kernel/syscall.c
```diff
@@ -3858,6 +3882,21 @@ static long _syscall(void* args_)
             ret = (long)myst_mremap(
```
paulcallen, line 3882: need to update the process mappings to remove the old and add the new... even if it is a partial remap

kernel/mmanutils.c
```diff
@@ -862,3 +884,160 @@ void myst_mman_stats(myst_mman_stats_t* buf)
...
+    /* addr must be aligned on a page boundary */
+    if ((uint64_t)addr % PAGE_SIZE)
+        ERAISE(-EINVAL);
+
+    /* the length must be a multiple of the page boundary */
+    if (length % PAGE_SIZE)
+        ERAISE(-EINVAL);
+
+    /* release every page in the mapping owned by this pid */
+    /* ATTN: optimize to handle consecutive pages by the same owner */
```
bodzhang, line 1019: Need this optimization. unmap() always triggers mprotect() OCALL. Calling unmap() page-by-page will see performance degradation.

kernel/mmanutils.c
```diff
@@ -862,3 +884,160 @@ void myst_mman_stats(myst_mman_stats_t* buf)
...
+        count = length / PAGE_SIZE;
+
+        if (index >= npids)
+            ERAISE(-ERANGE);
+
+        if (index + count > npids)
+            ERAISE(-ERANGE);
+    }
+
+    switch (op)
```
bodzhang, line 950:
```suggestion
    /* ATTN: optimize to use 64bit ops */
    switch (op)
```

bodzhang (changes requested):

kernel/syscall.c
mikbras: As noted above, this is a half-measure to workaround the crash caused by releasing memory that has since be given to another process. The full-measure will involve removing the process-mapping structures altogether (now that we have the pids[] vector). But the ``/proc/self/maps`` logic uses these mappings and needs to be reworked as well.

kernel/syscall.c
mikbras: Again, not part of the half-measure solution. We will be removing the process-mappings as noted above once we rework ``/proc/self/maps``.

kernel/mmanutils.c
mikbras: Fixed, thanks.

mikbras:
I optimized ``myst_mman_pids_set()`` for the zero-pid case so when clearing the pids[] vector we just call ``memset(.., 0, ...)``

bodzhang (approved):

bodzhang:
Please remove the patches to /tests/libcxx files before merging

PR #593, opened 2021-07-15
Enable GDB stack-walking for myst_call_on_stack

anakrish:
The implementation is moved to a function named `__morestack`
to which myst_call_on_stack jumps to. GDB treats functions
names `__morestack` in a special manner; these functions can
have split stacks.

### DIFF
--- a/asm/callonstack.s
+++ b/asm/callonstack.s
@@ -8,13 +8,25 @@
 //     [RSI] long (*func)(void* arg),
 //     [RDX] void* arg);
 //
-// Call a function on the given stack
+// Call a function on the given stack.
 //
+// GDB normally will stop stack-walking at a function if it determines that
+// the stack does not monotonically decrease at the function.
+// This could be the case for some invocations of myst_call_on_stack, if the
+// `stack` parameter is greater than RSP upon entry.
+// To support the split-stack feature, GDB however relaxes the above
+// constraint for functions named `__morestack`.
+// We leverage this GDB behavior by moving the body of myst_call_on_stack to
+// a local function name `__morestack` to which myst_call_on_stack jumps to.
 //==============================================================================
-
 .globl myst_call_on_stack
 .type myst_call_on_stack, @function
 myst_call_on_stack:
+   jmp __morestack
+.size myst_call_on_stack, .-myst_call_on_stack
+
+.type __morestack, @function
+__morestack:
 .cfi_startproc
 
     pushq %rbp
@@ -36,3 +48,4 @@ myst_call_on_stack:
     ret
 
 .cfi_endproc
+.size __morestack, .-__morestack


asm/callonstack.s
```diff
@@ -26,6 +38,7 @@ myst_call_on_stack:
     push %r14
     movq %rsp,%r14 # save original stack pointer
     movq %rdi,%rsp # set new stack pointer
+    and $-16, %rsp
```
paulcallen, line 41: what is the purpose of this?

paulcallen: rounding? If so add comment

asm/callonstack.s
```diff
@@ -8,13 +8,25 @@
...
 //
-// Call a function on the given stack
+// Call a function on the given stack.
 //
+// GDB normally will stop stack-walking at a function if it determines that
+// the stack does not monotonically decrease at the function.
+// This could be the case for some invocations of myst_call_on_stack, if the
+// `stack` parameter is greater than RSP upon entry.
+// To support the split-stack feature, GDB however relaxes the above
+// constraint for functions named `__morestack`.
```
paulcallen, line 18: This is interesting. We have a few other places in our code where we walk stacks... getting backtrace for one and our pseudo-fork implementation where we copy stack and fix-up stack frames. Wondering if we can leverage this fact in these other places to make it more robust.

anakrish: Can you point me to the places in code you are referring to and where to put a breakpoint to see how GDBs backtrace currently appears ?

asm/callonstack.s
anakrish: Will add a comment. Linux ABI requires that stack be aligned to 16 bytes. It is not clear whether we ensure elsewhere that the stack pointer being passed in is already aligned.
You can see similar stack alignment in OE's enter routines: https://github.com/openenclave/openenclave/blob/47456dbaea90425543a431bc2892cc4c0db773ef/enclave/core/sgx/enter.S#L121

anakrish: I will mostly remove this change from this PR, and create a separate PR for it if needed. It was part of the investigation I was doing for the mprotect issue (process mmap mapping).

paulcallen (approved):

asm/callonstack.s
```diff
@@ -8,13 +8,25 @@
...
+// We leverage this GDB behavior by moving the body of myst_call_on_stack to
+// a local function name `__morestack` to which myst_call_on_stack jumps to.
 //==============================================================================
-
 .globl myst_call_on_stack
 .type myst_call_on_stack, @function
 myst_call_on_stack:
+   jmp __morestack
```
paulcallen, line 25: how does gdb make this work? The return address on the stack is myst_call_on_stack still, not __morestack

jxyang:
Thanks for doing this. I remember @mikbras  tried to avoid the problem by allocating kernel stack in a specific region that is lower than the user stacks. But with PR #569, that is no longer the case, hence the need for this PR.

asm/callonstack.s
anakrish: When__morestack calls bar, the return address that gets pushed on the stack belongs to __morestack. Thus GDB split stack backtrace works.

However, since myst_call_on_stack jumps to __morestack instead of calling it, myst_call_on_stack does not appear in the backtrace.

PR #597, opened 2021-07-15
shutdown socket on shutdown

paulcallen:
when shutting down a process, if there is a blocking socket call, the
process shutdown will hang while waiting for this thread.
Sending a shutdown to the socket should be sufficient to wake it up so
we can process the signals.

### DIFF
--- a/kernel/sockdev.c
+++ b/kernel/sockdev.c
@@ -392,6 +392,11 @@ done:
     return ret;
 }
 
+static int _sd_interrupt(myst_sockdev_t* sd, myst_sock_t* sock)
+{
+    return _sd_shutdown(sd, sock, SHUT_RDWR);
+}
+
 static int _sd_getsockopt(
     myst_sockdev_t* sd,
     myst_sock_t* sock,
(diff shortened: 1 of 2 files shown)


paulcallen:
although the current fix solves a shutdown problem with the sockperf test, it breaks a number of ltp tests that are already passing

paulcallen:
Now we dont check the return code of the shutdown() when we are doing the close() things work a lot better.
Only limitation to this is IF someone has a blocking thread when close() is called an invalid error may get generated on that blocking call. LTP tests pass cleanly though.

anakrish (commented):
Do the increments and decrements of `active_calls` need to be atomic?

kernel/sockdev.c
```diff
@@ -660,6 +739,13 @@ static int _sd_close(myst_sockdev_t* sd, myst_sock_t* sock)
     if (!sd || !_valid_sock(sock))
         ERAISE(-EINVAL);
 
+    /* make sure it is shut down to make sure the socket gets unblocked */
+    if (sock->active_calls)
```
anakrish, line 743: There is a chance that `active_calls` is incremented immediately after this check returns false.

paulcallen:
This PR will be replaced by #609

kernel/sockdev.c
paulcallen: active_calls is in fact atomic. And you are right, there is a small change it will still hang. This is why we will move forward with a more substantial fix #609 which turns the synchronous calls into synchronous ones under the covers

PR #604, opened 2021-07-15
Add support for syscall_mkdirat

Sahakait:


kernel/syscall.c
```diff
@@ -1541,6 +1541,29 @@ long myst_syscall_mkdir(const char* pathname, mode_t mode)
     return ret;
 }
 
+long myst_syscall_mkdirat(int dirfd, const char* pathname, mode_t mode)
+{
+    char* abspath = NULL;
+    long ret = 0;
+
+    if (dirfd == AT_FDCWD)
+    {
```
jxyang, line 1550: I think `myst_get_absolute_path_from_dirfd` covers the case of `AT_FDCWD`

Sahakait: That's right, it does. The redundant check is now removed.

PR #609, opened 2021-07-16
EXPERIMENTAL: Mechanism for interrupting blocking socket calls

mikbras:
This PR provides a mechanism for interrupting blocking socket calls. It introduces two functions.

```
long myst_async_syscall(long num, int poll_flags, int fd, ...);
long myst_interrupt_async_syscall(int fd);
```

This ``myst_async_syscall`` function executes a file-descriptor oriented syscall using non-blocking I/O. The ``myst_interrupt_async_syscall`` function interrupts a thread that is blocking inside ``myst_async_syscall``, causing it to wake up and return ``-EINTR``. 

Prior work
------------

This PR builds on prior work by [Paul](https://github.com/paulcallen). Paul originally diagnosed to root cause of a hang in the ``sockperf`` test. The hang was caused by the following sequence of events.
- Process B calls ``recvmsg()``
- Process A calls ``close()`` on the same socket descriptor.
- Process A exits
- Process B continues to hang on ``recvmsg()``

Paul's earlier solution was to add reference counting to select socket device operations and to check whether any of these calls were active during the ``close`` operation. If so, ``shutdown()`` was called, which broke process B out of ``recvmsg``. Although this worked, it caused the peer to receive a shutdown exception instead of an end-of-file (to which the sockperf test complained). Paul intended this as a quick fix for the last release and suggested using asynchronous I/O operations (this PR).

Paul's earlier solution can be found here: [PR 597 (https://github.com/deislabs/mystikos/pull/597/files).

The new solution is similar except we call ``myst_interrupt_async_syscall()``.

### DIFF
--- a/include/myst/asyncsyscall.h
+++ b/include/myst/asyncsyscall.h
@@ -8,7 +8,7 @@
 
 // Perform an fd-oriented "asynchronous" syscall that can be interrupted by
 // calling myst_interrupt_async_syscall().
-long myst_async_syscall(long num, int fd, ...);
+long myst_async_syscall(long num, int poll_flags, int fd, ...);
 
 // Interrupt the invocation of myst_async_syscall() that is currently blocked
 // on the given fd, forcing myst_async_syscall() to return -EINTR.
--- a/include/myst/asyncsyscall.h
+++ b/include/myst/asyncsyscall.h
@@ -6,12 +6,12 @@
 
 #include <stddef.h>
 
-// Perform an fd-oriented "asynchronous" syscall that can be interrupted by
-// calling myst_interrupt_async_syscall().
+// Perform an fd-oriented syscall that can be asynchronously interrupted by
+// myst_interrupt_async_syscall().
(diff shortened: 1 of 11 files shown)


mikbras:
Note from discussion with Paul:
- two threads (reader/write) can put the same pipe into two different poll sets and both can read from the same pipe. One pipe will miss the wakeup (the wakers should be keyed by syscall-number + file-descriptor).
- Consider atexit() handler to close all waker pipes.

paulcallen (approved):

mikbras:
Performance impact
----------------------

sockperf test before
----------------------
sockperf: Total of 13215 messages sent in 1.002 sec
sockperf: Total of 14890 messages sent in 1.001 sec

sockperf test after
--------------------
sockperf: Total of 12274 messages sent in 1.020 sec
sockperf: Total of 13539 messages sent in 1.001 sec

summary
----------
The overall performance impact is between 7% and 9%.

mikbras:
Closing in favor of signal based approach.

Issue #512, opened 2021-06-28
Nodejs: web server sample stuck in a epoll_wait loop


rads-1996:
Hello,

I am trying to build a confidential app using Nodejs and Mystikos. I am not able to run even a simple Nodejs app.
This is my Nodejs file:

```
const http = require('http');

const hostname = '127.0.0.1';
const port = 3000;

const server = http.createServer((req, res) => {
res.statusCode = 200;
res.setHeader('Content-Type', 'text/plain');
res.end('Hello World');
});

server.listen(port, hostname, () => {
console.log(`Server running at http://${hostname}:${port}/`);
});

```
and this is my Dockerfile:

```
FROM node:11-alpine

WORKDIR /app
COPY app.js /app/app.js
```

These are the instructions I am using to run it with Mystikos:
```
myst-appbuilder Dockerfile
myst mkcpio appdir rootfs
myst exec-linux rootfs /usr/local/bin/node /app/app.js

```
I see this error message:

`Segmentation fault (core dumped)
`

Would appreciate any pointers on how to debug this further.

paulcallen:
try running with `--strace` and `--etrace` to see if this is due to some unsupported syscall.
We have not looked at Java or Javascript yet. Current focus has been on c/c++, .net and python. We will investigate more going forwards.

rads-1996:
Thanks for getting back. I tried running with strace and got this error:

```
=== SYS_prlimit64(pid=0, resource=3, new_rlim=(nil), old_rlim=0x7f57376d6940): tid=101   
SYS_prlimit64(): return=-EINVAL(-22): tid=101
=== SYS_rt_sigprocmask(how=0 set=0x7f57377bb318 oldset=0x7f57376d6850): tid=101   
SYS_rt_sigprocmask(): return=0(0): tid=101
=== SYS_tkill(tid=101 sig=6): tid=101   
SYS_tkill(): return=0(0): tid=101
=== SYS_rt_sigprocmask(how=2 set=0x7f57376d6850 oldset=(nil)): tid=101   
SYS_rt_sigprocmask(): return=0(0): tid=101
Makefile:23: recipe for target 'run' failed
make: *** [run] Segmentation fault (core dumped)
```

Fetching prlimit64 for RLIMIT_STACK is throwing an error. Any ideas about what to do next?

bodzhang:
The crash looks like a bug in the abort signal (sig=6) handling. @mikbras is looking into it.

[...]

PR #615, opened 2021-07-20
Poll revent fixes

bodzhang:
The PR addresses 3 issues:

- The existing implementation can return event not requested by the poll syscall (not part of the pollfd.events mask), which can cause problem such as issue #512. The PR makes sure the revents returned can include any of those specified in events, or one of the values POLLERR, POLLHUP, or POLLNVAL, following [Linux manual](https://www.man7.org/linux/man-pages/man2/poll.2.html).
- Added poll support for ext2fs open files, always returning POLLIN|POLLOUT, in the same manner as ramfs handling in PR #586 
- Return POLLNVAL event on closed or invalid fds

### DIFF
--- a/kernel/poll.c
+++ b/kernel/poll.c
@@ -45,10 +45,13 @@ long _poll_kernel(struct pollfd* fds, nfds_t nfds)
 
         if ((events = (*fdops->fd_get_events)(fdops, object)) >= 0)
         {
-            fds[i].revents = events;
-
-            if (events)
+            /* Only report events requested or POLLERR, POLLHUP and POLLNVAL*/
+            if (events =
+                    events & ((fds[i].events) | (POLLERR | POLLHUP | POLLNVAL)))
+            {
+                fds[i].revents = events;
                 total++;
+            }
         }
         else if (events != -ENOTSUP)
         {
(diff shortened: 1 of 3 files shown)


jxyang (approved):

kernel/poll.c
```diff
@@ -45,10 +45,13 @@ long _poll_kernel(struct pollfd* fds, nfds_t nfds)
 
         if ((events = (*fdops->fd_get_events)(fdops, object)) >= 0)
         {
-            fds[i].revents = events;
-
-            if (events)
+            /* Only report events requested or POLLERR, POLLHUP and POLLNVAL*/
+            if (events =
```
mikbras, line 49: Nit: maybe use "events &=" to shorten

kernel/poll.c
```diff
@@ -45,10 +45,13 @@ long _poll_kernel(struct pollfd* fds, nfds_t nfds)
...
+                    events & ((fds[i].events) | (POLLERR | POLLHUP | POLLNVAL)))
```
mikbras, line 50: Hmm? I missed this before from the man page. :)

```
       This field may be specified as zero, in which case the only events that
       can  be  returned  in  revents  are POLLHUP, POLLERR, and POLLNVAL (see
       below).
```

mikbras (approved):
Nice clean change. Just one nit (that does not really need to be addressed).

PR #618, opened 2021-07-20
Pipe fix and a python subprocess test

bodzhang:
The PR fixed the following issues:

- The current pipe implementation returns EPIPE on read() after the write side closes the pipe. According to Linux manual, in this case, end-of-file (value of 0) should be returned. 
- The current implementation of _pd_get_event()  does not set any event when the other end of the pipe is closed. 

The PR also adds a python_subprocess test. The test invokes python.subprocess module to create a child process to execute `ls` command, and returns the result to the parent process through pipes. The code flow exercises the relevant pipe scenario.

### DIFF
--- a/kernel/pipedev.c
+++ b/kernel/pipedev.c
@@ -173,7 +173,7 @@ static ssize_t _pd_read(
         /* block here while the pipe is empty */
         while (p->nbytes == 0)
         {
-            /* Handle non-blocking write */
+            /* Handle non-blocking read */
             if (pipe->flags & O_NONBLOCK)
             {
                 _unlock(pipe);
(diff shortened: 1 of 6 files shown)


kernel/pipedev.c
```diff
@@ -290,23 +285,33 @@ static ssize_t _pd_write(
...
+    /* The writer intends to write count bytes. Update wrsize immediately to
+     * signal to the reader about the intention, so a blocking reader won't bail
+     * out prematurely due 0 wrsize, when a writer will write more to the pipe
+     * soon. If the call returns with less than count bytes writtern, the caller
+     * might invoke the call again. For example, a non-blocking write returning
+     * EAGAIN did not write any data to the pipe. The caller likely will call
+     * again indicating count bytes to write. In those cases, adjust wrsize
+     * before returning to make sure no double counting
+     * */
+    p->wrsize += count;
```
jxyang, line 304: If we are here, it means the writer holds the lock, and there won't be readers checking `wrsize` at the same time. Why don't we increment `wrsize` when we actually know how many bytes are written to the pipe buffer?

bodzhang: The writer will release a lock when the pipe is full and wait for other threads to read the pipe (line 345 in the new code).

jxyang (approved):

PR #622, opened 2021-07-21
Separate oesignlib into a separate OE directory

mikbras:
The change will be seen in the ``mystikos.v4.oesignlib`` OE branch.


jxyang:
As discussed, please submit a PR against OE mystikos.v5 branch.

PR #627, opened 2021-07-21
Patch libc with argz* functions

jxyang:


crt/glibc_stub.c
```diff
@@ -62,6 +64,33 @@ int __argz_create_sep(const char* string, int delim, char** argz, size_t* len)
...
+weak_alias(__argz_append, argz_append);
+
+error_t __argz_insert(
+    char** argz,
+    size_t* argz_len,
+    char* before,
+    const char* entry)
+{
+    assert(0);
+    return 0;
```
bodzhang, line 85: 0 return indicate success.
```suggestion
    return ENOMEM;
```

crt/glibc_stub.c
```diff
@@ -62,6 +64,33 @@ int __argz_create_sep(const char* string, int delim, char** argz, size_t* len)
...
 weak_alias(__argz_create_sep, argz_create_sep);
 
+error_t __argz_append(
+    char** argz,
+    size_t* argz_len,
+    const char* buf,
+    size_t buf_len)
+{
+    assert(0);
+    return 0;
```
bodzhang, line 74: 0 rerun indicates success.
```suggestion
    return ENOMEM;
```
Also on line 60, to be consistent, and for readability, suggest to use `error_t` return type.

tests/dlopen/dlopen.c
```diff
@@ -31,9 +31,17 @@ int call_foo(const char* argv0)
 
 int main(int argc, const char* argv[])
 {
+    void* handle;
     int ret = call_foo(argv[0]);
     assert(ret == 12345);
 
+    if (!(handle = dlopen("/libodbc.so.2.0.0", RTLD_NOW)))
```
bodzhang, line 38: A comment on why this specific libodbc.so is chosen will be helpful.

bodzhang (changes requested):

bodzhang (approved):

PR #632, opened 2021-07-22
Delayed zero-fill for mmap(prot=0) pages

bodzhang:
.NET runtime, Java runtime, and potentially other SW stack uses mmap(addr= preferred_addr, prot=0) to probe availability of preferred memory allocation. If the preferred allocation is not available (indicated by the returned memory addr not matching the preferred addr), the runtime might immediately call munmap() on the return address. 

With the current Mystikos mmap() implementation, when the kernel allocated the memory as part of mmap flow, the kernel sets the memory content to 0. When the EPC size of the platform is small, the operation will take a long time due to EPC paging. 

As the app runtime is simply probing the availability of the preferred memory allocation, and will deallocate the memory immediately if the allocated memory is not at the preferred address, zeroing the content and its associated extra latency do not yield real value.

This PR marks the newly allocated inaccessible memory (by mmap(prot=0)) as "pending zero-fill", and only zero-fill the memory on a following mprotect() call that makes the memory accessible.

### DIFF
--- a/kernel/mman.c
+++ b/kernel/mman.c
@@ -117,6 +117,7 @@
 #include <myst/defs.h>
 #include <myst/fsgs.h>
 #include <myst/mman.h>
+#include <myst/panic.h>
 #include <myst/round.h>
 #include <myst/spinlock.h>
 #include <myst/strings.h>
@@ -124,6 +125,8 @@
 #include <myst/tcall.h>
 #include <stdio.h>
 #include <string.h>
+
+#define MYST_PENDING_ZEROING_FLAG 0x80
 
 /*
 **==============================================================================
@@ -410,7 +413,8 @@ static uintptr_t _mman_find_gap(
 
     /* No gaps in linked list so obtain memory from mapped memory area */
     {
-        uintptr_t start = mman->map - size;
+        // Avoid integer underflows when the requested size is huge.
+        uintptr_t start = mman->map >= size ? mman->map - size : 0;
 
         /* If memory was exceeded (overrun of break value) */
         if (!(mman->brk <= start))
@@ -454,7 +458,15 @@ done:
             (LEN) / PAGE_SIZE);                                \
     }
 
-/* check and get permission tracking, if insisitent, return -1*/
+/*  Check page permission consistency and report the permission value
+ *  OUTPUT
+ *      prot_out - the permission value of the first page in the memory range
+ *  RETURN
+ *      0: consistent
+ *      1: inconsistent, but only differ on MYST_PENDING_ZEROING_FLAG
+ *      -1: inconsistent
+ *
+ */
 static int _mman_get_prot(
     uint8_t* prot_vector,
     uint32_t offset,
@@ -462,10 +474,39 @@ static int _mman_get_prot(
     int* prot_out)
 {
     const uint8_t* start = prot_vector + offset;
+    const uint8_t* end = start + num_pages;
+    const uint8_t* p = start;
+    const uint8_t prot = (*p & ~MYST_PENDING_ZEROING_FLAG);
+    const uint8_t protz = (*p | MYST_PENDING_ZEROING_FLAG);
 
-    *prot_out = start[0];
+    /* set the output protection parameter */
+    *prot_out = *p++;
 
-    return (myst_memcchr(start + 1, start[0], num_pages - 1) == NULL) ? 0 : -1;
+    int ret = 0;
+
+    while (p < end)
+    {
+        if (*p == prot)
+        {
+            if ((p = myst_memcchr(p, prot, end - p)) == NULL)
+                return ret;
+        }
+        else if (*p == protz)
+        {
+            if ((p = myst_memcchr(p, protz, end - p)) == NULL)
+                return ret;
+        }
+        else
+        {
+            return -1;
+        }
+        /* if matching exit didn't happen in the first pass of the while loop,
+         * full range is not consistent. Set ret to indicate mixed flag, which
+         * will be returned in later matching exit  */
+        ret = 1;
+    }
+    /* Only reach here if num_pages=1.*/
+    return 0;
 }
 
 static int _munmap(myst_mman_t* mman, void* addr, size_t length)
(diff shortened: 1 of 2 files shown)


kernel/mman.c
```diff
@@ -617,6 +619,66 @@ static int _munmap(myst_mman_t* mman, void* addr, size_t length)
...
+    {
+        if (myst_tcall_mprotect(addr, length, prot))
+        {
+            _mman_set_err(mman, "mprotect tcall failed");
+            return -EINVAL;
+        }
+        /* Don't clear any pending zero-fill flag */
+        for (i = 0; i < npages; i++)
+        {
+            mman->prot_vector[start_page_index + i] &=
```
jxyang, line 675: Are we delaying the "delayed 0-fill" to a future `mprotect`? It takes me a while to figure out the intention. Please add comment.

jxyang: Is the loop equivalent to `memset(&mman->prot_vector[start_page_index], prot & MYST_PENDING_ZEROING_FLAG, npages);`? or `_MMAN_SET_PAGES_PROT(mman, addr, length, prot & MYST_PENDING_ZEROING_FLAG)`?

mikbras:
I have two main concerns:
- The performance may suffer because the vectors that keeps track protections is the same as the vector that keeps track of the delayed-zero-fill flag. 
- I am not sure (yet) that the page-protection consistency checking will be correct under all circumstances.

I am working with the branch now to get more insight.

jxyang:
This is the one line change that addresses integer underflow that could be potentially combined in this PR.

https://github.com/deislabs/mystikos/pull/639

kernel/mman.c
```diff
@@ -617,6 +619,66 @@ static int _munmap(myst_mman_t* mman, void* addr, size_t length)
...
+        memset(&mman->prot_vector[start_page_index], prot, npages);
+    }
+    else
+    {
+        if (myst_tcall_mprotect(addr, length, prot))
+        {
+            _mman_set_err(mman, "mprotect tcall failed");
+            return -EINVAL;
+        }
+        /* Don't clear any pending zero-fill flag */
```
mikbras, line 672: Where would it have been cleared?

kernel/mman.c
```diff
@@ -617,6 +619,66 @@ static int _munmap(myst_mman_t* mman, void* addr, size_t length)
...
+    int prot)
+{
+    uint32_t i;
+    uint32_t start_page_index = ((uintptr_t)addr - mman->start) / PAGE_SIZE;
+    uint32_t npages = length / PAGE_SIZE;
+    bool w_access_requested = false;
+
+    if (prot != MYST_PROT_NONE)
+    {
+        for (i = 0; i < npages; i++)
```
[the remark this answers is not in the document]
bodzhang: The delayed_zero_fill flag might not be consistent in the whole range. Had to check for that condition. The update PR does memset over consecutive sections with the flag set.

kernel/mman.c
```diff
@@ -617,6 +619,66 @@ static int _munmap(myst_mman_t* mman, void* addr, size_t length)
...
+                        return -EINVAL;
+                    }
+                    w_access_requested = true;
+                }
+                memset((void*)((uintptr_t)addr + i * PAGE_SIZE), 0, PAGE_SIZE);
+            }
+        }
+
+        if ((!(prot & MYST_PROT_WRITE) && w_access_requested) ||
+            (!w_access_requested))
```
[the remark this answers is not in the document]
bodzhang: Added comment.

kernel/mman.c
bodzhang: Added comment. The loop preserve elements of MYST_PENDING_ZEROING_FLAG and set other elements to MYST_PROT_NONE (value 0). It can't be replaced by one memset().

mikbras:
Still looking.

kernel/mman.c
```diff
@@ -1448,23 +1526,31 @@ int myst_mman_mremap(
         if (_end(vad) == old_end && _get_right_gap(mman, vad) >= delta)
         {
             vad->size += (uint32_t)delta;
-            /* Set W permission first before zeroing */
-            if (myst_tcall_mprotect(
-                    (void*)(start + old_size), delta, (prot | MYST_PROT_WRITE)))
+            /* If the old area is pending zero fill, the expanded area gets the
```
mikbras, line 1529: The VAD may not have consistent protections for every page I think.

mikbras: Never mind, I see that prot consistency was checked above.

mikbras:
It seems to me that the ``memcpy`` in remap would fail if the old memory writable but not readable. The ``memcpy`` would fail while reading the soruce buffer.

bodzhang (commented):
@jxyang, PR #639 change is adopted in the update to this PR.

kernel/mman.c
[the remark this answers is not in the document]
bodzhang: A `mprotect(prot=0)` call will not change the prot_vector value of `(MYST_PENDING_ZEROING_FLAG | 0)`.  With `&=` operation, any r/w/x bits will be cleared, but MYST_PENDING_ZEROING_FLAG will remain, if present. Added comments in the updated PR:
```c
        /* prot = MYST_PROT_NONE. The pages are still not accessible, don't
         * zero-fill the pages. Don't clear any pending zero-fill flag either,
         * so a later mprotect() enabling access will do the zero-fill */
```

bodzhang:
> It seems to me that the `memcpy` in remap would fail if the old memory writable but not readable. The `memcpy` would fail while reading the soruce buffer.

@mikbras, thanks for spotting this bug. It has been there for a while. Will fix in next update to the PR.

mikbras:
I think a remap would fail to find the protection vector consistent under these circumstances:
- mmap() 8 pages with PROT_NONE
- mprotect() the first 4 pages with PROT_READ|PROT_WRITE
- mprotect() the first 4 pages with PROT_NONE
- mremap() all 8 pages

I think the final mremap() will fail because:
- the first 4 pages will be PROT_NONE
- the last 4 pages will be PROT_NONE | MYST_PENDING_ZEROING_FLAG

bodzhang:
> I think a remap would fail to find the protection vector consistent under these circumstances:
> 
> * mmap() 8 pages with PROT_NONE
> * mprotect() the first 4 pages with PROT_READ|PROT_WRITE
> * mprotect() the first 4 pages with PROT_NONE
> * mremap() all 8 pages
> 
> I think the final mremap() will fail because:
> 
> * the first 4 pages will be PROT_NONE
> * the last 4 pages will be PROT_NONE | MYST_PENDING_ZEROING_FLAG

@mikbras, I realized this corner case too. Will explore possible solutions.

mikbras:
Here is a fast version of ``_mman_get_prot()`` that works with mixed zeroing-flags. Maybe we should just ``protz``.

```C++
static int _mman_get_prot(
    uint8_t* prot_vector,
    uint32_t offset,
    uint32_t num_pages,
    int* prot_out)
{
    const uint8_t* start = prot_vector + offset;
    const uint8_t* end = start + num_pages;
    const uint8_t* p = start;
    const uint8_t prot = (*p & ~MYST_PENDING_ZEROING_FLAG);
    const uint8_t protz = (*p | MYST_PENDING_ZEROING_FLAG);

    /* set the output protection parameter */
    *prot_out = *p++;

    while (p < end)
    {
        if (*p == prot)
        {
            if ((p = myst_memcchr(p, prot, end - p)) == NULL)
                return 0;
        }
        else if (*p == protz)
        {
            if ((p = myst_memcchr(p, protz, end - p)) == NULL)
                return 0;
        }
        else
        {
            return -1;
        }
    }

    return 0;
}
```

mikbras (approved):
Looks great Bo.

PR #633, opened 2021-07-22
Add SYS_sync support

Francis-Liu:


ext2/ext2.c
```diff
@@ -5555,6 +5555,12 @@ static int _ext2_fsync_and_fdatasync(myst_fs_t* fs, myst_file_t* file)
     return ret;
 }
 
+static void _ext2_sync(myst_fs_t* fs)
+{
+    // ramfs being a in-memory fs, treat fsync and datasync as NOP
```
vtikoo, line 5560:
```suggestion
    // ext2fs being a in-memory fs, treat sync as NOP
```

kernel/ramfs.c
```diff
@@ -2644,6 +2644,12 @@ static int _fs_fsync_and_fdatasync(myst_fs_t* fs, myst_file_t* file)
     return ret;
 }
 
+static void _fs_sync(myst_fs_t* fs)
+{
+    // ramfs being a in-memory fs, treat fsync and datasync as NOP
```
vtikoo, line 2649:
```suggestion
    // ramfs being a in-memory fs, treat sync as NOP
```

kernel/fdtable.c
```diff
@@ -570,3 +570,30 @@ int myst_fdtable_list(const myst_fdtable_t* fdtable)
...
+    myst_spin_lock(&fdtable->lock);
+    {
+        for (int i = 0; i < MYST_FDTABLE_SIZE; i++)
+        {
+            const myst_fdtable_entry_t* entry = &fdtable->entries[i];
+
+            if (entry->type == MYST_FDTABLE_TYPE_FILE)
+            {
+                myst_fs_t* fs = entry->device;
+                fs->fs_sync(fs);
```
vtikoo, line 590: For hostfs, the subsequent calls after the first one will be redundant.

vtikoo: An alternate approach could be to call sync once on the host, if there is atleast one hostfs file open.

Francis-Liu: I was thinking there could be multiple instances of hostfs. Like one hostfs rootfs plus several mounted hostfs. Let me think about if there is a way to distinguish those hostfs.

vtikoo:
Also just to point out, for hostfs calling `SYS_sync` on the host will probably affect files outside of the mount point. Not sure if this can have any repercussions.

Francis-Liu:
> Also just to point out, for hostfs calling `SYS_sync` on the host will probably affect files outside of the mount point. Not sure if this can have any repercussions.

I looked at sync() implementation. It doesn't distinguish files to sync on. But it only sync inodes that are belonging to the calling process.

vtikoo:
> I looked at sync() implementation. It doesn't distinguish which files to sync on. But it only sync inodes that are belonging to the calling process.

Then it wouldn't be differentiate between files opened by the host process vs the Mystikos kernel. It sounds like what we really want is calling fsync on all open hostfs fd's.

Francis-Liu:
> > I looked at sync() implementation. It doesn't distinguish which files to sync on. But it only sync inodes that are belonging to the calling process.
> 
> Then it wouldn't differentiate between files opened by the host process vs the Mystikos kernel. It sounds like what we really want is calling fsync on all open hostfs fd's.

What about fdatasync?

vtikoo:
I thought fsync(file data + metadata) is a superset of fdatasync(file data only).

vtikoo (approved):

PR #634, opened 2021-07-23
cpython test suite fixes and cleanup

vtikoo:
### Summary
- epoll fcntl fix.
- Support for __explicit_bzero_chk glibc symbol.
- Update test list.
- Test setup cleanup.


tests/cpython-tests/exec.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+make run-list TESTFILE=tests.passed
```
jxyang, line 3: I think we can fold these two lines into Makefile.

tests/cpython-tests/Makefile
```diff
@@ -23,20 +24,29 @@ ext2fs: appdir
...
 
-run-str:
-	$(MYST_EXEC) $(OPTS) ext2fs /cpython/python -c "$(PROG_STR)"
+run-list: $(FS)
+	$(MYST_EXEC) $(OPTS) $(FS) /cpython/python -m test -f $(TESTFILE) --timeout 30 -q
 
-pdb:
-	$(RUNTEST) $(MYST_EXEC) $(OPTS) ext2fs /cpython/python -m pdb /cpython/Lib/test/test_grammar.py -v 
+
+run-single: $(FS)
```
jxyang, line 42: The command to run a single test from other test suites are like: `make one TEST=<path-to-test>`.

tests/cpython-tests/Makefile
```diff
@@ -23,20 +24,29 @@ ext2fs: appdir
...
+	$(RUNTEST) $(MYST_EXEC) $(OPTS) $(FS) /cpython/python -m test $(TC) -q
+
+run-str: $(FS)
```
jxyang, line 45: `run-str` is not used anymore?

tests/cpython-tests/Makefile
```diff
@@ -23,20 +24,29 @@ ext2fs: appdir
...
-	$(RUNTEST) $(MYST_EXEC) $(OPTS) ext2fs /cpython/python -m test -f tests.passed.1 -q
+	$(RUNTEST) ./exec.sh
 
-run-single:
-	$(RUNTEST) $(MYST_EXEC) $(OPTS) ext2fs /cpython/python -m test $(TC) -v
+#########################
+##	Dev targets
+#########################
+hostfs:
+	cp -r appdir/ hostfs/
```
jxyang, line 36: hostfs is not used anymore?

tests/cpython-tests/Makefile
```diff
@@ -23,20 +24,29 @@ ext2fs: appdir
...
+run-str: $(FS)
+	$(MYST_EXEC) $(OPTS) $(FS) /cpython/python -c "$(PROG_STR)"
+
+pdb: $(FS)
+	$(RUNTEST) $(MYST_EXEC) $(OPTS) $(FS) /cpython/python -m pdb /cpython/Lib/test/test_grammar.py -v
 
 ct:
-	rm -rf /tmp/myst*
\ No newline at end of file
+	rm -rf /tmp/myst* @test_101*
```
jxyang, line 52: `ct` is not used anymore?

tests/cpython-tests/run-tests.sh
```diff
@@ -1,25 +1,26 @@
 #!/bin/bash
 TEST_FILE=tests.all
+if [[ ("$#" -eq 1) && (-f $1) ]]; then
```
jxyang, line 3: This file is called `run_tests_locally.sh` in other test suites.

tests/cpython-tests/exec.sh
vtikoo: runtest doesn't work well in case of multiple commands. We use this exec.sh trick in a few other tests as well.

tests/cpython-tests/Makefile
vtikoo: This is a dev target. Useful to run small python programs on cmdline with Mystikos.

tests/cpython-tests/Makefile
vtikoo: Again a dev target. Hostfs is useful for debugging as you can change .py files and not have to rebuild the rootfs.

tests/cpython-tests/Makefile
vtikoo: This is a dev target for cleaning up leaked resources.

tests/cpython-tests/run-tests.sh
vtikoo: Renamed.

tests/cpython-tests/Makefile
vtikoo: Renamed.

tests/cpython-tests/Makefile
jxyang: I understand the conveniences. But if we little the code base in the main branch with dev-specific code, it will take much longer to make a release due to clean ups.

When a test is added, we should only work on it when there is a regression. That's why the "dev targets" are not that useful. 

Please keep the dev targets locally.

tests/cpython-tests/Dockerfile
```diff
@@ -1,6 +1,4 @@
 FROM vtikoo/cpython:3.8-ubuntu as build
```
jxyang, line 1: Why don't we use the standard python:3.8-slim docker image?

If there is a need for a custom image, maybe for debugging purpose, please use the team's docker hub repo. https://hub.docker.com/orgs/mystikos

vtikoo: > Why don't we use the standard python:3.8-slim docker image?

Running all the test suites requires a python build.

tests/cpython-tests/Makefile
vtikoo: There are still quite some failing test suites. The dev targets may aid in debugging those.

jxyang: The title of this PR is "cpython test suite fixes and cleanup". If we still keep the dev targets, that's not a clean up IMO.

tests/cpython-tests/Dockerfile
jxyang: We prefer building docker images from docker files over pulling them from external sources except when it takes a long time to build from the source. If we have to build from cpython source to get the test suite, let's do that. It doesn't take very long as I recall.

Even if we have to pull docker images from an external source, the preferred location is the team's docker hub repo, as mentioned above.

vtikoo: Per offline discussion, instead of pulling an external prebuilt image, changed to building entirely from a Dockerfile.

jxyang (approved):

PR #635, opened 2021-07-23
mmap and mremap fixes

vtikoo:
- Preserve ENOMEM error return behavior for mmap.
- Fix pids ownership check in mremap.

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -3691,7 +3691,12 @@ static long _syscall(void* args_)
             /* this can return (void*)-errno */
             long ret = (long)myst_mmap(addr, length, prot, flags, fd, offset);
 
-            if (ret > 0)
+            // ATTN : always translating errors to ENOMEM is probably incorrect
+            if (ret <= 0)
+            {
+                ret = -ENOMEM;
+            }
+            else if (ret > 0)
             {
                 pid_t pid = myst_getpid();
                 void* ptr = (void*)ret;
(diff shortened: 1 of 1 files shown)


vtikoo:
Nightly pipeline run - https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=9503&view=results

kernel/syscall.c
```diff
@@ -3691,7 +3691,12 @@ static long _syscall(void* args_)
             /* this can return (void*)-errno */
             long ret = (long)myst_mmap(addr, length, prot, flags, fd, offset);
 
-            if (ret > 0)
+            // ATTN : always translating errors to ENOMEM is probably incorrect
+            if (ret <= 0)
```
bodzhang, line 3695:
```suggestion
            if (ret == -1 || !ret)
```

kernel/syscall.c
```diff
@@ -3691,7 +3691,12 @@ static long _syscall(void* args_)
             /* this can return (void*)-errno */
             long ret = (long)myst_mmap(addr, length, prot, flags, fd, offset);
 
-            if (ret > 0)
+            // ATTN : always translating errors to ENOMEM is probably incorrect
```
bodzhang, line 3694:
```suggestion
            // ATTN : temporary workaround for myst_mmap()  inaccurate return value issue
```

bodzhang (changes requested):

kernel/syscall.c
vtikoo: @bodzhang does mman return only -1 or 0 for failure?

bodzhang: It can return many other negative value. The old code only translate -1/0 to ENOMEM

bodzhang (approved):

PR #636, opened 2021-07-23
Use wrfsbase/wrgsbase on Linux target

mikbras:
When run under the Linux target, the ``redis`` sample suffers a 75% performance penalty when compared with SGX. This is due to the overhead of executing the following system calls.

```
myst_syscall2(SYS_arch_prctl, ARCH_SET_FS, fsbase);
myst_syscall2(SYS_arch_prctl, ARCH_GET_FS);
```

This PR detects whether ``/proc/cpuinfo`` defines the ``fsgsbase`` flag. If so, then the Linux uses the **WRFSBASE** instruction rather than the syscall.

### DIFF
--- a/kernel/fsgs.c
+++ b/kernel/fsgs.c
@@ -12,61 +12,46 @@
 
 void myst_set_fsbase(void* p)
 {
-    if (__options.have_syscall_instruction)
+    if (__myst_kernel_args.have_fsgsbase_instructions)
     {
-        const long n = SYS_arch_prctl;
-        myst_syscall2(n, ARCH_SET_FS, (long)p);
+        __asm__ volatile("wrfsbase %0" ::"r"(p));
+    }
+    else if (__options.have_syscall_instruction)
+    {
+        myst_syscall2(SYS_arch_prctl, ARCH_SET_FS, (long)p);
     }
     else
     {
-        __asm__ volatile("wrfsbase %0" ::"r"(p));
+        myst_panic("unsupported");
     }
 }
 
 void* myst_get_fsbase(void)
 {
-    if (__options.have_syscall_instruction)
-    {
-        const long n = SYS_arch_prctl;
-        void* p;
-        myst_syscall2(n, ARCH_GET_FS, (long)&p);
-        return p;
-    }
-    else
-    {
-        void* p;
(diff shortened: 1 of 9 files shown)


kernel/fsgs.c
```diff
@@ -12,61 +12,46 @@
...
-        return p;
-    }
-    else
-    {
-        void* p;
-        __asm__ volatile("mov %%fs:0, %0" : "=r"(p));
-        return p;
-    }
+    void* p;
+    __asm__ volatile("mov %%fs:0, %0" : "=r"(p));
```
anakrish, line 32: To be accurate, this need to be `rdfsbase` instruction.

kernel/fsgs.c
```diff
@@ -12,61 +12,46 @@
...
-        return p;
-    }
-    else
-    {
-        void* p;
-        __asm__ volatile("mov %%gs:0, %0" : "=r"(p));
-        return p;
-    }
+    void* p;
+    __asm__ volatile("mov %%gs:0, %0" : "=r"(p));
```
anakrish, line 55: To be accurate, this needs to be `rdgsbase` instruction. GS[0] may not always be equal to GS base.

kernel/fsgs.c
mikbras: Thank you Anand, I considered that too but we would need to emulate RDFSBASE in OE too, which we don't. We are relying on the fact that in all of our scenarios, we expect to find a structure first field contains the address of the thread pointer itself.

kernel/fsgs.c
mikbras: Fixed. Thank you.

mikbras:
I updated the PR to test for FSGSBASE support through a SIGILL handler. The old method was wrong (looking for fsgsbase inside ``/proc/cpuinfo``) and only checked for CPU support of FSGSBASE instructions and not user-mode support.

mikbras:
I added the ``myst fsgsbase`` command so users can easily determine whether their platform supports FSGSBASE instructions. For example:

```
$ myst fsgsbase
FSGSBASE instructions are supported
```

anakrish (approved):

anakrish (approved):

jxyang (approved):

PR #644, opened 2021-07-27
[DO NOT MERGE] MAP_SHARED flag

vtikoo:
Maintain a list of shared mappings. If an existing mapping with the same
(filename, offset, length) triple is encountered in mmap, instead of
creating a new mapping the existing mapping is returned.

On unmapping, the reference count of the shared region is decremented.
The mapping is released when the reference count drops to zero.

While releasing process mappings on process exit, pid vector ownership
check is avoided for shared mappings. myst_munmap is called directly,
where the reference count for the shared region is decremented.

Write back to file is currently performed on last munmap.


kernel/mmanutils.c
```diff
@@ -261,6 +276,63 @@ void* myst_mmap(
...
+        }
+        myst_spin_lock(&_shared_mappings_lock);
+        shared_mapping_t* sm = (shared_mapping_t*)_shared_mappings.head;
+        while (sm)
+        {
+            if ((strcmp(realpath, sm->path) == 0) && offset == sm->offset &&
+                length == sm->length)
+            {
+                sm->nusers++;
+                myst_spin_unlock(&_shared_mappings_lock);
```
jxyang, line 312: Can we support write through here? According to the main page:
>        MAP_SHARED
              Share this mapping.  Updates to the mapping are visible to
              other processes mapping the same region, and (in the case
              of file-backed mappings) are carried through to the
              underlying file.

PR #648, opened 2021-07-28
[DO NOT MERGE] Set & compare CONFIG_ID for icelake

Francis-Liu:
Basic logic is ready. Will add more test cases.

### DIFF
--- a/tools/myst/common.c
+++ b/tools/myst/common.c
@@ -1,12 +1,16 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include "shared.h"
 
 #include <myst/args.h>
+#include <myst/eraise.h>
+#include <myst/hex.h>
+#include <myst/sha256.h>
 
 int myst_expand_size_string_to_ulong(const char* size_string, size_t* size)
 {
@@ -185,4 +189,18 @@ bool myst_validate_mount_config(myst_mounts_config_t* mounts)
     }
 
     return true;
+}
+
+int myst_generate_config_id(
+    const char* augmented_app_config_buf,
+    size_t augmented_app_config_size,
+    uint8_t* config_id)
+{
+    int ret = 0;
+    myst_sha256_t sha256;
+    ECHECK(myst_sha256(
+        &sha256, augmented_app_config_buf, augmented_app_config_size));
+    memcpy(config_id, sha256.data, sizeof(sha256.data));
+done:
+    return ret;
 }
--- a/tools/myst/enc/enc.c
+++ b/tools/myst/enc/enc.c
@@ -9,6 +9,8 @@
 #include <string.h>
 #include <sys/mount.h>
 
+#include <openenclave/attestation/attester.h>
+#include <openenclave/attestation/sgx/evidence.h>
 #include <openenclave/enclave.h>
 
 #include <elf.h>
@@ -48,6 +50,11 @@ static bool _trace_syscalls = false;
 extern const void* __oe_get_heap_base(void);
 
 long myst_tcall_isatty(int fd);
+static int _get_config_id(uint8_t* out_config_id);
+static int _check_config_id(
+    const uint8_t* config_id,
+    const uint8_t* augmented_app_config_buf,
+    size_t augmented_app_config_size);
 
 long _exception_handler_syscall(long n, long params[6])
 {
@@ -470,6 +477,8 @@ struct enter_arg
     uint64_t start_time_nsec;
     const void* enter_stack;
     size_t enter_stack_size;
+    const void* augmented_app_config_buf;
+    size_t augmented_app_config_size;
 };
 
 static long _enter(void* arg_)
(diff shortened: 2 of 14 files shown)


tools/myst/common.c
```diff
@@ -186,3 +190,44 @@ bool myst_validate_mount_config(myst_mounts_config_t* mounts)
...
+    char* concat_buf = NULL;
+    // Omit the '\0' at the end of config buffer
+    size_t concat_size =
+        MYST_SHA256_ASCII_SIZE + (config_size ? config_size - 1 : 0);
+    char* root_hash_ascii_buf = NULL;
+
+    if ((root_hash_ascii_buf = concat_buf = malloc(concat_size)) == NULL)
+        ERAISE(-ENOMEM);
+
+    if (config_size)
```
mingweishih, line 212: Also need to ensure `config_buffer` is not null

tools/myst/common.c
```diff
@@ -186,3 +190,44 @@ bool myst_validate_mount_config(myst_mounts_config_t* mounts)
...
+    ECHECK(myst_sha256(&sha256, concat_buf, config_size));
+    memcpy(config_id, sha256.data, sizeof(sha256.data));
+
+done:
+
+    if (concat_buf)
+        free(concat_buf);
+
+    return ret;
+}
```
mingweishih, line 233: new line

tools/myst/common.c
```diff
@@ -186,3 +190,44 @@ bool myst_validate_mount_config(myst_mounts_config_t* mounts)
...
+    const uint8_t* root_hash,
+    size_t root_hash_size,
+    const char* config_buffer,
+    size_t config_size,
+    uint8_t* config_id)
+{
+    int ret = 0;
+    myst_sha256_t sha256;
+    char* concat_buf = NULL;
+    // Omit the '\0' at the end of config buffer
```
mingweishih, line 204: It's unclear how this contract is enforced. If we want to enforce this, I'd suggest adding a check like
```c
if (config_buffer[config_size - 1] != '\0')
  ERAISE(...);
```

Also, need to ensure `config_buffer` is not NULL and the `config_size` greater than 0.

tools/myst/common.c
```diff
@@ -186,3 +190,44 @@ bool myst_validate_mount_config(myst_mounts_config_t* mounts)
...
+    {
+        memcpy(concat_buf, config_buffer, config_size);
+        root_hash_ascii_buf += (config_size - 1);
+    }
+    ECHECK(myst_bin_to_ascii(
+        root_hash,
+        root_hash_size,
+        root_hash_ascii_buf,
+        MYST_SHA256_ASCII_SIZE));
+    printf("config_size=%ld, concat_buf: %s\n", config_size, concat_buf);
```
mingweishih, line 222: Is this for debugging only?

tools/myst/enc/enc.c
```diff
@@ -521,6 +534,24 @@ static long _enter(void* arg_)
     /* Get the enclave size */
     enclave_size = __oe_get_enclave_size();
 
+    if (_get_config_id(config_id) != 0)
```
mingweishih, line 537: The current implementation won't work for coffeelake.

tools/myst/enc/enc.c
```diff
@@ -749,6 +780,8 @@ static long _enter(void* arg_)
         _kargs.start_time_sec = arg->start_time_sec;
         _kargs.start_time_nsec = arg->start_time_nsec;
         _kargs.report_native_tids = report_native_tids;
+        // _kargs.config_id_roothash = config_id;
```
mingweishih, line 783: remove commented code

tools/myst/enc/enc.c
```diff
@@ -1076,6 +1113,68 @@ int __vfprintf_chk(FILE* stream, int flag, const char* format, va_list ap)
...
+    {
+        fprintf(stderr, "oe_parse_report failed.(%s)", oe_result_str(res));
+        return (int)res;
+    }
+
+    sgx_report_body = (sgx_report_body_t*)report.enclave_report;
+    memcpy(out_config_id, sgx_report_body->configid, 64);
+
+    if (evidence_buffer)
+        oe_free_report(evidence_buffer);
```
mingweishih, line 1147: The function won't free this on early return. Could result in memory leak.

tools/myst/host/exec.c
```diff
@@ -129,14 +129,62 @@ long myst_wake_wait_ocall(
...
+    size_t config_size)
+{
+    int ret = 0;
+    oe_sgx_enclave_setting_config_data* config_data_setting_optional = NULL;
+    oe_enclave_setting_t* optional_settings = NULL;
+
+    if ((config_data_setting_optional =
+             malloc(sizeof(oe_sgx_enclave_setting_config_data))) == NULL)
+        ERAISE(-ENOMEM);
+
```
mingweishih, line 147: Set `config_data_setting_optional->config_svn` to 0 if it's not used.
Also, to support both icelake and coffeelake, set `config_data_setting_optional->ignore_if_unsupported` to true.

tools/myst/host/package.c
```diff
@@ -580,6 +582,10 @@ int _exec_package(
     int ret = -1;
     const char** exec_args = NULL;
     myst_args_t mount_mappings = {0};
+    myst_fssig_t fssig;
+    // oe_sgx_enclave_setting_config_data* config_data_setting_optional = NULL;
```
mingweishih, line 586: remove commented code

tools/myst/host/package.c
```diff
@@ -663,6 +669,7 @@ int _exec_package(
                     sizeof(options.rootfs));
                 goto done;
             }
+            // printf("--rootfs=%s\n", options.rootfs);
```
mingweishih, line 672: remove commented code

tools/myst/common.c
Francis-Liu: @mingweishih Thanks for reviewing. This is a temporary solution to append roothash to JSON buffer. In the final version, roothash will be inserted into the config JSON structure.

bodzhang: In the product implementation, the roothash info will be in json name/value pair, as as the app's config.json entries, for example:

"rootfs-measurement-alg": "sha256-flat" 
"rootfs-measurement": “0x1F302E54382B30DFCC21A34B6678F320347890A92D3D45098720C5CCAA303123”

tools/myst/enc/enc.c
```diff
@@ -521,6 +534,24 @@ static long _enter(void* arg_)
...
+        fprintf(stderr, "_get_config_id() failed\n");
+        assert(0);
+    }
+
+    myst_load_fssig(options->rootfs, &fssig);
+    if (_check_config_id(
+            config_id,
+            arg->config_data,
+            arg->config_size,
+            fssig.root_hash,
```
bodzhang, line 548: With the future implementation of rootfs roothash info as json name/value pairs, and as part of the config_data passed in, the check will be to make sure sha256(config_data) = config_id. Then the code should check the iinfo in the rootfs roothash  name/value pairs matches `fssig.root_hash`.

tools/myst/host/package.c
```diff
@@ -580,6 +582,10 @@ int _exec_package(
     int ret = -1;
     const char** exec_args = NULL;
     myst_args_t mount_mappings = {0};
+    myst_fssig_t fssig;
```
bodzhang, line 585: The ext2 rootfs roothash is part of the rootfs image, the function should be able to extract the roothash info from the image instead of requiring the caller to provide it as input.

bodzhang: Thinking over the difference between "packaged" mode and myst-exec mode, I started doubt the relevance of the config_id to the "packaged" mode. In the packaged mode, the app configuration data is part of the enclave image, the CPIO rootfs or the roothash of an EXT2 rootfs is also part of the enclave image, with the enclave built/signed in a trusted build process. The value of config_id is to enable  "loading" different app rootfs/configuration into a prebuilt enclave, which might be provided by and third party. 

With config_id, "myst exec" mode will be able to support "loading" different EXT2 rootfs/configuration into a pre-built enclave. In current implementation, "myst_exec-sgx" command build and sign the enclave, on the running platform (typically, not a trusted build process, therefore "myst exce-sgx" is suitable for dev/debug purpose only). With config_id, an enclave prebuilt in a trusted build process can be used to support `myst exec-sgx [ext2roorfs] [--app-config-path config.json]`

tools/myst/enc/enc.c
```diff
@@ -521,6 +534,24 @@ static long _enter(void* arg_)
...
+            sizeof(fssig.root_hash)) != 0)
+    {
+        fprintf(stderr, "_check_config_id() failed\n");
+        assert(0);
+    }
+
```
bodzhang, line 554: With configuration info passed in, the config_region is the enclave image is not needed anymore. It's OK to leave the enclave build process unchanged for now, for POC purpose, but the data in the config region should be ignored, and instead, the configuration data passed in should be used in the code below, after config_id based verification of the input.

radhikaj:

PR #649, opened 2021-07-28
Add rlimit values to thread_t

Sahakait:
We made the decision not to support setting rlimit values for the moment, as enforcing limits such file size would require significant changes without a current customer use-case to back up. 

Regardless, this PR preserves changes to rlimit setup that will allow support for setting rlimit values if needed in the future.


kernel/syscall.c
```diff
@@ -2933,6 +2933,30 @@ long myst_syscall_prlimit64(
...
+        {
+            old_rlim->rlim_cur =  __myst_kernel_args.max_threads;
+            old_rlim->rlim_max =  __myst_kernel_args.max_threads;
+        }
+    }
+    else if (resource == RLIMIT_AS)
+    {
+        if (old_rlim)
+        {
+            old_rlim->rlim_cur =  -1;
```
vtikoo, line 2948: Is -1 a value with any special meaning here?

Sahakait: Good call out, it means infinity and turns out there is a defined constant (RLIM_INFINITY).

kernel/syscall.c
```diff
@@ -2933,6 +2933,30 @@ long myst_syscall_prlimit64(
...
     }
+    else if (resource == RLIMIT_NPROC)
+    {
+        if (old_rlim)
+        {
+            old_rlim->rlim_cur = __myst_kernel_args.max_threads;
+            old_rlim->rlim_max = __myst_kernel_args.max_threads;
+        }
+    }
+    else if (resource == RLIMIT_AS)
```
vtikoo, line 2944: The cases for RLIMIT_AS and RLIMIT_FSIZE can probably be folded together.

vtikoo:
Could you also fix the reporting for RLIMIT_NOFILE in this PR? We are reporting an incorrect static value of 65536. It should be replaced with [MYST_FDTABLE_SIZE](https://github.com/deislabs/mystikos/blob/main/include/myst/fdtable.h#L22).

kernel/syscall.c
Sahakait: I would prefer to keep them as separate cases as we look into supporting setting RLIMIT.

kernel/syscall.c
```diff
@@ -2933,6 +2933,30 @@ long myst_syscall_prlimit64(
...
+        }
+    }
+    else if (resource == RLIMIT_FSIZE)
+    {
+        if (old_rlim)
+        {
+            old_rlim->rlim_cur = RLIM_INFINITY;
+            old_rlim->rlim_max = RLIM_INFINITY;
+        }
+    }
```
bodzhang, line 2959: RLIMIT_DATA is relevant too. The app won't be able to allocate more than the Mystikos mman region for sure. I don't think it's a must-have for the implementation to enforce the memory allocation limit, but providing a "hint" on how much memory is available might allow some apps or runtimes to adjust their memory usage accordingly. Maybe the implementation can report a percentage of the mman region size, say 70%?

kernel/syscall.c
```diff
@@ -2933,6 +2933,30 @@ long myst_syscall_prlimit64(
...
+        {
+            old_rlim->rlim_cur = __myst_kernel_args.max_threads;
+            old_rlim->rlim_max = __myst_kernel_args.max_threads;
+        }
+    }
+    else if (resource == RLIMIT_AS)
+    {
+        if (old_rlim)
+        {
+            old_rlim->rlim_cur = RLIM_INFINITY;
```
bodzhang, line 2948: App's address space can't exceed the Mystikos mman region. Maybe the Mystikos mman region size should be returned.

kernel/limit.c
```diff
@@ -0,0 +1,151 @@
...
+
+    if (resource < 0 || resource >= RLIM_NLIMITS)
+        ERAISE(-EINVAL);
+
+    if (resource == RLIMIT_MEMLOCK || resource == RLIMIT_SIGPENDING ||
+        resource == RLIMIT_MSGQUEUE)
+        ERAISE(-ENOTSUP);
+
+    if (pid == 0)
+        thread = myst_thread_self();
```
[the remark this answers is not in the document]
vtikoo: `myst_find_process` is now available in `main`.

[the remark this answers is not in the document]
Sahakait: Change has been made.

kernel/limit.c
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// #define _GNU_SOURCE
```
vtikoo, line 4: Pls remove these headers if not needed

kernel/limit.c
```diff
@@ -0,0 +1,148 @@
...
+
+    // RLIMIT_CPU (match linux)
+    rlimits[RLIMIT_CPU].rlim_cur = RLIM_INFINITY;
+    rlimits[RLIMIT_CPU].rlim_max = RLIM_INFINITY;
+
+    // RLIMIT_FSIZE (match linux)
+    rlimits[RLIMIT_FSIZE].rlim_cur = RLIM_INFINITY;
+    rlimits[RLIMIT_FSIZE].rlim_max = RLIM_INFINITY;
+
+    // RLIMIT_DATA (70% of mman region)
```
vtikoo, line 35: Curious if theres any rationale behind this number?

kernel/syscall.c
```diff
@@ -2909,41 +2909,15 @@ long myst_syscall_prlimit64(
...
-
-    if (resource == RLIMIT_NOFILE)
-    {
-        if (old_rlim)
-        {
-            // ATTN: return currently effective RLIMIT_NOFILE settings
-            old_rlim->rlim_cur = 65536;
-            old_rlim->rlim_max = 65536;
-        }
+    long ret = 0;
```
vtikoo, line 2912: this function should probably also be moved to limit.c.

tests/rlimit/rlimit.c
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include <assert.h>
+#include <errno.h>
+#include <stdlib.h>
+
+// #include <myst/thread.h>
```
vtikoo, line 8: remove commented lines

tests/rlimit/rlimit.c
```diff
@@ -0,0 +1,72 @@
...
+    struct rlimit rlim;
+    assert(getrlimit(resource, &rlim) == -1);
+    printf("%s: not supported \n", name);
+}
+
+int main(int argc, char* argv[])
+{
+    assert_rlimit(RLIMIT_CPU, "RLIMIT_CPU", RLIM_INFINITY, RLIM_INFINITY);
+    assert_rlimit(RLIMIT_FSIZE, "RLIMIT_FSIZE", RLIM_INFINITY, RLIM_INFINITY);
+    assert_rlimit_gt(RLIMIT_DATA, "RLIMIT_DATA", 40000000, 40000000);
```
vtikoo, line 58: You could get the total ram via `SYS_sysinfo`.

kernel/limit.c
Sahakait: This was the suggested limit from Bo.

kernel/limit.c
```diff
@@ -0,0 +1,156 @@
...
+        ERAISE(-EFAULT);
+
+    if (pid < 0)
+        ERAISE(-EINVAL);
+
+    if (resource < 0 || resource >= RLIM_NLIMITS)
+        ERAISE(-EINVAL);
+
+    // ATTN: Setting rlimit value is currently unsupported
+    // Returning a failure will break tests/sysinfo and solutions/memcached
```
vtikoo, line 153: Can you allow the silent passthrough only for `RLIMIT_NOFILE`? Thats the case which memcached requires.

Sahakait: Done!

kernel/limit.c
```diff
@@ -0,0 +1,158 @@
...
```
vtikoo, line 153: nit: tests/sysinfo no longer tests setrlimit.

vtikoo (approved):

PR #650, opened 2021-07-29
Tensorflow_Lite sample

bodzhang:


samples/tensorflow_lite/config.json
```diff
@@ -0,0 +1,14 @@
+{
+    "Debug": 1,
+    "StackMemSize": "800k",
+    "NumUserThreads": 8,
```
jxyang, line 4: We don't need NumUserThreads and StackMemSize anymore.

bodzhang: Removed the unused elements.

jxyang (approved):

PR #656, opened 2021-07-30
Various performance enhancements

mikbras:
This PR contains performance enhancement originally prepared for Redis.

Optimizations include:
- limit the buffer size in read/write related ocalls (to avoid extra ocalls to obtain and release host memory).
- inline functions for getting and setting fsbase/fsbase.
- short-circuit syscall for ``SYS_clock_gettime`` (to avoid context switch into kernel).
- use switchless ocalls for read/write related operations.

When building with ``-O3``, I discovered the following problems with ``myst_tcall_get_file_size()`` and ``myst_tcall_read_file()``.
- ``myst_tcall_get_file_size()`` reads the entire file to obtain the size (should only be necessary for virtual files).
- ``myst_tcall_read_file()`` did not correctly determine failure on read.
- there is a window between ``myst_tcall_get_file_size()`` and ``myst_tcall_read_file() where the file size might have changed.

I removed these tcalls and replaced them with the new ``myst_lost_host_file()`` that is implemented within the kernel.


kernel/hostfile.c
```diff
@@ -0,0 +1,121 @@
...
+        ERAISE(-ENOMEM);
+
+    if (myst_buf_reserve(&buf, min_buf_size) != 0)
+        ERAISE(-ENOMEM);
+
+    ECHECK(fd = _host_open(path, O_RDONLY, 0));
+
+    for (;;)
+    {
+        ssize_t n = _host_read(fd, locals->buf, sizeof(locals->buf));
```
jxyang, line 85: Are we sure this implementation is more performant than the previous, i.e., getting the file size first, then reading the whole file in one operation? What about huge files that with this approach would require multiple trips of ocall/tcalls?

tools/myst/host/exec.c
```diff
@@ -146,13 +146,34 @@ int exec_launch_enclave(
...
 
     /* get the start time and pass it into the kernel */
     if (clock_gettime(CLOCK_REALTIME, &start_time) != 0)
         _err("clock_gettime() failed");
 
+    // Initialize the switchless setting; this applies to ocalls with the
+    // transition_using_threads attribute.
+    {
+        switchless_setting.max_enclave_workers = 0;
+        switchless_setting.max_host_workers = 4;
```
jxyang, line 161: Can you use `top` to see if the `max_host_workers` makes a difference on CPU usage? We have to be conscious of not being a `CPU hog`.

kernel/hostfile.c
jxyang: Never mind. I see the answer from the comments below about ocall buffer and switchless.

mikbras: Getting the file size has two problems:
- The file size of a virtual file (/proc) cannot be determined with ``stat``, so the entire file must be enumerated first.
- The size of the file can change after the size has been determined.

tools/myst/host/exec.c
mikbras: Good idea. Will check.

mikbras:
Here is the peak usage for 1, 2, 4, and 8 threads respectively.

```
16535 mike      20   0  544776  19120   6180 R 207.9  0.2   0:10.90 myst [1]
16242 mike      20   0  552972  19304   6356 R 208.6  0.3   0:11.89 myst [2]
16673 mike      20   0  569364  19168   6204 R 207.9  0.2   0:11.22 myst [4]
16386 mike      20   0  602148  19128   6132 R 208.3  0.2   0:11.03 myst [8]
```

tools/myst/host/exec.c
mikbras: Here is the peak usage for 1, 2, 4, and 8 threads respectively.

```
16535 mike      20   0  544776  19120   6180 R 207.9  0.2   0:10.90 myst [1]
16242 mike      20   0  552972  19304   6356 R 208.6  0.3   0:11.89 myst [2]
16673 mike      20   0  569364  19168   6204 R 207.9  0.2   0:11.22 myst [4]
16386 mike      20   0  602148  19128   6132 R 208.3  0.2   0:11.03 myst [8]
```

mikbras: The performance seems more or less the same no matter how many worker threads there are. What do you think is a good number?

jxyang: So the number of worker threads has an impact on memory but not CPU load. Interesting.

jxyang (approved):

PR #662, opened 2021-08-02
redis sample update

salsal97:
added chroot
added kill step for native
added perf tools
make a standard testing process for perf


jxyang:
How long does the test last? Do we want to move to `solutions`?

salsal97:
It takes about 10 seconds, we could add it to solutions if that would enhance the tests. Currently it tests the redis server-client experience by spinning up a server in sgx and client on native

jxyang:
I think it's better to move this to `solutions` and please add a README to describe the nice test harnesses you added.

mikbras (approved):
Look great Saloni. Thank you for your work.

solutions/redis/kill.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+pid=$(ps -eaf | grep redis | grep -v grep | awk '{ print $2 }')
```
vtikoo, line 2:
```suggestion
pid=$(ps -eaf | grep redis-server | grep -v grep | awk '{ print $2 }')
```

salsal97: Added this, thank you!

PR #663, opened 2021-08-02
Set core dump bit in process return code

jxyang:


paulcallen (changes requested):
This is not actually needed. The tests do pass without this change and no core dump reporting is acceptable
Then the tests report TFAIL they dont specify TFATAL (if I remember correctly) which means the test still passes

paulcallen (approved):
my bad, this is correct

PR #664, opened 2021-08-03
Update the mmap error code

### LINKED ISSUE #643: Inaccurate mmap() return error code
[Linux manual ](https://linux.die.net/man/2/mmap#:~:text=mmap%20%28%29%20creates%20a%20new%20mapping%20in%20the,most%20portable%20method%20of%20creating%20a%20new%20mapping.) describes the error code mmap syscall might return. The current Mystikos implementation returns -1 in many cases. The [code](https://github.com/deislabs/mystikos/blob/d0dcb081972ab5ac0ef34acdedaf36f4c3ee749f/kernel/mmanutils.c#L197) has an ATTN comment of "fix the return type". The return value coming form lower layer code also seems to be inaccurate in certain situation.

mingweishih:
Fixes #643

### DIFF
--- a/kernel/mmanutils.c
+++ b/kernel/mmanutils.c
@@ -127,14 +127,11 @@ static ssize_t _map_file_onto_memory(
     };
     struct locals* locals = NULL;
 
-    if (fd < 0 || !addr || !length)
+    if (fd < 0 || !addr || !length || offset % PAGE_SIZE)
         ERAISE(-EINVAL);
 
     if (!(locals = malloc(sizeof(struct locals))))
         ERAISE(-ENOMEM);
-
-    // ATTN: generate EACCES error if non-regular file or file not opened
-    // for write when mmap_flags has MMAP_WRITE.
 
     /* read file onto memory */
     {
@@ -194,8 +191,7 @@ done:
     return ret;
 }
 
-/* ATTN-A: fix return types for this function */
-void* myst_mmap(
+long myst_mmap(
     void* addr,
     size_t length,
     int prot,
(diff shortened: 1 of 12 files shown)


kernel/mmanutils.c
```diff
@@ -194,106 +194,127 @@ static ssize_t _map_file_onto_memory(
...
     }
 
     if (fd >= 0 && addr)
     {
         ssize_t n;
+
         // ATTN: call mmap or mremap here so that this range refers to
         // a mapped region.
-        // ATTN: is non-page-aligned length valid?
+        // ATTN: Use EACCES for failures related to fd access. This may not
```
bodzhang, line 262: LTP mmap15.c test tries to mmap to an address typically outside of app's linear address space -  `SYS_mmap(addr=fffffffffffff000 length=4096(1000) prot=1 flags=17 fd=3 offset=0)`, and expects EINVAL or ENOMEM error. The implementation here probably should check the whether the request address range is out of bound, before attempting to read file data (with mprotect()  call to make sure the memory range is writable).

kernel/mmanutils.c
```diff
@@ -194,106 +194,127 @@ static ssize_t _map_file_onto_memory(
...
+            // conform the Linux kernel behavior.
+            if (!(prot & MYST_PROT_WRITE))
+            {
+                if (myst_mman_mprotect(
+                        &_mman, (void*)ret, length, prot | MYST_PROT_WRITE))
+                    ERAISE(-EACCES);
+            }
+            if ((n = _map_file_onto_memory(
+                     fd, offset, (void*)ret, length, flags)) < 0)
+                ERAISE((unsigned long)-n);
```
bodzhang, line 302: `n` is negative, shouldn't it be `ERAISE(n)`?

kernel/mmanutils.c
mingweishih: `myst_mman_mprotect` have done such checks and return EINVAL when the checks failed. I updated the PR to cover the case when `myst_mman_mprotect` is not called.

kernel/mmanutils.c
mingweishih: updated.

bodzhang (approved):

PR #676, opened 2021-08-04
Enable zero-base enclave feature

mikbras:
This PR enables the **zero-base enclave feature** in Mystikos. The relevant addresses are as follows.

- The enclave base address is zero.
- The enclave start address is **0x800000** (8 megabytes).
- The host start address is **0x10000000000** (1 terabyte).

The following test obtains the enclave start address (via an extended syscall) and verifies the address.

```
./tests/oe/zerobase
```

Zero-base enclaves are enabled by default. They can be disabled by commenting out the following line in  ``config.mak``.

```
MYST_ENABLE_ZERO_BASE_ENCLAVES=1
```

Our ELF parser could not handle loading ``myst`` (to extract the sections) because now all the virtual addresses are relative to 1TB. The ELF parser still works fine for ``libmystcrt.so`` and ``libmystkernel.so``. To workaround, I wrote a single function called ``load_sections()`` that loads  only selected sections from the ``myst`` program.


kernel/syscall.c
```diff
@@ -6304,6 +6304,8 @@ static long _syscall(void* args_)
         case SYS_myst_oe_free_attestation_certificate:
         case SYS_myst_oe_verify_attestation_certificate:
         case SYS_myst_oe_result_str:
+        case SYS_myst_oe_get_enclave_start_address:
```
mingweishih, line 6307: Do we need to expose these two information (we didn't expose the image_base before)?

mingweishih (commented):
LGTM. Just one comment regarding whether we need to expose the base/start addresses.

kernel/syscall.c
mikbras: This is an internal undocumented syscall that exposes it only for the sake of testing.

mingweishih: Another option is getting the value via __myst_kernel_args. We already set the start address now (image_data), see https://github.com/deislabs/mystikos/blob/ab55ee086dd2b687488dc34637602c8d6325c39e/tools/myst/enc/enc.c#L515 . We could add additional member to store the base address.

tools/myst/host/package.c
```diff
@@ -907,8 +886,9 @@ int _exec_package(
     if (unpack_dir)
         remove_recursive(unpack_dir);
 
-    if (elf_loaded)
-        elf_image_free(&myst_elf);
+#if 0
```
mingweishih, line 889: Why this is disabled?

tests/oe/zerobase/zerobase.c
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the MIT License.
+
+#include <assert.h>
+#include <openenclave/enclave.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+/* non-public OE functions */
+extern const void* __oe_get_enclave_start_address(void);
```
mingweishih, line 11: Question: in this PR, does this function calls the OE implementation directly or using the mystikos syscall path implemented above?

kernel/syscall.c
mikbras: I think it is okay. It provides us what we need. :)

mikbras: Yes, I like the idea. Let me see if it works out.

mikbras: I looked into this and we also have to support the Linux target and other TEEs in the future. The base-address/start-address is an SGX-only concept and exposing these to the kernel would break our separation of TEE from kernel. We added the extended syscalls only for the SGX TEE. These calls are not available in the Linux target at all.

tools/myst/host/package.c
mikbras: Good catch. I put it  back.

tests/oe/zerobase/zerobase.c
mikbras: This test uses out OE-compatibility library (``libopenenclave.a``), located under ``./oe``.

So you can call OE functions from a mystikos app that links with ``libopenenclave.a``. The functions are actually reimplemented to make extended syscalls.

mikbras:
```C++
const void* __oe_get_enclave_start_address(void)
{
    uint64_t args[] = {0};
    return (const char*)syscall(SYS_myst_oe_get_enclave_start_address, args);
}
```

mikbras:
```C++
oe_result_t oe_get_seal_key_by_policy_v2(
    oe_seal_policy_t seal_policy,
    uint8_t** key_buffer,
    size_t* key_buffer_size,
    uint8_t** key_info,
    size_t* key_info_size)
{
    uint64_t args[] = {
        (uint64_t)seal_policy,
        (uint64_t)key_buffer,
        (uint64_t)key_buffer_size,
        (uint64_t)key_info,
        (uint64_t)key_info_size,
    };
    return (oe_result_t)syscall(SYS_myst_oe_get_seal_key_by_policy_v2, args);
}
```

mikbras: So you can include the standard OE header (``<openenclave/enclave.h>``) and call OE functions as though you were in an enclave.

mingweishih (approved):

PR #680, opened 2021-08-05
Enable 0-base using config

RRathna:
Enables 0-base oesign configs.

### DIFF
--- a/tools/myst/config.c
+++ b/tools/myst/config.c
@@ -69,6 +69,32 @@ done:
     return ret;
 }
 
+static json_result_t _extract_start_address(
+    json_type_t type,
+    const json_union_t* un,
+    uint64_t* start_address)
+{
+    json_result_t ret = JSON_FAILED;
+    uint64_t value;
+
+    if (start_address == NULL)
+        CONFIG_RAISE(JSON_FAILED);
+
+    if (type == JSON_TYPE_INTEGER && un->integer > 0)
+        value = (uint64_t)un->integer; // number in bytes
+    else
+        CONFIG_RAISE(JSON_TYPE_MISMATCH);
+
+    if (value % PAGE_SIZE) // check if aligned to PAGE_SIZE
+        CONFIG_RAISE(JSON_REASON_VALUE);
+
+    *start_address = value;
+
+    ret = JSON_OK;
+done:
+    return ret;
+}
+
 static json_result_t _json_read_callback(
     json_parser_t* parser,
     json_reason_t reason,
@@ -148,6 +174,26 @@ static json_result_t _json_read_callback(
                 }
                 else
                     CONFIG_RAISE(JSON_TYPE_MISMATCH);
+            }
+            else if (json_match(parser, "CreateZeroBaseEnclave") == JSON_OK)
+            {
+                if ((type == JSON_TYPE_BOOLEAN) && un->boolean)
+                {
+                    parsed_data->oe_create_zero_base = true;
+                }
+                else if (type == JSON_TYPE_BOOLEAN)
+                {
+                    parsed_data->oe_create_zero_base = false;
+                }
+                else
+                    CONFIG_RAISE(JSON_TYPE_MISMATCH);
+            }
+            else if (json_match(parser, "EnclaveStartAddress") == JSON_OK)
+            {
+                ret = _extract_start_address(
+                    type, un, &parsed_data->oe_start_address);
+                if (ret != JSON_OK)
+                    CONFIG_RAISE(ret);
             }
 
             // Mystikos configuration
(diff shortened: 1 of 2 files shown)


mikbras (commented):
Thank you Rathna. This looks great. A few minor comments.

And yes, you are right. We need some kind of test, but I would wait for my zero-based integration PR, because we can then check the values in the app.

Please see this PR:
https://github.com/deislabs/mystikos/pull/676

tools/myst/config.c
```diff
@@ -476,6 +528,15 @@ int write_oe_config_fd(int fd, config_parsed_data_t* parsed_data)
 
     fprintf(out_file, "SecurityVersion=%d\n", parsed_data->oe_security_version);
 
+    fprintf(
+        out_file,
+        "CreateZeroBaseEnclave=%d\n",
+        parsed_data->oe_create_zero_base);
+    if (parsed_data->oe_create_zero_base)
+    {
+        fprintf(out_file, "StartAddress=%d\n", parsed_data->oe_start_address);
```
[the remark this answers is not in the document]
mingweishih: The default interaction between OE and Intel SGX libraries is OE passes addr as NULL, and allows the latter the choose a start address. OE currently does not allow developer to specify the start address. Do you think it's worth supporting this option?

tools/myst/config.c
```diff
@@ -69,6 +69,32 @@ static json_result_t _extract_mem_size(
...
+    if (start_address == NULL)
+        CONFIG_RAISE(JSON_FAILED);
+
+    if (type == JSON_TYPE_INTEGER && un->integer > 0)
+        value = (uint64_t)un->integer; // number in bytes
+    else
+        CONFIG_RAISE(JSON_TYPE_MISMATCH);
+
+    if (value && value % OE_PAGE_SIZE != 0) // check if aligned to OE_PAGE_SIZE
+        ret = JSON_REASON_VALUE;
```
mingweishih, line 89: Do CONFIG_RAISE

tools/myst/config.c
```diff
@@ -69,6 +69,32 @@ static json_result_t _extract_mem_size(
...
+    const json_union_t* un,
+    uint64_t* start_address)
+{
+    json_result_t ret = JSON_FAILED;
+    uint64_t value;
+
+    if (start_address == NULL)
+        CONFIG_RAISE(JSON_FAILED);
+
+    if (type == JSON_TYPE_INTEGER && un->integer > 0)
```
mingweishih, line 83: The greater than zero check is redundant if check in line 84 is done correctly.

tools/myst/config.c
```diff
@@ -69,6 +69,32 @@ static json_result_t _extract_mem_size(
...
+        value = (uint64_t)un->integer; // number in bytes
+    else
+        CONFIG_RAISE(JSON_TYPE_MISMATCH);
+
+    if (value && value % OE_PAGE_SIZE != 0) // check if aligned to OE_PAGE_SIZE
```
mingweishih, line 88: Shouldn't this be if (!value || value % OE_PAGE_SIZE)?

tools/myst/config.c
```diff
@@ -69,6 +69,32 @@ static json_result_t _extract_mem_size(
...
+        ret = JSON_REASON_VALUE;
+    else
```
mingweishih, line 90: Nit: else branch is not needed. Can just set start_address after the check.

tools/myst/config.h
```diff
@@ -6,6 +6,7 @@
 
 #include <myst/json.h>
 #include <myst/kernel.h>
+#include <openenclave/bits/defs.h>
```
mingweishih, line 9: Is this needed?

RRathna: yes, for OE_PAGE_SIZE used in the start_address value check.

mingweishih: I think mystikos has its own PAGE_SIZE definition. Can you check?

tools/myst/config.c
RRathna: Negative numbers should not be allowed as input and should be checked before the conversion to unsigned in line 84. At line 88, I am checking only for page alignment.

tools/myst/config.h
RRathna: I see PAGE_SIZE = OE_PAGE_SIZE = 0x1000. Changed code to use PAGE_SIZE instead.

mikbras (approved):
Looks good Rathna. Thank you so much for doing this.

mingweishih (approved):

mingweishih (commented):
Give Mike's 0-base PR has been merged already, can we also have tests for this PR?

RRathna:
discarding this PR since pipeline is not getting triggered. New PR - https://github.com/deislabs/mystikos/pull/746

PR #681, opened 2021-08-05
Pre-allocate thread entry stack upon thread creation

mingweishih:
This PR moves the allocation of temporary thread stack from upon the thread entry to thread creation. This avoids the stack overflow when enabling the switchless for mprotect ocall. Note that the run-time performance improvement on the switchless mprotect is about 5x.

Here is more analysis:
Upon the thread entry (`myst_run_thread`), the kernel will allocate a temporary stack (we called the entry stack) and switch to that stack from TCS stack (hard-coded as 2-page size). The stack allocation internally invokes the `mprotect` (by the internal mmap implementation). When making the `mprotect` switchless, there could be two cases:
1. Host is spinning and will perform the `mprotect` right away then return.
2. Host is sleeping. In this case, the enclave needs to make another ocall to wake up the host.
    This additional OCALL causes the stack usage over the size limit (2 pages).

The work-around is pre-allocating the entry stack upon the stack creation. However, the drawback on this approach is that the myst kernel needs to keep this stack throughout the thread lifetime.


bodzhang (approved):

paulcallen (approved):

tools/myst/myst.edl
```diff
@@ -132,7 +134,8 @@ enclave
 
         long myst_nanosleep_ocall(
             [in] const struct timespec* req,
-            [out] struct timespec* rem);
+            [out] struct timespec* rem)
```
anakrish, line 137: Does this cause performance improvement? The thread that makes myst_nanosleep_ocall would keep spinning while the host worker thread would actually do the nanosleep.

tools/myst/myst.edl
```diff
@@ -120,7 +121,8 @@ enclave
             uint64_t self_event,
             [in] const struct myst_timespec* timeout);
 
-        long myst_sched_yield_ocall();
+        long myst_sched_yield_ocall()
```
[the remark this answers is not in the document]
mingweishih: Reverted the change

tools/myst/myst.edl
mingweishih: Reverted the change given there is no much benefit

anakrish: If, on the other hand, you do see performance benefit with this change, then it would imply that for really small sleep amounts, it is beneficial to just spin for an approximate equivalent time instead of making an ocall.

anakrish (approved):

PR #682, opened 2021-08-06
Change .NET coreclr test suite configuration

bodzhang:
- Divide the total of 2496 test cases to multiple groups, and run each sub-group with different configuration of memory size and timeout threshold;
- Use "packaged mode" for the sub-group with the largest number of test cases to avoid the extra Enclave measurement calculation associated with "myst-exec mode" on each test case;
- Disable parallelization of test execution. On pre-icelake SGX platforms, EPC size is very small, parallel execution of test cases triggers EPC paging and quite often cause more test cases to timeout.


bodzhang:
dotnet coreclr pipeline run result is [here](https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10223&view=results)

bodzhang:
dotnet coreclr pipeline run on the lastest version of the PR is [here](https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10358&view=logs&j=ab477ad6-9781-5b54-82ca-a0141b26726e)

vtikoo (approved):
Thanks for introducing test classification and improving the pass percentage!
For the general case(pr0-256m), did you see any performance difference in running the tests on 2 cores vs 4?

bodzhang:
> Thanks for introducing test classification and improving the pass percentage!
> For the general case(pr0-256m), did you see any performance difference in running the tests on 2 cores vs 4?

When setting parallel execution to 4, I still see some random timeout (much less than the configuration before this PR though) even with 30s timeout threshold.

PR #684, opened 2021-08-07
Fix for process-mappings and file-mappings

mikbras:
This PR introduces the **fdmappings vector**, which keeps track of mappings between file descriptors and memory pages. It adds a new region called **fdmappings** which contains an entry for every page in the memory-management region. An **fdmapping** is defined as follows.

```C++
typedef struct myst_fdmapping
{
    uint32_t used;           /* whether entry is used */
    int32_t fd;              /* fd that page is mapped to */
    uint64_t offset;         /* offset of page within file */
    myst_refstr_t* pathname; /* full pathname associated with fd */
} myst_fdmapping_t;
```

This PR does the following:
- [x] reworks the ``msync`` implementation.
- [x] reworks the ``/proc/self/maps`` virtual mapping.
- [x] reworks the process-mappings.
- [x] adds ``--nobrk`` option so libcxx tests can be run in one pass with only 32m of RAM.
- [x] fixes memory leaks associated with the execve system call.
- [x] made main thread static (to eliminate memory leak).
- [x] incremented reference-counted string type (used in fd-mappings).  
- [x] implemented recursive spinlocks (for locking mman-related structures). 
- [x] added ``vaprintf`` and ``asprintf`` (needed by ``/proc/self/maps`` implementation).


kernel/mmanutils.c
```diff
@@ -92,35 +109,145 @@ int myst_teardown_mman(void)
...
+    {
+        const size_t count = length / PAGE_SIZE;
+        uint64_t off = offset;
+
+        for (size_t i = index; i < index + count; i++)
+        {
+            myst_fdmapping_t* p = &v.fdmappings[i];
+
+            /* ATTN: the entry might already be in use for unknown reasons */
+            if (p->used == MYST_FDMAPPING_USED)
```
paulcallen, line 218: how? Someone is trying to map to an address someone else is using? Is this code hit?

kernel/mmanutils.c
```diff
@@ -350,114 +454,36 @@ int myst_mprotect(const void* addr, const size_t len, const int prot)
...
+    {
+        const size_t count = length / PAGE_SIZE;
 
-            p = next;
+        for (size_t i = index; i < index + count; i++)
+        {
+            myst_fdmapping_t* p = &v.fdmappings[index];
+
+            /* remove this fd-mapping */
+            p->used = 0;
```
paulcallen, line 479: assert if it was not mapped?

kernel/mmanutils.c
```diff
@@ -517,168 +528,141 @@ int myst_get_free_ram(size_t* size)
...
-    myst_spin_unlock(&_mappings_lock);
+        assert(index < v.pids_count);
+        assert(index + count <= v.pids_count);
 
-done:
+        myst_rspin_lock(&_mman.lock);
+        {
+            for (size_t i = index; i < index + count;)
+            {
+                if (v.pids[i] == (uint32_t)pid)
```
paulcallen, line 558: huh. Technically we could refcount the mappings if more than one process has it and the last one wins.
this way we should never get into this case, and if it does it means a real bug

kernel/mmanutils.c
```diff
@@ -517,168 +528,141 @@ int myst_get_free_ram(size_t* size)
...
 {
     int ret = 0;
-    myst_process_mapping_t* m = NULL;
-    struct locals
-    {
-        char realpath[PATH_MAX];
-    }* locals = NULL;
 
-    if (pid < 0 || !addr || (addr == (void*)-1) || !size)
+    if (pid <= 0)
```
paulcallen, line 536: assert

kernel/mmanutils.c
paulcallen, line 558: I think this is too vague for a check.
Lets say a while file is mapped to a file. Then if part of the mapping is unmapped (code in the else) then what happens if the same process gets this newly unmapped page? It will release it incorrectly.

paulcallen (commented):
still reviewing mmanutils.c, but publishing comments so far... 
So far I believe these is only one actual issue I have found, and one recommendation based on previous discussions. Both are on line 558 of mmanutils.c

paulcallen (approved):

tests/procfs/procfs.c
```diff
@@ -93,14 +94,37 @@ int test_readonly()
...
+        const size_t length = statbuf.st_size / 2;
+        if (mmap(addr, length, prot, flags, fd, 8192) == MAP_FAILED)
+            assert(0);
+    }
+
+    {
+        int fd = open("/proc/self/maps", O_RDONLY);
+        assert(fd > 0);
+        assert(read(fd, buf, sizeof(buf)));
+        printf("%s\n", buf);
```
vtikoo, line 125: Can we assert that datafile was found in the mappings.

mikbras: Good idea. I will try.

kernel/mmanutils.c
```diff
@@ -849,33 +943,40 @@ void myst_mman_close_notify(int fd)
...
+            while (p < end)
             {
-                msync_mapping_t* next = p->next;
+                // Efficiently skip over zero-valued bytes. In-use fdmappings
+                // begin with a non-zero byte.
+                if ((p = myst_memcchr(p, '\0', bytes_remaining)) == NULL)
+                    break;
 
-                if (p->fd == fd)
+                if (p->used == MYST_FDMAPPING_USED && p->fd == fd)
```
vtikoo, line 972: Clearing the fd mapping might conflict with writing back modifications at `munmap()` time for MAP_SHARED.

kernel/mmanutils.c
```diff
@@ -780,59 +902,31 @@ int myst_msync(void* addr, size_t length, int flags)
...
     }
-    myst_spin_unlock(&_msync_mappings_lock);
+    myst_rspin_unlock(&_mman.lock);
 
 done:
     return ret;
 }
 
-/* notified on close to remove msync mappings involving fd */
+/* notified on close to clear msync mappings involving fd */
```
vtikoo, line 932:
```suggestion
/* notified on close to clear fdmappings involving fd */
```

kernel/mmanutils.c
```diff
@@ -517,219 +571,288 @@ int myst_get_free_ram(size_t* size)
...
+                    /* count consecutive pages with same traits */
+                    for (size_t j = i + 1; j < index + count; j++)
+                    {
+                        int tmp_prot = 0;
+
+                        /* if the pid changes */
+                        if (v.pids[j] != (uint32_t)pid)
+                            break;
+
+                        if (v.fdmappings[i].used != used)
```
vtikoo, line 800: Should the index variable be `j`, as the others?

kernel/mmanutils.c
```diff
@@ -517,219 +571,288 @@ int myst_get_free_ram(size_t* size)
...
+                            (pathname ? pathname->data : ""),
+                            &str) == 0)
+                    {
+                        ECHECK(myst_buf_insert(vbuf, 0, str, strlen(str)));
+                        free(str);
+                    }
+
+                    i += n;
+                    addr += len;
+                    length -= len;
```
vtikoo, line 844: `length` is unused in the critical section.

kernel/mmanutils.c
vtikoo: This also affects reporting of pathname information in /proc/[pid]/maps.

include/myst/rspinlock.h
```diff
@@ -0,0 +1,63 @@
...
+
+    /* wait on the lock */
+    myst_spin_lock(&s->lock);
+    s->owner = __myst_rspin_self();
+    s->count = 1;
+}
+
+MYST_INLINE void myst_rspin_unlock(myst_rspinlock_t* s)
+{
+    if (--s->count == 0)
```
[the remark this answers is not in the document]
mikbras: Good idea. I added these asserts.

```
    assert(s->count > 0);
    assert(s->owner == __myst_rspin_self());
```

kernel/refstr.c
```diff
@@ -0,0 +1,43 @@
...
+    {
+        refstr->count++;
+    }
+}
+
+void myst_refstr_unref(myst_refstr_t* refstr)
+{
+    if (refstr)
+    {
+        if (--refstr->count == 0)
```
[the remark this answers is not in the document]
mikbras: Fixed.

kernel/stack.c
```diff
@@ -0,0 +1,82 @@
...
+static size_t _nstacks;
+static myst_spinlock_t _lock = MYST_SPINLOCK_INITIALIZER;
+
+int myst_register_stack(const void* stack, size_t size)
+{
+    int ret = -EINVAL;
+
+    myst_spin_lock(&_lock);
+    {
+        if (_nstacks < MAX_STACKS)
```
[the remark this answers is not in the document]
mikbras: There can never be more than MAX_STACKS (else it is an error)

Added assert as suggested.

kernel/mmanutils.c
mikbras: It is legal for it not to be mapped. This is called by ``munmap()`` and it removes any file mappings if there are any (and there usually will not be).

kernel/mmanutils.c
mikbras: I didn't think of that until you mentioned it to me. I created this issues to track:

https://github.com/deislabs/mystikos/issues/690

kernel/mmanutils.c
mikbras: Unresolved! I need to think more about how this could happen.

kernel/stack.c
```diff
@@ -0,0 +1,82 @@
...
+
+int myst_unregister_stack(const void* stack, size_t size)
+{
+    int ret = -EINVAL;
+
+    myst_spin_lock(&_lock);
+    {
+        for (size_t i = 0; i < _nstacks; i++)
+        {
+            if (_stacks[i].stack == stack && _stacks[i].size == size)
```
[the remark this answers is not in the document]
mikbras: I cannot unless the wrong parameter is passed. It is mainly a sanity check.

kernel/mmanutils.c
mikbras: Done. Thanks.

tests/procfs/procfs.c
mikbras: I added a more elaborate test to check for existence of ``/datafile`` and also to do a partial unmap and check that there are now two entries.

kernel/mmanutils.c
mikbras: ``mmap`` does not currently support ``MAP_SHARED``. This is something to keep in mind when we implemented shared file mappings.

mikbras: Should the mapping be preserved after the fd has already been closed? In any case, it is not something we supported before so we can create an issue if necessary.

kernel/mmanutils.c
mikbras: This diff had a conflicts but I manually applied it. Thanks!

kernel/mmanutils.c
mikbras: Oh my! Great catch! Fixed and running pipeline again.

kernel/mmanutils.c
mikbras: Good find. I removed it.

kernel/mmanutils.c
mikbras: I added a comment to the code to explain.

mikbras:
Thank you Paul and Vikas for your review. I addressed all comments.

kernel/mmanutils.c
vtikoo: I think we supported it because /maps relied on process mappings which were released on process shutdown. But agree we can track it in an issue, should not be PR blocking.

vtikoo (approved):

paulcallen (approved):
looking good

PR #687, opened 2021-08-09
Add SYS_pause support

Francis-Liu:


kernel/syscall.c
```diff
@@ -6605,3 +6610,30 @@ long myst_syscall_unload_symbols(void)
     long params[6] = {0};
     return myst_tcall(MYST_TCALL_UNLOAD_SYMBOLS, params);
 }
+
+long myst_syscall_pause(void)
+{
+    long ret = 0;
+    myst_thread_t* thread = myst_thread_self();
+    // Set futex unavilable. This is to make sure the futex which was made
+    // available by signals delivered before calling pause won't affect the
```
vtikoo, line 6619: Does a futex wake set the futex value to 1?

vtikoo: Oh it looks like `__sync_val_compare_and_swap(pause_futex, 0, 1)` in `myst_signal_deliver` does that.

Francis-Liu: Yes @vtikoo that is correct.

kernel/signal.c
```diff
@@ -451,6 +451,14 @@ long myst_signal_deliver(
     /* Make sure any polls get woken up to process any outstanding events */
     myst_tcall_poll_wake();
 
+    if (!__sync_val_compare_and_swap(&thread->pause_futex, 0, 1))
+    {
+        ret = myst_futex_wake(&thread->pause_futex, 1);
```
vtikoo, line 456: I'm confused about whether the futex word or its value gets dereferenced from the host in [target/shared/waitwake.c](https://github.com/deislabs/mystikos/blob/main/target/shared/waitwake.c#L18). Just want to make sure an address within the enclave is not being accessed from the host?

vtikoo: The `thread->event` futex stores the value of/is backed by a host side thread local variable - https://github.com/deislabs/mystikos/blob/main/tools/myst/host/exec.c#L73.

Francis-Liu: Great finding. Do you know why does the `thread->event` futex point to a host address?

vtikoo: I guess because for sgx target, the `SYS_futex` calls in the target needs a futex address in the target/host.

Francis-Liu: Well, if I understand correctly, futex like `thread->pause_futex` and `thread->fork_exec_futex_wait` reference address within enclave. These futex are safe meaning that their addresses are not accessible from the host. But under the hood, they depend on `thread->event` which resides on the host for `wake` TCALL. Please correct me if I'm wrong.

vtikoo: Oh didn't realize this is another futex usage like this. For both of these usages, `myst_tcall_wait/wake` are called directly with the address of `thread->pause_futex` or `fork_exec_futex_wait`. `thread->event` I think is used primarily for the condition variables and mutex implementation, and is not being used here.
cc: @paulcallen

vtikoo: You're right. I was getting confused between `myst_futex_wait` and `myst_tcall_wait`.

vtikoo (approved):
Looks good!

PR #691, opened 2021-08-10
[Blocked] Add option to use onnxruntime to pytorch sample

Sahakait:
Currently blocked on sys_futex() flags FUTEX_WAIT_BITSET and FUTEX_CLOCK_REALTIME


radhikaj (commented):
Couple of highlevel comments - should the dir name be changed to reflect that it runs ONNX as well?
One thing that would be helpful is if the makefile can be changedto exercise the ONNX option and the pytorch option.
A readme would be great(need not be tied to this PR since this sample existed before this PR),

radhikaj:
@bodzhang @jxyang , can you please share thoughts on adding ONNX to this sample vs. having a different sample.

PR #692, opened 2021-08-10
WIP - Decouple makefiles and create submodules

rs--:
### Decoupling Makefiles

As a precursor to enabling the signing pipeline, the targets/functionality of the makefiles have to be decoupled. Due to the restrictions of the signing pipeline, the build phase is air-gapped.

New makefile targets (init, build, release-build) are added, along with their descriptions.

`make all` or simply `make` should run as normal; subsequent dependencies are automatically pulled in, and all targets are built.

`make init` initializes the project dependencies.

`make build` skips the initialization steps (dependencies, and prereqs), and builds all targets.

`make release-build` runs `make build` with the MYST_PRODUCT_BUILD flag set, and runs `make bindist`.

### Submodules

Since there was no consistency with how dependencies were being managed (submodules, or manually cloning with/without patching), I figured that having a consistent dependency management workflow would be beneficial.

Instead of cloning dependency repositories and applying patches, I created forks of said repos, and applied the respective patches in branches.  This (along with the decoupling) is up for debate, I just figured that using submodules would allow for more flexibility when testing different dependency configurations; instead of having different patch versions, only references to the submodules need to change.

The submodules currently in use are placeholder submodules, and pending approval, forks of the dependencies should be managed by an Github organization.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,31 @@
+[submodule "third_party/openenclave/openenclave"]
+	path = third_party/openenclave/openenclave
+	url = https://github.com/openenclave/openenclave.git
+	branch = mystikos.v5
+[submodule "third_party/musl/musl"]
+	path = third_party/musl/musl
+	url = https://github.com/rs--/openenclave-musl.git
+	branch = mystikos-1.2.0
+[submodule "third_party/gcompat/gcompat"]
+	path = third_party/gcompat/gcompat
+	url = https://github.com/rs--/gcompat.git
+	branch = mystikos-1.0.0
+[submodule "third_party/ltp/ltp"]
+	path = third_party/ltp/ltp
+	url = https://github.com/lsds/ltp.git
+	branch = sgx-lkl
+[submodule "third_party/mbedtls/mbedtls"]
+	path = third_party/mbedtls/mbedtls
+	url = https://github.com/ARMmbed/mbedtls
+	branch = mbedtls-2.16
+[submodule "third_party/musl/crt/musl"]
+	path = third_party/musl/crt/musl
+	url = https://github.com/rs--/openenclave-musl.git
+	branch = mystikos-1.2.0
+[submodule "third_party/musl/pristine/musl"]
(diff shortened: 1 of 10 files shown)


mikbras:
Than you Rob for your interest. Can you say more about what you mean by this.

> As a precursor to enabling the signing pipeline, the targets/functionality of the 
> makefiles have to be decoupled. Due to the restrictions of the signing pipeline, the 
> build phase is air-gapped.

What does it mean to sign the pipeline and how do submodules enable that?

jxyang:
Being worked on submodules and the current approach, i.e., cloning projects and applying patches, I have to say submodules are much harder to work with. The submodule commands of git are apparently an after-thought addition, and require memorizing a command set that is dramatically different from other git commands.

It's hard to say when we can get a github organization, probably never as the company open source guideline advises against issuing new github organizations.

rs--:
@mikbras Submodules don't enable the use of the CDPx signing pipeline, but decoupling of the makefiles do. Because the build phase of the signing pipeline is air-gapped the current workflow of cloning dependencies during the build process. Signing of packages/executables is needed if there are plans on releasing Mystikos through Microsoft channels (namely the APT repo).

@jxyang Totally understandable, I just wanted to see what folks thought of submodules - I can revert the submodule changes. We don't necessarily need a new organization, or specifically an organization directly related to Microsoft. Would it be possible to create forks of the dependencies on the deislabs org?

If it is possible to create said forks, are you open to the idea of using branches in those forks instead of apply patches? If it's not possible to create the forks, then disregard this question.

radhikaj:
@jxyang, @mikbras  For CDPX - during build you dont have internet connection. Having the code checked out and available before the build phase is the right way to do it.
git submodules are used in many projects and saying that they are harder than makefiles with patches is a matter of opinion. In OE, we have been using submodules since 1.5 years and have had no problems with them. 
We have to work with the constraints of CDPX to get this working. submodules are a better and cleaner way of doing this.

PR #696, opened 2021-08-11
Add --main-stack-size <size> option to myst-exec

Francis-Liu:

### DIFF
--- a/tools/myst/enc/enc.c
+++ b/tools/myst/enc/enc.c
@@ -20,6 +20,7 @@
 #include <myst/kstack.h>
 #include <myst/mmanutils.h>
 #include <myst/mount.h>
+#include <myst/process.h>
 #include <myst/ramfs.h>
 #include <myst/regions.h>
 #include <myst/reloc.h>
@@ -492,6 +493,7 @@ static long _enter(void* arg_)
     bool perf = false;
     bool report_native_tids = false;
     size_t max_affinity_cpus = options ? options->max_affinity_cpus : 0;
+    size_t main_stack_size = options ? options->main_stack_size : 0;
     const char* rootfs = NULL;
     config_parsed_data_t parsed_config;
     bool have_config = false;
@@ -626,6 +628,12 @@ static long _enter(void* arg_)
     if (have_config && parsed_config.max_affinity_cpus)
     {
         max_affinity_cpus = parsed_config.max_affinity_cpus;
+    }
+
+    // Override main stack size if present in config
+    if (have_config && parsed_config.main_stack_size)
+    {
+        main_stack_size = parsed_config.main_stack_size;
     }
 
     // record the configuration for which fork mode
@@ -766,6 +774,8 @@ static long _enter(void* arg_)
         _kargs.report_native_tids = report_native_tids;
         _kargs.enter_stack = arg->enter_stack;
         _kargs.enter_stack_size = arg->enter_stack_size;
+        _kargs.main_stack_size =
+            main_stack_size ? main_stack_size : MYST_PROCESS_INIT_STACK_SIZE;
 
         /* whether user-space FSGSBASE instructions are supported */
         _kargs.have_fsgsbase_instructions = options->have_fsgsbase_instructions;
--- a/tools/myst/host/exec_linux.c
+++ b/tools/myst/host/exec_linux.c
@@ -21,6 +21,7 @@
 #include <myst/file.h>
 #include <myst/hex.h>
 #include <myst/kernel.h>
+#include <myst/process.h>
 #include <myst/regions.h>
 #include <myst/reloc.h>
 #include <myst/round.h>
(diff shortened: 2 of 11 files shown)


tools/myst/host/package.c
```diff
@@ -647,6 +649,28 @@ int _exec_package(
         }
     }
 
+    /* Get --main-stack-size */
+    {
+        const char* opt = "--main-stack-size";
+        const char* arg = NULL;
+
+        if ((cli_getopt(&argc, argv, opt, &arg) == 0))
```
bodzhang, line 657: If the config file does not include main-stack-size entry or has the entry with a specific value, in packaged mode, is commandline option of main-stack-size ignored?

Francis-Liu: The following is true in both package and exec mode: If the config file does not include MainStackSize, the commandline option --main-stack-size will be effective. If config file includes MainStackSize, the commandline option --main-stack-size will be ignored.

bodzhang: "If the config file does not include MainStackSize, the commandline option --main-stack-size will be effective" -- this is not right. In packaged mode, the commandline line should be able allowed to change the stacksize. Allowing the untrusted side changing the stacksize might have security implication.

Francis-Liu: Fixed.

tools/myst/host/exec_linux.c
```diff
@@ -313,6 +343,12 @@ static int _enter_kernel(
         }
     }
 
+    // Override main stack size if present in config
```
bodzhang, line 346: Does this mean the command line parameter will take effect, if the config.json does not contain a correspdoning entry? At least in "packaged mode", if the config.json does not contain main_stack_size entry, the developer's intention is to use the default stack size, which is captured in remote attestation. Commandline parameter should not be allowed to override the developer intention.

doc/user-getting-started-tee-aware.md
```diff
@@ -224,6 +224,9 @@ First we declare the desire in the config.json:
 
     // The heap size of the user application. Increase this setting if your app experienced OOM.
     "MemorySize": "1g",
+    // The stack size for the user application's main process. If increase heap size won't solve
+    // OOM, try increase this setting, e.g. to "8m".
+    "MainStackSize": "1536k",
```
bodzhang, line 229: main_stack_szie is an advanced configuration. Most developers probably do not need to bother with this configuration. The sample should use most common configuration and leave main_stack_szie as implicit default

kernel/enter.c
```diff
@@ -820,7 +820,8 @@ int myst_enter_kernel(myst_kernel_args_t* args)
                  args->envc,
                  args->envp,
                  NULL,
-                 NULL)) != 0)
+                 NULL,
+                 args->main_stack_size)) != 0)
```
bodzhang, line 824: Is it necessary to add a new input parameter for main_stack_size? It seems that other configurations did not get specific input parameter.

tools/myst/enc/enc.c
```diff
@@ -492,6 +493,7 @@ static long _enter(void* arg_)
     bool perf = false;
     bool report_native_tids = false;
     size_t max_affinity_cpus = options ? options->max_affinity_cpus : 0;
+    size_t main_stack_size = options ? options->main_stack_size : 0;
```
bodzhang, line 496: In case option is not provided, shouldn't main_stack_size be set to implicit default of MYST_PROCESS_INIT_STACK_SIZE?

tools/myst/host/exec_linux.c
Francis-Liu: The logic of the if condition below is the opposite: if the user has specified main_stack_size in config.json, that value will override the commandline parameter as in line 292.

However, if the user hasn't specified main_stack_size in config.json, and that user has specified the main_stack_size commandline option, that commandline option value will be effective.

The commandline parameter is supported by exec mode only. Package mode commandline doesn't recognize the main_stack_size parameter.

tools/myst/enc/enc.c
Francis-Liu: Correct. Actually the default value is set at line 767-768. In this function, the ultimate value `_kargs.main_stack_size` is determined in three steps:

1. In line 496, check commandline option value. If it is not provided, then set 0.
2. In line 631-634, check config.json parameter value. If it is set, override commandline option.
3. In line 767-768, if main_stack_size is set by neither commandline option nor config.json, then assign it with default value MYST_PROCESS_INIT_STACK_SIZE.

The reason why I didn't set main_stack_size to default value here in step 1, but in step 3, is in case user specify main_stack_size=0 in config.json by mistake. Step 3 will change that value to MYST_PROCESS_INIT_STACK_SIZE.

kernel/enter.c
Francis-Liu: You're right. It is interesting that `myst_exec` takes many input parameters but `crt_data`, `crt_size`, `crt_reloc_data`, `crt_reloc_data` can be retrieved from global data structure `__myst_kernel_args.` So does `main_stack_size`.

Francis-Liu: Fixed.

doc/user-getting-started-tee-aware.md
Francis-Liu: Fixed.

bodzhang (approved):

PR #697, opened 2021-08-11
Pin cpython test suite to python3.8.11 release

bodzhang:

Current implementation of /tests/cpython attempts to use the tip of the 3.8 branch of cpython github repo. Some recent commit ot the branch broke the Mytikos test suite. This PR pins the test suite to v3.8.11 release, and adds --nobrk to Mystikos configuration.


vtikoo (approved):
Currently we have to split the passing tests due to OOM issues. With ​--no-brk​, the situation might've improved. Could you try to see if we can revert the split with your PR?

PR #698, opened 2021-08-11
The design doc of secret provisioning

jxyang:


doc/design/secret-provisioning.md
```diff
@@ -0,0 +1,128 @@
...
+Users can plug in their own client library by including the library in rootfs
+and providing the corresponding `ClientLibPath` in config.json.
+
+## The server library
+
+Mystikos will provide a simple reference implementation of the server library
+that exposes the following APIs to be called by the service that releases the
+secrets:
+
+* `verify_attestation_evidence_with_key` (for attestation model 1)
```
mingweishih, line 115: Will the implementation of the APIs for model 1 and 2 directly base on OE's verify APIs? If so, this implies that the server takes hard dependency on the DCAP library.  Or will the implementation supports a model where the server talks remote verifier (e.g., MAA) to validate the SGX report first?

mingweishih (commented):
Another part that is missing in the doc is the support of init-time claims in the API.
For model 1, the init-time claims can just be appended to the evidence buffer, and the server API should do the extraction (see https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/InitTimeConfigurationInterface.md).

For model 2, the new OE APIs will support embedding init-time claims into the certificate (see https://github.com/openenclave/openenclave/pull/4138) and extract them from the certificate.

doc/design/secret-provisioning.md
jxyang: I clarified in the update. There are options that should be weighed by the users.

doc/design/secret-provisioning.md
```diff
@@ -0,0 +1,167 @@
...
+The providers of the secrets could be a key vault service such as AKV,
+M-HSM, or a service run by the user in a locked down and trusted
+environment.
+
+This design document describes the process of secret provisioning
+ with Mystikos in 5 aspects:
+
+1. **the user interface**: how does Mystikos allow users to specify the
+secrets they want in their applications?
+1. **the client library**: how does Mystikos provide a reference SSR
```
vtikoo, line 18: What is SSR?

doc/design/secret-provisioning.md
```diff
@@ -0,0 +1,167 @@
...
+        "Kid": <identifier of the key>,
+        "LocalPath": <Path to the file that will store the secret after its retrieval>,
+        "ClientLibPath": <Path to the client library>
+    },
+    ...
+]
+```
+
+Users are responsible for getting the dependent client library into the rootfs consumed
+by Mystikos. They are also responsible to linking with the corresponding server library
```
vtikoo, line 57:
```suggestion
by Mystikos. They are also responsible for linking with the corresponding server library
```

doc/design/secret-provisioning.md
```diff
@@ -0,0 +1,167 @@
...
+
+* `generate_attestation_evidence_with_keys` (for attestation model 1)
+* `generate_x509_cert_with_attestation_extension` (for attestation model 2)
+* `generate_attestation_service_token` (for attestation model 3)
+* `unwrap_secret` (for all attestation models)
+
+A client library could choose to implement only 1 or 2 `generate_xxx` functions
+for the specific attestation models supported on the server side. For example,
+the client library we provide for Azure M-HSM only implements
+`generate_attestation_service_token` (attestation model 3)
```
vtikoo, line 104:
```suggestion
`generate_attestation_service_token` (attestation model 3),
```

doc/design/secret-provisioning.md
jxyang: Secure Secret Release. Now should be replaced with "Secret Provisioning".

doc/design/secret-provisioning.md
```diff
@@ -0,0 +1,167 @@
...
+
+Users who wish to provision secrets to their applications running with
+Mystikos can add the following section to config.json:
+
+```json
+secrets: [
+    # Secret 1
+    {
+        "Name": <name of the secret>,
+        "AttestationModel": <a number between 1 to 3. See attestation models below>
```
Francis-Liu, line 37: I suggest using a meaningful keyword rather than a number.

doc/design/secret-provisioning.md
```diff
@@ -0,0 +1,167 @@
...
+        "HolderURL": <URL to the secret release service>,
+        "Kid": <identifier of the key>,
```
Francis-Liu, line 39: I suggest "KeyId" instead of "Kid", which sounds like a child.

mingweishih (commented):
I think this design can be built on top of newly proposed APIs in OE: https://github.com/openenclave/openenclave/pull/4138

Few ideas:
- The attestation model can become: `background_check` (model 1 & 2) and `passport` (model 3). This allows the design to take advantage of the init-time (when we support production-version exec mode in the future) and run-time claims (can be just the public key) feature.
- The `wrap_secret`/`unwrap_secret` can be implemented on top of attested TLS channel using `background_check` or `passport` certificate generated from OE APIs. This way, we can take advantage of the benefit of the TLS protocol.

doc/design/secret-provisioning.md
```diff
@@ -0,0 +1,167 @@
...
+]
+```
+
+Users are responsible for getting the dependent client library into the rootfs consumed
+by Mystikos. They are also responsible for linking with the corresponding server library
+in their services that release the secrets.
+
+## Attestation models
+
+At present, we consider 3 attestation models, from 1 to 3, that can be adopted by the
```
bodzhang, line 62: Is it necessary to expose the attestation model difference to the app developer? Can the client library abstract out the attestation protocol details?

doc/design/secret-provisioning.md
```diff
@@ -0,0 +1,167 @@
...
+
+## The client library
+
+Mystikos will provide a reference implementation of the client library
+that exposes the following APIs to be called by Mystikos runtime while
+performing secret provisioning:
+
+* `generate_attestation_evidence_with_keys` (for attestation model 1)
+* `generate_x509_cert_with_attestation_extension` (for attestation model 2)
+* `generate_attestation_service_token` (for attestation model 3)
```
bodzhang, line 98: Does the client lib need to invoke certain Mystikos APIs or access certain special file(s) created by the Mystikos runtime?

bodzhang: What does the Mystikos runtime do with the output of the three generate_xyz APIs? Treating the output as anopaque buffer and send the buffer to `HolderURL` with `Kid`?

bodzhang: Is the server at `HolderURL` expected to return a opaque blob (wrapped secret)?

doc/design/secret-provisioning.md
jxyang: `Kid` is used in M-HSM documents like [here](https://docs.microsoft.com/en-us/azure/key-vault/managed-hsm/hsm-protected-keys-byok). But I agree, `KeyID` is a better name.

doc/design/secret-provisioning.md
jxyang: It's hard to capture the nature of the models in simple words/phrases.

jxyang:
> I think this design can be built on top of newly proposed APIs in OE: [openenclave/openenclave#4138](https://github.com/openenclave/openenclave/pull/4138)
> 
> Few ideas:
> 
> * The attestation model can become: `background_check` (model 1 & 2) and `passport` (model 3). This allows the design to take advantage of the init-time (when we support production-version exec mode in the future) and run-time claims (can be just the public key) feature.
> * The `wrap_secret`/`unwrap_secret` can be implemented on top of attested TLS channel using `background_check` or `passport` certificate generated from OE APIs. This way, we can take advantage of the benefit of the TLS protocol.

This design doesn't necessarily depend on OE, that's the flexibility I want to drive. Model 1 and 2 don't use x509 certificates. InitTime claims is applicable to SGX, and when Mystikos targets SGX/OE, which is the only TEE we currently support, we can certainly take advantage of the new APIs in your design doc.

doc/design/secret-provisioning.md
jxyang: My current design is to have Mystikos runtime making different calls to the client library depending on the attestation model specified in config.json. When a client library supports multiple models, we want the Mystikos runtime to be able to differentiate between the models.

doc/design/secret-provisioning.md
jxyang: > Does the client lib need to invoke certain Mystikos APIs or access certain special file(s) created by the Mystikos runtime?

The reference client library will use the extended systcalls to generate quotes/certs. But a user-provided client library doesn't have to.

jxyang: > What does the Mystikos runtime do with the output of the three generate_xyz APIs? Treating the output as anopaque buffer and send the buffer to `HolderURL` with `Kid`?

That's exactly right.

jxyang: > Is the server at `HolderURL` expected to return a opaque blob (wrapped secret)?

Correct. The opaque blob is interpreted and secrets are extracted by the client lib.

doc/design/secret-provisioning.md
bodzhang: For a plugin that supports multiple protocols or options, an alternative to the "predefined attestation model" design is to support an opaque plugin-defined parameter, where the plugin exposes a single "generate_xyz" API, with an optional parameter, and the Mystikos runtime invokes the API with the parameter from the json configuration.

jxyang: The C language is notorious for its difficulty in supporting function overloading and optional parameters. Mystikos is implemented in C, and I think the current design is easier to implement.

bodzhang: Do the three "generate_xyz" APIs have different input parameter definition? I was assuming they don't take any input.

jxyang: They take pointers to the buffers that will hold the generated artifacts. They could be augmented with using nonce as one input. But so far no verifier I know of takes advantage of nonces. 

If we convert the attestation models as a parameter to the `generate_xyz` API, I think it actually exposes deeper internals of design and is harder for users to understand, and harder to adapt to a possible future re-implementation.

doc/design/secret-provisioning.md
```diff
@@ -0,0 +1,167 @@
...
+
+```json
+secrets: [
+    # Secret 1
+    {
+        "Name": <name of the secret>,
+        "AttestationModel": <a number between 1 to 3. See attestation models below>
+        "HolderURL": <URL to the secret release service>,
+        "Kid": <identifier of the key>,
+        "LocalPath": <Path to the file that will store the secret after its retrieval>,
```
mikbras, line 40: Are secrets obtained on demand or up-front?

jxyang: Up-front. We can add the on-demand later if there are customer requirements.

mikbras (approved):

PR #699, opened 2021-08-11
Temp workaround for gcov issue

bodzhang:

- Fix the bug in /tools/myst/host/Makefile that caused gcov build failure
- Temporary workaroud to address the  gcov test timeout issue due to switchless read/writer hostfs OCALLs, switching off the "switchless" configuration for those two OCALLs in gcov build

### DIFF
--- a/tools/myst/enc/Makefile
+++ b/tools/myst/enc/Makefile
@@ -57,5 +57,9 @@ EDGER8R_OPTS += --trusted
 EDGER8R_OPTS += --search-path $(OE_INCDIR)
 EDGER8R_OPTS += --trusted-dir $(SUBOBJDIR)
 
+ifdef MYST_ENABLE_GCOV
+EDGER8R_OPTS += -DGCOV
+endif
+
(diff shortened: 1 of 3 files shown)


tools/myst/gcov/myst.edl
```diff
@@ -0,0 +1,513 @@
...
+            size_t argp_size);
+
+        long myst_open_ocall(
+            [in, string] const char* pathname,
+            int flags,
+            mode_t mode,
+            uid_t uid,
+            gid_t gid);
+
+        long myst_read_ocall(
```
mingweishih, line 274: FYI: we can actually do something like
```
#ifdef GCOV
    long myst_read_ocall(
            int fd,
            [out, size=count] void* buf,
            size_t count);
#else
    long myst_read_ocall(
            int fd,
            [out, size=count] void* buf,
            size_t count)
    transition_using_threads;
#endif
```
Just need to do
```
EDGER8R_OPTS += -DGCOV
```
in the makefile to define the variable.

mingweishih (approved):

PR #701, opened 2021-08-12
Forward terminating signal to main process thread

paulcallen:
If the process thread has not already got an exit status, forward a signal
from a child thread to it.

### DIFF
--- a/kernel/signal.c
+++ b/kernel/signal.c
@@ -186,15 +186,34 @@ static long _default_signal_handler(unsigned signum)
     myst_thread_t* thread = myst_thread_self();
     myst_thread_t* process_thread = myst_find_process_thread(thread);
 
-    // A hard kill. Never returns.
-    // For SIGABRT, throw it to the process thread.
-    if (signum == SIGABRT)
+    // If the main thread has not been sent signal or has not exited already
+    // forward this exception to the main thread to cause the whole process to
+    // exit with this signal
+    if ((thread != process_thread) &&
+        ((signum == SIGABRT) || (signum == SIGSEGV)))
     {
-        myst_signal_deliver(process_thread, signum, NULL);
-        if (process_thread->signal.waiting_on_event)
+        enum myst_thread_status expected = MYST_RUNNING;
+        if (__atomic_compare_exchange_n(
+                &process_thread->status,
+                &expected,
+                MYST_RUNNING,
+                false,
+                __ATOMIC_RELEASE,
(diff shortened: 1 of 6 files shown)


kernel/signal.c
```diff
@@ -186,16 +186,35 @@ static long _default_signal_handler(unsigned signum)
...
+            {
+                myst_tcall_wake(process_thread->event);
+            }
         }
     }
+    else if (thread == process_thread)
+    {
+        // Make sure all other threads are shutdown properly.
+        // This is usually initiated via a process call to exit() or _Exit().
+        // If we are here then there is a good change it was not called
```
vtikoo, line 215:
```suggestion
        // If we are here then there is a good chance it was not called
```

tests/shutdown/parent.c
```diff
@@ -137,6 +247,10 @@ int main(int argc, const char* argv[])
     else if (
         strcmp("parent-wait-child-fork-exit-with-sighup-handler", argv[1]) == 0)
         test4(argv[1]);
+    else if (strcmp("child-process-child-thread-assert", argv[1]) == 0)
+        test5(argv[1]);
+    else if (strcmp("child-process-child-thread-crash", argv[1]) == 0)
+        test5(argv[1]);
```
vtikoo, line 253:
```suggestion
        test6(argv[1]);
```

tests/shutdown/parent.c
```diff
@@ -120,6 +122,114 @@ int test4(const char* test_name)
...
+// Child throws an SEGV on a child thread
+static void* _test6_thread_assert(void* arg)
+{
+    printf("hello from crashing thread\n");
+    char* crash = NULL;
+    crash[0] = 'a';
+    return NULL;
+}
+
+int test6(const char* test_name)
```
vtikoo, line 190: test5 and test6 can possibly share the thread creation and status checking bits.

paulcallen: I like to keep them separate unless the code is really complex. I have seen a situation where common code for a test is changed and one of the tests then does not test what it is supposed to

tests/shutdown/parent.c
paulcallen: fixed... and fixed test6 also :)

kernel/signal.c
paulcallen: fixed

vtikoo (approved):

PR #702, opened 2021-08-12
[DO NOT MERGE] Enable dotnet stack overflow test for linux target

vtikoo:
### Summary
- Set alternate stack of 8 pages for handling SIGSEGVs on linux target.
- For early stack overflow detection, make first and last page of stack for process threads inaccessible with `mprotect(PROT_NONE)`. Currently we are writing a guard pattern on these pages.
- For process threads aborted by a signal, if other threads exist, attempt to kill the thread group.
- Enable dotnet stack overflow test for linux target. The test is blocked for sgx target due to the overflowed stack being used for handling of the exception.

### DIFF
--- a/tests/dotnet-ubuntu/Makefile
+++ b/tests/dotnet-ubuntu/Makefile
@@ -23,7 +23,7 @@ clean:
 OPTS += --memory-size $(HEAP_SIZE)
 
 tests:
-	$(RUNTEST) $(MYST_EXEC) ext2fs /app/hello $(OPTS)
+	$(RUNTEST) ./exec.sh $(MYST_EXEC) $(OPTS) ext2fs /app/hello $(TARGET)
 
 ##############################
 # dev targets
--- a/tests/dotnet-ubuntu/exec.sh
+++ b/tests/dotnet-ubuntu/exec.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+$*
+ret=$?
+TARGET=$8
+
+if [ "$TARGET" == "sgx" -a "$ret" != "0" ]; then
+    exit 1
+fi
+
+if [ "$TARGET" == "linux" ]; then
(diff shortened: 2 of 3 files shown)


kernel/enter.c
```diff
@@ -867,6 +867,10 @@ int myst_enter_kernel(myst_kernel_args_t* args)
             myst_assume(thread->group_prev == NULL);
             while (thread->group_next)
             {
+                // We are currently only killing the thread group in
+                // SYS_exit_group. For threads aborted via a signal this flow is
+                // not hit. So another attempt to kill thread group.
+                myst_kill_thread_group();
```
jxyang, line 873: Can we do this in the signal handler before jumping back?

tools/myst/host/exec_linux.c
```diff
@@ -225,12 +225,27 @@ static void _install_signal_handlers()
...
         _err("Failed to register SIGILL signal handler\n");
     if (sigaction(SIGFPE, &sa, NULL) == -1)
         _err("Failed to register SIGFPE signal handler\n");
+
+    /* Set SIGSEGV handler to use alternate stack */
+    {
+        stack_t ss;
+        ss.ss_size = SIGSTKSZ * 4; // 8 pages
+        ss.ss_flags = 0;
+        if ((ss.ss_sp = malloc(SIGSTKSZ * 4)) == NULL)
```
jxyang, line 239: Are we freeing the alt stack somewhere?

kernel/enter.c
vtikoo: @paulcallen mentioned hes looking into a solution similar to that. So I'll probably back out this particular change.

tools/myst/host/exec_linux.c
vtikoo: What would be a good place to free the alt stack?

PR #704, opened 2021-08-12
added glibc test suite

salsal97:
uses pseudo fork


jxyang:
The blocking assertion failure here can be overcome by removing "assert(0)". The `mallopt` function modifies behaviors of `malloc` which don't apply to  musl `malloc` implementation anyway. We can simply use `return 0` to tell the caller there is an error.
 https://github.com/deislabs/mystikos/blob/3058d5031225cac26801323a03ae05d292b41bd0/crt/glibc_stub.c#L116

jxyang (approved):

PR #710, opened 2021-08-16
Remove exiting non-process thread from pipe wait queues

vtikoo:
If a non-process thread is killed by a SIGKILL while on a pipe operation
it exits without removing itself from the wait queue associated with the
pipe's conditional variable.
`myst_fdtable_remove_thread` iterates over the open pipedev file
descriptors and if the thread exists on the pipe's wait queue removes it.

### DIFF
--- a/include/myst/fdops.h
+++ b/include/myst/fdops.h
@@ -46,6 +46,8 @@ struct myst_fdops
 
     /* returns POLLIN | POLLOUT | POLLERR */
     int (*fd_get_events)(void* device, void* object);
+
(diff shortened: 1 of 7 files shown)


vtikoo:
coreclr p0 testsuite run [here](https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10118&view=logs&j=ab477ad6-9781-5b54-82ca-a0141b26726e).

vtikoo:
Will open another PR with cleaner fix.

PR #711, opened 2021-08-16
Process signals after clearing wait queues

vtikoo:
Currently we check for outstanding signals on the return path in
myst_tcall_wait. This short-circuiting can leave the higher level
synchronization mechanims like condition variables and mutexes in an
inconsistent state.
This patch moves the signal processing check into mutex and cond variable
implementation.

Signed-off-by: Vikas Tikoo <vitikoo@microsoft.com>

PR #713, opened 2021-08-17
[DO NOT MERGE] Return EINTR on signals from myst_tcall_wait

vtikoo:


vtikoo:
This approach is more unstable than #711 and #717. Closing for now.

PR #715, opened 2021-08-17
Fixed the bug related to handling the last ext2 dirent in a block

bodzhang:
The existing ext2 implementation assumes the last dirent entry always fill to the end of the disk block. In a loop to traverse the dirent list within a buffer, the implementation can read beyond the boundary of the buffer. 

### DIFF
--- a/ext2/ext2.c
+++ b/ext2/ext2.c
@@ -1143,12 +1143,19 @@ static const ext2_dirent_t* _find_dirent(
     const uint8_t* end = (uint8_t*)data + size;
     size_t len = strlen(name);
 
-    while (p < end)
+    /* The end of the block might not have a full ext2_dirent structure. Make
+     * sure the fixed-length header portion is in range before accessing */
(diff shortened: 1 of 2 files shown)


ext2/ext2.c
```diff
@@ -1551,7 +1558,9 @@ static int _count_dirents(
     if ((end - p) % ext2->block_size)
         ERAISE(-EINVAL);
 
-    while (p < end)
+    /* The end of the block might not have a full ext2_dirent structure. Make
+     * sure the fixed-length header portion is in range before accessing */
+    while (p + sizeof(ext2_dirent_t) - EXT2_PATH_MAX < end)
```
vtikoo, line 1563: The `if (p !=end)` check after the while loop might fail if its the last block with trailing zeros?

bodzhang: changed the logic to `if (p > end)` in the latest update

PR #721, opened 2021-08-19
remove flaky tests for gcov + others

salsal97:


mingweishih (commented):
Other than moving the flaky tests to the remove list, can we document what tests are flaky, and what are unsupported?

mingweishih (approved):
Realized that we have tests.failed for unsupported tests, and the tests.remove for flaky tests.

PR #724, opened 2021-08-20
Add myst_is_bad_addr

mingweishih:
This PR adds `myst_is_bad_addr`, which checks the following
- Whether the pages within the address range are mapped (based on the pid vectors)
   - The PR also extends the pid vectors to track both process stack and crt data
- Whether the address range has corresponding page permissions

The checks are based on the assumption of using the `--nobrk` option.

The PR also adds  `myst_is_bad_addr_read`,  `myst_is_bad_addr_write`, and `myst_is_bad_addr_read_write` for convenience  based on `myst_is_bad_addr`.

### DIFF
--- a/kernel/exec.c
+++ b/kernel/exec.c
@@ -1094,27 +1094,28 @@ int myst_exec(
     free(hashbang_buff);
     hashbang_buff = NULL;
 
-    /* if this is a nested exec, then release previous exec's stack */
-    if (process->exec_stack)
-    {
-        myst_munmap(process->exec_stack, process->exec_stack_size);
-        process->exec_stack = NULL;
-        process->exec_stack_size = 0;
-    }
+    /* if this is a nested exec, then release previous process mappings
+     * assicoated to the pid */
+    myst_release_process_mappings(process->pid);
 
-    /* if this is a nested exec, then release previous exec's CRT data */
-    if (process->exec_crt_data && process->exec_crt_size)
-    {
-        myst_munmap(process->exec_crt_data, process->exec_crt_size);
-        process->exec_crt_data = NULL;
-        process->exec_crt_size = 0;
-    }
+    /* set pid to the newly allocated crt data (now part of the process
+     * mappings) */
+    if (myst_mman_pids_set(crt_data, crt_size, process->pid) != 0)
+        myst_panic("myst_mman_pids_set()");
 
-    /* The thread is responsible for freeing the stack */
+    /* set pid to the newly allocated stack (now part of the process
+     * mappings) */
+    if (myst_mman_pids_set(
+            stack, __myst_kernel_args.main_stack_size, process->pid) != 0)
+        myst_panic("myst_mman_pids_set()");
+
+    /* used by myst_syscall_get_process_thread_stack */
     process->exec_stack = stack;
     process->exec_stack_size = __myst_kernel_args.main_stack_size;
+#ifdef MYST_THREAD_KEEP_CRT_PTR
     process->exec_crt_data = crt_data;
     process->exec_crt_size = crt_size;
+#endif
 
     /* close file descriptors with FD_CLOEXEC flag */
     {
@@ -1152,14 +1153,16 @@ int myst_exec(
 
     process->exec_stack = NULL;
     process->exec_stack_size = 0;
+#ifdef MYST_THREAD_KEEP_CRT_PTR
     process->exec_crt_data = NULL;
     process->exec_crt_size = 0;
+#endif
     ERAISE(-ENOEXEC);
 
 done:
 
     if (stack)
-        free(stack);
+        myst_munmap(stack, __myst_kernel_args.main_stack_size);
 
     if (crt_data)
         myst_munmap(crt_data, crt_size);
--- a/kernel/mmanutils.c
+++ b/kernel/mmanutils.c
@@ -1150,13 +1150,10 @@ static long _handle_mman_pids_op(
     ECHECK((ssize_t)(index = _get_page_index(addr, length)));
     count = length / PAGE_SIZE;
 
-    assert(index < v.pids_count);
-    assert(index + count <= v.pids_count);
-
     if ((size_t)index >= v.pids_count)
         ERAISE(-ERANGE);
 
-    if (index + count >= v.pids_count)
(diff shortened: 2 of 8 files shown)


kernel/thread.c
```diff
@@ -996,6 +1001,11 @@ static long _run_thread(void* arg_)
             {
                 long r = myst_munmap(
                     thread->main.exec_crt_data, thread->main.exec_crt_size);
+                if (myst_mman_pids_set(
```
anakrish, line 1004: `myst_mman_pids_set` with process id `0` follows `myst_munmap` in this PR.
Can `myst_munmap` do `myst_mman_pids_set` instead of the caller having to do it?

kernel/syscall.c
```diff
@@ -516,6 +516,30 @@ static bool _iov_bad_addr(const struct iovec* iov, int iovcnt)
     return false;
 }
 
+static int _brk_pids_set(uint64_t new_brk)
+{
+    int ret = 0;
+    static uint64_t old_brk = 0;
+
+    if (!old_brk)
```
anakrish, line 524: Does the context guarantee that `_brk_pids_set` will always be called with appropriate locks?
If yes, it is better to add a comment reflecting that.
If no, then `_brk_pids_set` needs to modify `old_brk` in a thread-safe manner.

kernel/mmanutils.c
```diff
@@ -1148,6 +1148,15 @@ ssize_t myst_mman_pids_test(const void* addr, size_t length, pid_t pid)
     return _handle_mman_pids_op(MMAN_PIDS_OP_TEST, addr, length, pid);
 }
 
+bool myst_is_within_process(const void* addr, pid_t pid)
+{
+    uint64_t addr_page = myst_round_down_to_page_size((uint64_t)addr);
```
anakrish, line 1153: How would `myst_is_within_process` handle shared memory?

kernel/mmanutils.c
```diff
@@ -1155,6 +1152,37 @@ ssize_t myst_mman_pids_test(const void* addr, size_t length, pid_t pid)
...
+    bool ret = true;
+    uint64_t page_addr;
+
+    if (!addr || !length)
+        goto done;
+
+    /* check if the pages within the address range are unmapped (i.e.,
+     * associated pid is zero). */
+    page_addr = myst_round_down_to_page_size((uint64_t)addr);
+    if (myst_mman_pids_test((const void*)page_addr, length, 0) > 0)
```
anakrish, line 1166: since addr is being rounded down, length needs to be updated.
[page_addr, page_addr+length] may not always enclose [addr, addr+length]

mingweishih: the length will be rounded up in both `myst_mman_pids_test` and `myst_mman_get_prot`. But I can also do here to ensure the consistency.

kernel/syscall.c
mingweishih: brk is not supported by this PR

kernel/mmanutils.c
mingweishih: the updated PR is now simply checking if the page is mapped or not instead of the ownership.

kernel/exec.c
```diff
@@ -1094,27 +1094,24 @@ int myst_exec(
...
     hashbang_buff = NULL;
 
-    /* if this is a nested exec, then release previous exec's stack */
-    if (process->exec_stack)
-    {
-        myst_munmap(process->exec_stack, process->exec_stack_size);
-        process->exec_stack = NULL;
-        process->exec_stack_size = 0;
-    }
+    /* if this is a nested exec, then release previous process mappings */
```
vtikoo, line 1097: Comment is obsolete now?

kernel/exec.c
```diff
@@ -1094,27 +1094,24 @@ int myst_exec(
...
+    myst_release_process_mappings(process->pid);
```
vtikoo, line 1098: For all process threads except the main thread, we release process mappings when the thread long jumps back to the exit sequence [in _run_thread](https://github.com/deislabs/mystikos/blob/main/kernel/thread.c#L1054). Is there a reason to duplicate it/move it here?

vtikoo: Oh nevermind, the preceding comment makes sense now. However is there a check for nested exec missing?

kernel/thread.c
```diff
@@ -1053,6 +1032,26 @@ static long _run_thread(void* arg_)
             /* unmap any mapping made by the process */
             myst_release_process_mappings(process->pid);
 
+            if (process->exec_stack)
```
vtikoo, line 1035: Just to confirm, both the exec stack and crt data are setup by the process itself?

kernel/exec.c
mingweishih: This call covers the nested exec already:
- For non-nested exec, it acts as no-op as there is no mapping associated with the process->pid
- For nested exec, it releases the previous mapping associated with the process->pid (retaining the exiting behavior).

vtikoo: > For non-nested exec, it acts as no-op as there is no mapping associated with the process->pid

How about crt_data and exec_stack?

mingweishih: This PR adds these two region to the pid vectors, which will be freed by myst_release_process_mappings per pid

kernel/thread.c
mingweishih: Yes. The logic here is only for the main thread.

kernel/exec.c
vtikoo: I'm still a bit confused. For the non-nested exec case, crt_data and exec_stack are being `myst_mmap`ed earlier in the function. Wouldn't an unconditional release of mappings here cause issues?

mingweishih: The pid vectors update is done after here, see line 1101 - 1106. The call here only affects the existing mapping associated with the pid. The newly allocated stack/crt are set to the `process` later.

vtikoo: Ah gotcha, thanks for clarifying that. Could you elaborate on this in the code comment for future reference of someone looking at this code.

vtikoo (approved):
Looks good. Have a minor comment about expanding on a comment.

mingweishih:
Merge as Jenkins is passing

PR #728, opened 2021-08-24
delete main process on exit

paulcallen:

### DIFF
--- a/kernel/enter.c
+++ b/kernel/enter.c
@@ -931,6 +931,9 @@ int myst_enter_kernel(myst_kernel_args_t* args)
         free(process->cwd);
         process->cwd = NULL;
 
+        free(process);
+        process = NULL;
+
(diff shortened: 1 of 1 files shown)


mingweishih (approved):

kernel/enter.c
```diff
@@ -931,6 +931,8 @@ int myst_enter_kernel(myst_kernel_args_t* args)
         free(process->cwd);
         process->cwd = NULL;
 
+        free(process);
```
asvrada, line 934: Maybe set process to NULL?

paulcallen: The whole kernel is shutting down at this point so I didnt think it was worth it, but I will add this in case someone tries to use it after my check-in

paulcallen: fixed.

PR #729, opened 2021-08-24
Add support for futex bitset and realtime clock

Sahakait:
Futex syscall now supports `FUTEX_CLOCK_REALTIME` and `FUTEX_{WAKE,WAIT}_BITSET`.

**Changes**:
Thread Queue
image
No function outside of `thread.h` should modify thread queues. Bitset was added to the thread struct. An alternative design was considered with thread queue nodes, such that queue nodes would keep track of threads and their corresponding bitset. This design has the benefit of separating condition logic from the thread struct. However, we would need to `malloc` and `free` queue nodes.
(Below: alternative thread queue design)
image

Real time clock
This is implemented by converting the timeout parameter into a relative timeout.

Bitset support
Most of the change is in `cond.c`. See [futex manpage](https://man7.org/linux/man-pages/man2/futex.2.html) for expected functionality. 

**Testing**:
Added test cases for wait/wake with bitset in `test\futex`.
Added test case for realtime clock in `test\futex`.
Added new test file in `test\thread_queue`


kernel/futex.c
```diff
@@ -359,19 +375,35 @@ long myst_syscall_futex(
...
+    if (op & FUTEX_CLOCK_REALTIME)
+    {
+        if (!((_op == FUTEX_WAIT_BITSET) ||
+              (_op == FUTEX_WAIT))) // || (op | FUTEX_WAIT_REQUEUE_PI)
+            ERAISE(-EINVAL);
+
+        clock = CLOCK_REALTIME;
+    }
+
+    if (_op == FUTEX_WAIT_BITSET || _op == FUTEX_WAKE_BITSET)
```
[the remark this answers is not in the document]
Sahakait: Done

include/myst/thread.h
```diff
@@ -81,7 +82,8 @@ struct myst_thread
     uint64_t magic;
 
     /* used by myst_thread_queue_t (condition variables and mutexes) */
-    struct myst_thread* qnext;
+    struct myst_thread* _qnext;
```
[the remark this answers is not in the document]
Sahakait: The change was made to highlight that only the `thread_queue` functions that live inside `thread.h` should access the fields directly. I'm open to suggestions on what the best way to achieve that is.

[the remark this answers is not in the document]
mingweishih: For consistency, I think removing `_` prefix is better. This concept does not make much sense for struct members to me (unless there are precedence).

include/myst/thread.h
```diff
@@ -405,13 +438,66 @@ MYST_INLINE int myst_thread_queue_remove_thread(
     return found ? pos : -1;
 }
 
+/** Find and remove up to `n` threads from the queue such that
+ * <waiting bitset> & <passed in bitset> != 0.
+ * Removed threads are placed in the `matches` queue.
+ */
+MYST_INLINE size_t myst_thread_queue_search_remove_bitset(
```
mingweishih, line 445: use return type `int` given you're returning -1 on failure

kernel/cond.c
```diff
@@ -141,61 +150,63 @@ int myst_cond_signal_thread(myst_cond_t* c, myst_thread_t* thread)
...
+    myst_thread_queue_t queue = {NULL, NULL};
 
     if (!c)
         return -EINVAL;
 
     myst_spin_lock(&c->lock);
-
-    waiter = myst_thread_queue_pop_front(&c->queue);
-
+    myst_thread_queue_search_remove_bitset(&c->queue, &queue, 1, bitset);
```
mingweishih, line 162: check return

kernel/cond.c
```diff
@@ -141,61 +150,63 @@ int myst_cond_signal_thread(myst_cond_t* c, myst_thread_t* thread)
...
-        /* Select at most n waiters to be woken up */
-        while (i < n && (p = myst_thread_queue_pop_front(&c->queue)))
-        {
-            myst_thread_queue_push_back(&waiters, p);
-            i++;
-        }
-    }
+    myst_spin_lock(&c->lock);
+    /* Select at most n waiters to be woken up */
+    myst_thread_queue_search_remove_bitset(&c->queue, &waiters, n, bitset);
```
mingweishih, line 192: check return

include/myst/thread.h
```diff
@@ -405,13 +438,66 @@ MYST_INLINE int myst_thread_queue_remove_thread(
...
+            num_found++;
+            if (prev != NULL)
+            {
+                prev->_qnext = next;
+                if (next == NULL)
+                    queue->_back = prev;
+            }
+            else
+            {
+                queue->_front = queue->_front->_qnext;
```
mingweishih, line 476: Isn't `queue->_front->_qnext` the same as `next`? If so, is 477 & 478 the same as 471 & 472?

kernel/futex.c
```diff
@@ -359,19 +375,35 @@ long myst_syscall_futex(
 {
     long ret = 0;
 
-    (void)val3;
+    uint32_t bitset = FUTEX_BITSET_MATCH_ANY;
+    clockid_t clock = CLOCK_MONOTONIC;
+
+    int _op = op & ~(FUTEX_CLOCK_REALTIME | FUTEX_PRIVATE);
```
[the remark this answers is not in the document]
Sahakait: image
The flags are not designed for this. For example, `op & FUTEX_WAKE` would be true for both `op =  FUTEX_WAKE` and `op = FUTEX_REQUEUE`.

include/myst/thread.h
Sahakait: Can you elaborate on the reasoning for this? I based this function off of `myst_thread_queue_remove_thread()`.
image

include/myst/thread.h
```diff
@@ -405,13 +438,66 @@ MYST_INLINE int myst_thread_queue_remove_thread(
...
 
+/** Find and remove up to `n` threads from the queue such that
+ * <waiting bitset> & <passed in bitset> != 0.
+ * Removed threads are placed in the `matches` queue.
+ */
+MYST_INLINE size_t myst_thread_queue_search_remove_bitset(
+    myst_thread_queue_t* queue,
+    myst_thread_queue_t* matches,
+    size_t n,
+    uint32_t bitset)
```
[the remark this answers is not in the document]
Sahakait: I based the variable name off of the operation name `FUTEX_WAIT_BITSET`. Do you think bitmask would provide more clarity?

include/myst/thread.h
Sahakait: This handles removing the only node
```
if (queue->_front == NULL)
                    queue->_back = NULL;
```

This handles removing the last node
```
if (next == NULL)
                    queue->_back = prev;

```

Sahakait: I simplified the cases.

include/myst/cond.h
```diff
@@ -27,13 +28,20 @@ int myst_cond_timedwait(
     myst_cond_t* c,
     myst_mutex_t* mutex,
     const struct timespec* timeout);
+int myst_cond_timedwait_ops(
```
mikbras, line 31: Please add blank line between definitions (struct and function).

include/myst/cond.h
```diff
@@ -27,13 +28,20 @@ int myst_cond_timedwait(
...
+    myst_cond_t* c,
+    myst_mutex_t* mutex,
+    const struct timespec* timeout,
+    uint32_t bitset);
 
 int myst_cond_signal(myst_cond_t* c);
+int myst_cond_signal_ops(myst_cond_t* c, uint32_t bitset);
```
mikbras, line 38: Here too.

include/myst/cond.h
```diff
@@ -27,13 +28,20 @@ int myst_cond_timedwait(
...
 
 int myst_cond_signal_thread(myst_cond_t* c, myst_thread_t* thread);
 
 /* Wake up n waiters */
 int myst_cond_broadcast(myst_cond_t* c, size_t n);
+int myst_cond_broadcast_ops(myst_cond_t* c, size_t n, uint32_t bitset);
```
mikbras, line 44: Blank line here too.

include/myst/futex.h
```diff
@@ -18,12 +18,21 @@
...
+
 int myst_futex_wait(int* uaddr, int val, const struct timespec* to);
+int myst_futex_wait_ops(
+    int* uaddr,
+    int val,
+    const struct timespec* to,
+    uint32_t bitset);
 
 int myst_futex_wake(int* uaddr, int val);
+int myst_futex_wake_ops(int* uaddr, int val, uint32_t bitset);
```
mikbras, line 36: Please add blank line before definition to follow existing style.

include/myst/futex.h
```diff
@@ -18,12 +18,21 @@
...
 #define FUTEX_WAIT_BITSET    9
+#define FUTEX_WAKE_BITSET    10
 #define FUTEX_PRIVATE        128
 #define FUTEX_CLOCK_REALTIME 256
 // clang-format on
 
+#define FUTEX_BITSET_MATCH_ANY 0xffffffff
+
 int myst_futex_wait(int* uaddr, int val, const struct timespec* to);
+int myst_futex_wait_ops(
```
mikbras, line 29: Please add blank line before this. I wish the formatter could do that but unfortunately it cannot.

mikbras, line 29: I wonder whether ``myst_futex_wait_bitset()`` might be a better name (since that is the only parameter that is different).

include/myst/cond.h
mikbras, line 38: I think I would prefer changing the existing ``myst_cond_signal()`` prototype and adding a ``uint32* bitset`` parameter.

include/myst/thread.h
```diff
@@ -342,50 +344,81 @@ MYST_INLINE bool myst_is_zombied_process(myst_process_t* process)
...
     myst_thread_queue_t* queue)
 {
-    myst_thread_t* thread = queue->front;
+    if (queue)
+        return queue->_front;
+
+    return NULL;
+}
+
+MYST_INLINE myst_thread_t* myst_thread_queue_pop_front_bitset(
```
mikbras, line 393: The bitset seems unrelated to the thread queue operation (bitset is a member of the thread not the queue).

include/myst/thread.h
```diff
@@ -398,22 +431,22 @@ MYST_INLINE int myst_thread_queue_remove_thread(
...
+                prev->_qnext = t->_qnext;
+                if (t->_qnext == NULL)
+                    queue->_back = prev;
             }
             else
             {
-                queue->front = queue->front->qnext;
-                if (queue->front == NULL)
-                    queue->back = NULL;
+                queue->_front = queue->_front->_qnext;
```
mikbras, line 447: The prefixing of fields with underscores seems unrelated to this PR and I would ask that it be backed out and discussed separately. I understand the intent is to enforce data encapsulation in the hopes of discouraging use of various fields in certain contexts, but it would be a reversal of the previous authors design decision and I think the discussion could be handled separately.

mikbras (changes requested):

include/myst/thread.h
mingweishih: The reference you put uses int

include/myst/thread.h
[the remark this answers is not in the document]
mingweishih: According to https://man7.org/linux/man-pages/man2/futex.2.html, `BITSET` refers to an operation of setting a bit mask.

> FUTEX_WAIT_BITSET (since Linux 2.6.25)
              This operation is like FUTEX_WAIT except that val3 is used
              to provide a 32-bit bit mask to the kernel.

include/myst/thread.h
Sahakait: the naming of this operation signifies the ability to get the bitset data when popping the thread queue as well. Do you have suggestions on how to handle this?

include/myst/thread.h
Sahakait: Got it, I initially misread your comment as removing the MYST_INLINE

include/myst/thread.h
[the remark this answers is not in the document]
Sahakait: I think bitset in this context means "set of bits" rather than "setting bits". For example, C++ has data structure called [bitset](https://www.cplusplus.com/reference/bitset/bitset/). I would prefer to keep the name bitset as it also aligns with the operation name.

include/myst/futex.h
Sahakait: That was the original name, until I realized that we may have to pass on more parameters for the clock type. I renamed all the operations with additional "optional" arguments to `ops` for consistency.

kernel/futex.c
```diff
@@ -358,20 +374,45 @@ long myst_syscall_futex(
...
 {
     long ret = 0;
+    uint32_t bitset = FUTEX_BITSET_MATCH_ANY;
+    int _op = op & ~(FUTEX_CLOCK_REALTIME | FUTEX_PRIVATE);
 
-    (void)val3;
+    if (op & FUTEX_CLOCK_REALTIME)
+    {
+        if (!((_op == FUTEX_WAIT_BITSET) ||
+              (_op == FUTEX_WAIT))) // || (op | FUTEX_WAIT_REQUEUE_PI)
```
mikbras, line 383: Suggest taking out commented out clause.

mikbras:
Sahakait: This is looking good but I will need more time to review. Will get back to it soon.

include/myst/cond.h
Sahakait: `pipedev.c` and `eventfddev.c` uses the original function, meaning we would have to add bitset to that context, which doesn't necessarily need to know about it.

kernel/futex.c
Sahakait: I added that there since according to the specs `FUTEX_WAIT_REQUEUE_PI` is allowed to have the `FUTEX_CLOCK_REALTIME` set as well, in case we implement this functionality in the future. If that is more distracting than helpful, I can remove it.

include/myst/thread.h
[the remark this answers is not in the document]
mikbras: I prefer bitset because it matches the Linux definition.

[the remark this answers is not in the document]
mikbras: Also, Linux itself uses the name `bitset` within the kernel in the futex implementation.

include/myst/cond.h
mikbras: You can pass null into those context.

include/myst/futex.h
mikbras: But currently, the only additional parameter is ``bitset``. I think just adding additional (possibly null) parameters to ``myst_futex_wait`` generates fewer changes.

include/myst/thread.h
```diff
@@ -5,6 +5,7 @@
 #define _MYST_THREAD_H
```
mikbras, line 5: Please don't use this technique to work around circular dependencies. I think there is a better way to resolve.

include/myst/thread.h
```diff
@@ -208,6 +209,7 @@ struct myst_thread
 
     /* used by myst_thread_queue_t (condition variables and mutexes) */
     struct myst_thread* qnext;
+    uint64_t bitset;
```
mikbras, line 212: Should be `uint32_t since `bitset` is `uint32_t` elsewhere and the mask itself is 32-bits.

include/myst/thread.h
```diff
@@ -370,8 +374,25 @@ MYST_INLINE void myst_thread_queue_push_back(
...
 
-MYST_INLINE myst_thread_t* myst_thread_queue_pop_front(
+MYST_INLINE void myst_thread_queue_push_back(
+    myst_thread_queue_t* queue,
+    myst_thread_t* thread)
+{
+    myst_thread_queue_push_back_bitset(queue, thread, FUTEX_BITSET_MATCH_ANY);
+}
+
+MYST_INLINE myst_thread_t* myst_thread_queue_get_front(
```
mikbras, line 384: I don't see the need for this function. Also it introduces a branch that was not in the original code. It is used in contexts where the queue is known to be non-null.

kernel/cond.c
```diff
@@ -34,7 +34,7 @@ int myst_cond_destroy(myst_cond_t* c)
     myst_spin_lock(&c->lock);
 
     /* Fail if queue is not empty */
-    if (c->queue.front)
+    if (!myst_thread_queue_empty(&c->queue))
```
mikbras, line 37: I would like to remove this accessor since the only consumer of this field is the condition itself.

kernel/futex.c
```diff
@@ -269,6 +280,11 @@ int myst_futex_wake(int* uaddr, int val)
     return ret;
 }
 
+int myst_futex_wake(int* uaddr, int val)
+{
+    return myst_futex_wake_ops(uaddr, val, FUTEX_BITSET_MATCH_ANY);
```
mikbras, line 285: Does this function now require checking of futex-bitsets? If so, it modifies its original behavior for the non-bitset case, which could introduce instability. I would prefer that existing code paths remain exactly as before.

kernel/futex.c
```diff
@@ -358,20 +374,45 @@ long myst_syscall_futex(
     int val3)
 {
     long ret = 0;
+    uint32_t bitset = FUTEX_BITSET_MATCH_ANY;
+    int _op = op & ~(FUTEX_CLOCK_REALTIME | FUTEX_PRIVATE);
```
mikbras, line 378: we use underscore prefixes to denote static definitions.

kernel/mutex.c
```diff
@@ -8,6 +8,7 @@
 #include <stdbool.h>
 #include <string.h>
 
+#include <myst/cond.h>
```
mikbras, line 11: Including cond.h here introduces a circular inclusion dependency (cond.h also includes mutex.h). This was made necessary be the definition of the `myst_thread_queue_empty()`` accessors which I would remove.

kernel/mutex.c
```diff
@@ -42,7 +43,7 @@ int __myst_mutex_trylock(myst_mutex_t* m, myst_thread_t* self)
     if (m->owner == NULL)
     {
         /* If the waiters queue is empty */
-        if (m->queue.front == NULL)
```
mikbras, line 45: All changes to mutex.h are unecessary and were introduced by the accessors for condition. Mutexes and conditions are tightly coupled intentionally and mutex should not be considered a consumer of condition.

mikbras:
Overall the spirit of the change looks good, but I have the following concerns.
- The change may possibly modify existing behavior when bit sets are not in use. I need to spend more time looking at this.
- There are preferential changes that complicate the PR and make it hard to review. The use of accessors for example (between mutex and condition) creates a lot of churn and makes it hard to really see the essential changes. Also the accessors have required header inclusion cycles.

mikbras:
Sakakait: any idea why this fails the pipeline. It looks like the Java tests fails and crashes before running other tests.

include/myst/thread.h
mikbras: Please disregard. I am concerned about circular inclusions elsewhere though.

mikbras:
Hi Sahakai:

I checked out your branch and spent some time examining and testing. The futex tests fails often under stress. I run the following script on my machine (you will have to install sysbench on your machine).

Here is my ``stress`` script:

```C++
#!/bin/bash
while true
do
        sysbench --test=cpu --cpu-max-prime=10000000 --threads=16 run
done
```

Sahakait:
> Sakakait: any idea why this fails the pipeline. It looks like the Java tests fails and crashes before running other tests.

I just rebased and I see additional failures, I suspect this may have to do with merging this PR with other changes in `cond.c`, will need to investigate further.

Sahakait:
> Hi Sahakait:
> 
> I checked out your branch and spent some time examining and testing. The futex tests fails often under stress. I run the following script on my machine (you will have to install `sysbench` on your machine).
> 
> ```c++
> #!/bin/bash
> while true
> do
>         sysbench --test=cpu --cpu-max-prime=10000000 --threads=16 run
> done
> ```
> 
> Also, `tests/hello_world/java` hangs on my machine.

This is likely due to the fact that the wait tests uses a timer to make sure the wait duration is within a certain threshold of the specified wait time.

kernel/mutex.c
Sahakait: `kernel/mutex.c` has been reverted.

kernel/mutex.c
Sahakait: `kernel/mutex.c` has been reverted.

kernel/futex.c
Sahakait: `_op` renamed.

kernel/cond.c
Sahakait: Reverted.

Sahakait:
Resubmitted as a separate PR.

PR #733, opened 2021-08-25
EXPERIMENTAL: defer inode free until last file is closed

mikbras:


ext2/ext2.c
```diff
@@ -59,6 +59,25 @@ struct myst_file
...
+{
+    assert(_valid_ino(ext2, ino));
+    ext2->inode_refs[ino - 1].nopens++;
+}
+
+static ext2_ino_t _inode_unref(ext2_t* ext2, ext2_ino_t ino)
+{
+    assert(_valid_ino(ext2, ino));
+    assert(ext2->inode_refs[ino - 1].nopens > 0);
+    return --ext2->inode_refs[ino - 1].nopens;
```
radhikaj, line 78: Do the refcount increases and decreases need to be atomic?

mikbras: Thank you for looking at this. Atomic operations are not needed at this level because the kernel performs locking on per-operation basis.

mikbras: In other words, the kernel guarantees that only one thread can enter the EXT2 file system at a time.

PR #737, opened 2021-08-28
Add thread destructors when terminating signal

paulcallen:
when a thread handles a signal that is not handled the thread can be
doing anything at the time and may be left in a state that can cause
locks to be left helf, or threads can be left in queues.

this PR adds the ability to queue a 'destructor' at various points
within the callstack, and if an unhandled signal is generated all
destructors that are still registered will be executing, allowing each
destructor to do the necessary cleanup.

initially this functionality is being used by the condition variable and
mutex to unregister itself from the thread waiter queue before the
thread gets torn down.

other places where this code can be used is for things like the
filesystem lock that can generate a signal while the lock is held and if
left locked would cause any other process trying to access the
filesystem to hang. A destructor here would automatically release the
lock when the signal is being processed.


mikbras (approved):

kernel/mutex.c
```diff
@@ -66,10 +67,61 @@ int __myst_mutex_trylock(myst_mutex_t* m, myst_thread_t* self)
...
+        (myst_mutex_thread_sig_handler_t*)_sig_handler;
+    myst_thread_t* thread = myst_thread_self();
+
+    myst_spin_lock(&sig_handler->mutex->lock);
+
+    myst_thread_queue_remove_thread(&sig_handler->mutex->queue, thread);
+
+    thread->signal.waiting_on_event = false;
+
+    if (sig_handler->mutex->queue.front != NULL)
```
vtikoo, line 96: Should we do the wake only if there is no `owner` for the mutex? The owner should be responsible for waking the next waiter.

PR #739, opened 2021-08-30
Fix myst_release_process_mappings() and add nobrk config to numpy test suite

bodzhang:

### DIFF
--- a/kernel/mmanutils.c
+++ b/kernel/mmanutils.c
@@ -663,7 +663,7 @@ int myst_release_process_mappings(pid_t pid)
                     }
 
                     /* count consecutive pages with same pid */
-                    for (size_t j = i + 1; j < m; j++)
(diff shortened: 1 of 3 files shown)


kernel/mmanutils.c
```diff
@@ -693,6 +693,10 @@ int myst_release_process_mappings(pid_t pid)
                         ERAISE(-EINVAL);
 #endif
                     }
+                    else
+                    {
+                        myst_mman_pids_set(addr, len, 0);
```
mingweishih, line 698: I'm thinking whether this can be done regardless myst_munmap succeeds or not (given that the fdmappings is freed already). It's unclear what's the expected behavior if myst_munmap fails here.

mingweishih (approved):

mikbras (approved):

PR #741, opened 2021-08-31
Switch to OE's OpenSSL crypto implementation

### LINKED ISSUE #392: Use OpenSSL by default in OE layer for better performance
OE team has observed much better runtime performance on attestation operations when switching to OpenSSL from MbedTLS. Mystikos applications could benefit from it if doing the switch. To do so, we need

- Update the OE version that mystikos points to such that OpenSSL will be built by default.
- Update defs.mak (https://github.com/deislabs/mystikos/blob/main/defs.mak#L89) such that the OpenSSL libraries will be picked by the linker.
  - Optionally, one can add a variable to choose which crypto implementation to use. However, I guess having a good default is preferable.

Related issue #12: mystikos can consider using OpenSSL as default crypto implementation on the host as well.

mingweishih:
Fixes #392


defs.mak
```diff
@@ -94,7 +94,7 @@ OEENCLAVE_LIBDIR = $(BUILDDIR)/openenclave/lib/openenclave/enclave
 
 OEENCLAVE_LDFLAGS = $(shell PKG_CONFIG_PATH=$(BUILDDIR)/openenclave/share/pkgconfig pkg-config oeenclave-gcc --libs)
 
-OEENCLAVE_LDFLAGS += $(shell PKG_CONFIG_PATH=$(BUILDDIR)/openenclave/share/pkgconfig pkg-config oeenclave-gcc --variable=mbedtlslibs)
+OEENCLAVE_LDFLAGS += $(shell PKG_CONFIG_PATH=$(BUILDDIR)/openenclave/share/pkgconfig pkg-config oeenclave-gcc --variable=openssllibs)
```
jxyang, line 97: Should we use a Makefile flag to allow users choose between mbedtls and openssl?

tools/myst/enc/Makefile
```diff
@@ -47,7 +47,12 @@ LIBS += $(LIBDIR)/libmystgcovclient.a
 endif
 
 LDFLAGS += $(OEENCLAVE_LDFLAGS)
-LDFLAGS += -lcrypto
+# The following workaround can be removed once internal crypto
+# implementation is ported to use OpenSSL
+LDFLAGS += -lmbedx509
```
jxyang, line 52: Should we do the same for `target/sgx/enclave`?

mingweishih: No. This is only required for linking (when linking against `libmysttargetsgxenclave.a`).

defs.mak
mingweishih: updated.

jxyang (approved):

jxyang:
Please make sure tests still pass when MYST_OE_CRYPTO=mbedtls.

mingweishih:
Test for mbedtls is passing per #754

config.mak
```diff
@@ -4,6 +4,8 @@ MYST_PREFIX=/opt/mystikos
 #MYST_ENABLE_LEAK_CHECKER=1
 MYST_ENABLE_HOSTFS=1
 MYST_ENABLE_EXT2FS=1
+# openssl or mbedtls
+MYST_OE_CRYPTO=openssl
```
radhikaj, line 8: Use openssl_3 here for openssl

tools/myst/enc/Makefile
```diff
@@ -47,7 +47,19 @@ LIBS += $(LIBDIR)/libmystgcovclient.a
 endif
 
 LDFLAGS += $(OEENCLAVE_LDFLAGS)
-LDFLAGS += -lcrypto
+
+ifeq ($(MYST_OE_CRYPTO), openssl)
```
radhikaj, line 51: Ifeq myst_oe_crypto openssl or openssl_3

tools/myst/enc/Makefile
```diff
@@ -47,7 +47,19 @@ LIBS += $(LIBDIR)/libmystgcovclient.a
...
+# The following workaround can be removed once internal crypto
+# implementation is ported to use OpenSSL
+LDFLAGS += -lmbedx509
+LDFLAGS += -lmbedcrypto
+LDFLAGS += -loecore
+else ifeq ($(MYST_OE_CRYPTO), mbedtls)
+# Support OpenSSL-based sha256 implementation
+LDFLAGS += -lopensslcrypto
```
radhikaj, line 59: add LDFLAGS for openssl_3
Danny, help here?

defs.mak
```diff
@@ -94,7 +94,7 @@ OEENCLAVE_LIBDIR = $(BUILDDIR)/openenclave/lib/openenclave/enclave
 
 OEENCLAVE_LDFLAGS = $(shell PKG_CONFIG_PATH=$(BUILDDIR)/openenclave/share/pkgconfig pkg-config oeenclave-gcc --libs)
 
-OEENCLAVE_LDFLAGS += $(shell PKG_CONFIG_PATH=$(BUILDDIR)/openenclave/share/pkgconfig pkg-config oeenclave-gcc --variable=mbedtlslibs)
+OEENCLAVE_LDFLAGS += $(shell PKG_CONFIG_PATH=$(BUILDDIR)/openenclave/share/pkgconfig pkg-config oeenclave-gcc --variable=${MYST_OE_CRYPTO}libs)
 
 OEENCLAVE_CFLAGS = $(shell PKG_CONFIG_PATH=$(BUILDDIR)/openenclave/share/pkgconfig pkg-config oeenclave-gcc --cflags) -g
```
dcarpente, line 99: To build with OpenSSL 3.1, up above, we need to add:
```OEENCLAVE_INCLUDES += -I$(OE_INCDIR)/openenclave/3rdparty/openssl_3```
And this needs to be included _before_ the line:
```OEENCLAVE_INCLUDES += -I$(OE_INCDIR)/openenclave/3rdparty```

There's some information about this here:
https://github.com/openenclave/openenclave/blob/cdeb95c1ec163117de409295333b6b2702013e08/pkgconfig/README.md#building-enclave-applications
"OpenSSL 3.0 needs a different set of flags from cflags, because it uses additional include directories."

tools/myst/enc/Makefile
dcarpente: `-lopensslcrypto_3` instead of `-lopensslcrypto`

PR #744, opened 2021-09-01
Clear pid vector for "unmap_on_exit" memory

bodzhang:

This PR addresses the unexpected munmap() failure in myst_release_process_mappings().  Certain memory region owned by an app process gets unmapped at thread exit, but the existing code does not clear the corresponding pid vector values. As a result, when the app process exits, myst_release_process_mappings() attempts to unmap those region again, based on pid vector.


mikbras (approved):
Good catch!

PR #747, opened 2021-09-02
update azure sdk for cpp version

salsal97:


salsal97:
Successful runs here: https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10438&view=results

paulcallen (approved):

PR #748, opened 2021-09-02
updating aspnet to kill server based on running process data

salsal97:


mikbras (approved):

solutions/aspnet/Makefile
```diff
@@ -25,10 +25,11 @@ export COMPlus_EnableDiagnostics=0
 export COMPlus_GCHeapHardLimit=0x8000000
 
 run: rootfs private.pem
-	test -f server.pid && kill -9 `cat server.pid` || true
-	timeout $(TIMEOUT) $(MYST_EXEC) $(OPTS) rootfs /app/webapp > server.output & echo $$! > server.pid
+	./kill.sh
+	timeout $(TIMEOUT) $(MYST_EXEC) $(OPTS) rootfs /app/webapp > server.output &
```
jxyang, line 29: Do we still need server.output?

solutions/aspnet/kill.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+pid=$(ps -eaf | grep /app/webapp | grep -v grep | awk '{ print $2 }')
+if [ ! -z "${pid}" ]; then
+    sudo kill -9 ${pid}
```
jxyang, line 4: The previous approach using `server.pid` ensures the the same `webapp` instance we just launched get killed. This script actually would find the first instance of `webapp` and kill it. What's the reason to switch to the current one?

salsal97: This will actually find all instances and kill them, not just the first - its tighter on its approach and ensures that we are killing things that are actually running, and killing the right process

solutions/aspnet/Makefile
```diff
@@ -25,10 +25,11 @@ export COMPlus_EnableDiagnostics=0
 export COMPlus_GCHeapHardLimit=0x8000000
 
 run: rootfs private.pem
-	test -f server.pid && kill -9 `cat server.pid` || true
-	timeout $(TIMEOUT) $(MYST_EXEC) $(OPTS) rootfs /app/webapp > server.output & echo $$! > server.pid
```
jxyang, line 29: The reason we recorded the wrong server pid is because of the `timeout` command. server.pid actually contains the pid of the `timeout` command, not the `myst` command.

salsal97: Thats what I thought, but having the kill.sh script is what we've been using as a standard for most of such applications, so I added that here - two birds with one stone

jxyang (approved):

PR #753, opened 2021-09-07
Enable cpython3.9 test suite

bodzhang:
- Add crt/fork.c fix from jxyang to copy tsd and locale, addressing cpython3.9 GIL issue after fork();
- Enable both cpython3.9 and cpython3.8 tests suites in tests/cpython_tests; updated pass/fail list;
- In test/cpython_tests docker files, overwrites default shell of dash as bash, to workaround the issue that Mystikos currently does not support dash


vtikoo:
@bodzhang there are some timeout failures for both sgx and linux target - https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10593&view=logs&j=ab477ad6-9781-5b54-82ca-a0141b26726e

bodzhang:
[Nightly pipeline run](https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10721&view=results) passed

tests/cpython-tests/Dockerfile3.9
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:18.04
+WORKDIR /
+
+RUN apt update && apt install -y build-essential libssl-dev zlib1g-dev libncurses5-dev libncursesw5-dev \
+        libreadline-dev libsqlite3-dev libgdbm-dev libdb5.3-dev libbz2-dev \
+        libexpat1-dev liblzma-dev libffi-dev git
+
+RUN git clone -b v3.9.7 https://github.com/python/cpython
```
vtikoo, line 8: If only the version string is different, would recommend using the same dockerfile and adding version as a build time argument - https://docs.docker.com/engine/reference/builder/#arg

vtikoo (commented):
LGTM.

vtikoo (approved):

PR #756, opened 2021-09-07
Create init and build Makefile steps

rs--:
Introduced `init`,`build`, and `release-build` makefile steps. The goal is to decouple the initialization of prerequisites and building of Mystikos.

Added a makefile to the `ltp` directory; it seemed to be a "hanging submodule" when submodule references were removed.

### DIFF
--- a/third_party/Makefile
+++ b/third_party/Makefile
@@ -15,3 +15,9 @@ clean:
 
 distclean:
 	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) distclean $(NL) )
+
+init:
+	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) $(i) $(NL) )
+
+build:
+	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) build $(NL) )
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ TARBALL=$(PKGNAME).tar.gz
 
 all:
 	$(MAKE) .git/hooks/pre-commit
+	$(MAKE) init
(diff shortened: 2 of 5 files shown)


rules.mak
```diff
@@ -55,6 +55,11 @@ ifdef DIRS
 	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) $(NL) )
 endif
 
+init:
+ifdef DIRS
+	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) init $(NL) )
```
jxyang, line 60: Which external projects depend on this definition?

third_party/Makefile
```diff
@@ -15,3 +15,9 @@ clean:
 
 distclean:
 	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) distclean $(NL) )
+
+init:
+	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) $(i) $(NL) )
```
jxyang, line 20: Which 3rd-party projects depend on this definition?

rules.mak
rs--: `make init` called from the home directory will perform a recursive `make init` call in the `third_party` directory. This change isn't actually needed - I'll remove it

third_party/Makefile
rs--: All the projects in line 4: 

```
DIRS = openenclave musl mbedtls libc-test gcompat
```

It clones those 3rd party projects; subsequent calls won't do anything if the projects have already been cloned

jxyang (approved):

PR #757, opened 2021-09-08
Create Nginx sample

asvrada:
A simple nginx sample serving static content.

Certain configurations have to be used in order to avoid unsupported feature of Mystikos
1. Nginx needs to run with daemon off
2. Nginx needs to run with master_process off (however this option is only intended for nginx developers (http://nginx.org/en/docs/ngx_core_module.html#master_process))

Following error would occur if above configuration is not set:
1. If daemon set to on
`nginx: [emerg] fork() failed (95: Not supported)`

2. If master_process set to on
```
*** kernel panic: syscall.c(6339): _syscall(): unhandled syscall: SYS_rt_sigsuspend()
0xc3b7ca: __myst_panic()
0xc4b65e: _syscall()
0xc6a0f3: __morestack()
```

### DIFF
--- a/samples/nginx/Dockerfile
+++ b/samples/nginx/Dockerfile
@@ -1,0 +1,9 @@
+FROM nginx
+
+COPY ./nginx.conf /etc/nginx/nginx.conf
+
+COPY ./index.html /usr/share/nginx/html/index.html
+
+# nginx image by default runs with daemon off
+# Turn daemon off to avoid forking
+# CMD ["nginx", "-g", "daemon off;"]
--- a/samples/nginx/Makefile
+++ b/samples/nginx/Makefile
@@ -1,0 +1,18 @@
+TOP = $(abspath ../..)
+include $(TOP)/defs.mak
+
+APPBUILDER=$(TOP)/scripts/appbuilder
+
+all: appdir rootfs
+
+appdir:
+	$(APPBUILDER) Dockerfile
+
+rootfs: appdir
+	$(MYST) mkcpio appdir rootfs
+
+run:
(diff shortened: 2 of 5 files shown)


bodzhang:
@asvrada, does NGINX with master_process on have a fallback flow if it detects `SYS_rt_sigsuspend()` is not supported? Mystikos support `--unhandled-syscall-enosys true` command line option (and equivalent configuration entry in config.json) to return -ENOSYS on unsupported syscall instead of triggering `_myst_panic()` to crash the app.

asvrada:
I tried to run nginx with `master_process on` with UnhandledSyscallEnosys turned on, it failed. Nginx doesn't seem to have a fallback plan for this problem

```
myst exec-sgx  rootfs /usr/sbin/nginx --app-config-path config.json
2021/09/08 18:45:59 [notice] 101#101: using the "epoll" event method
2021/09/08 18:45:59 [notice] 101#101: nginx/1.21.1
2021/09/08 18:45:59 [notice] 101#101: built by gcc 8.3.0 (Debian 8.3.0-6) 
2021/09/08 18:45:59 [notice] 101#101: OS: Linux 5.4.0
2021/09/08 18:45:59 [notice] 101#101: getrlimit(RLIMIT_NOFILE): 65536:65536
2021/09/08 18:45:59 [notice] 101#101: start worker processes
2021/09/08 18:45:59 [alert] 101#101: ioctl(FIOASYNC) failed while spawning "worker process" (95: Not supported)
```

vtikoo:
Any reason to put this under samples instead of solutions? Putting it under solutions has the benefit of being part of the CI pipeline.

bodzhang (approved):

mingweishih:
Next time, please ensure both checks are passed before merging.

PR #760, opened 2021-09-09
Defer fdmappings cleanup from file descriptor close to munmap

vtikoo:
@bodzhang and I encountered following sequence of syscalls from different dotnet applications:

1. file opened for reading and mmapped
2. file is closed
3. msync is called in the mmap address range

Because [we don't cleanup fdmappings when read-only files are closed](https://github.com/deislabs/mystikos/blob/main/kernel/mmanutils.c#L1000), msync finds an "in-use" fdmapping, but with a closed fd.
This causes msync to fail with an `EPERM`.

#### Summary of changes:

- fdmappings cleanup at file descriptor close time is deferred to munmap. This allows functionalities relying on fdmappings vector - /proc/[pid]/maps and msync, to function.
- Added new state to fdmapping `used` field - MYST_FDMAPPING_FD_CLOSED. A fdmapping is in this state between the time close is called on a descriptor associated with a file mapping, and munmap is called on the file mapping.

### DIFF
--- a/kernel/mmanutils.c
+++ b/kernel/mmanutils.c
@@ -984,57 +984,46 @@ done:
 /* notified on close to clear fd-mappings involving fd */
 void myst_mman_close_notify(int fd)
 {
-    int flags;
     struct stat buf;
     bool locked = false;
-
-    /* get the file open flags */
-    if (fd < 0 || (flags = myst_syscall_fcntl(fd, F_GETFL, 0)) < 0)
-        return;
 
     /* only do this for regular files */
     if (myst_syscall_fstat(fd, &buf) != 0 || !S_ISREG(buf.st_mode))
         return;
 
-    /* if file is open for write */
-    if (flags & (O_RDWR | O_WRONLY))
+    _rlock(&locked);
+    uint8_t* addr = (uint8_t*)_mman.map;
+    size_t length = ((uint8_t*)_mman.end) - addr;
+    _runlock(&locked);
+
+    vectors_t v = _get_vectors();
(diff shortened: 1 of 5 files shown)


kernel/mmanutils.c
```diff
@@ -984,57 +984,46 @@ int myst_msync(void* addr, size_t length, int flags)
 /* notified on close to clear fd-mappings involving fd */
 void myst_mman_close_notify(int fd)
 {
-    int flags;
     struct stat buf;
     bool locked = false;
 
-    /* get the file open flags */
-    if (fd < 0 || (flags = myst_syscall_fcntl(fd, F_GETFL, 0)) < 0)
```
mingweishih, line 992: Do we need to retain the fd < 0 check?

mingweishih (commented):
Can we have a unit test that tests this specific behavior?

vtikoo:
dotnet tests pipeline run [here](https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10592&view=logs&j=ab477ad6-9781-5b54-82ca-a0141b26726e).

kernel/mmanutils.c
vtikoo: The sole user of this function `myst_syscall_close`, does check for a valid fd. But added the check back to be defensive.

vtikoo:
> Can we have a unit test that tests this specific behavior?

Added test mirroring dotnet's usage.

vtikoo:
Overruled by #770

PR #761, opened 2021-09-09
Return EINVAL on setrlimit, silent for NOFILE and STACK

Sahakait:
Originally, `setrlimit` was only silent for `NOFILE` to allow memacached solution to pass. A failure in the glibc test case was discovered due to `ENOTSUP` from setting `STACK`. Added `STACK` to silent and changed error to `EINVAL` (allowed by specs).

### DIFF
--- a/kernel/limit.c
+++ b/kernel/limit.c
@@ -152,7 +152,8 @@ int myst_limit_set_rlimit(pid_t pid, int resource, struct rlimit* rlim)
     // ATTN: Setting rlimit value is currently unsupported
     // Returning a failure will break solutions/memcached (RLIMIT_NOFILE)
     // and tests/glibc (RLIMIT_STACK)
-    // ERAISE(-ENOTSUP);
+    if (resource != RLIMIT_NOFILE && resource != RLIMIT_STACK)
+        ERAISE(-EINVAL);
 done:
     return ret;
 }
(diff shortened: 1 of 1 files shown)


mingweishih (commented):
Can you ensure the nightly build passes with the PR? Also, can you update the tests.passed/tests.failed list accordingly?

vtikoo:
Can we just allow NOFILE and STACK to silently pass? With #696, the process stack size can be configured with a cmdline option. Once thats in, with some additional work it should be possible to support setrlimit for RLIMIT_STACK.

vtikoo:
For setrlimit issues with other resources, I wonder if its better to triage the issue as and when it happens.

Sahakait:
> Can we just allow NOFILE and STACK to silently pass? With #696, the process stack size can be configured with a cmdline option. Once thats in, with some additional work it should be possible to support setrlimit for RLIMIT_STACK.

I'm concerned about similar scenarios from other calls to setrlimit, since ENOTSUP is not actually specified as a possible error in the man page.

Sahakait:
> Can you ensure the nightly build passes with the PR? Also, can you update the tests.passed/tests.failed list accordingly?

Running [here](https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10583&view=results).

vtikoo (approved):
Looks good. Pls ensure nightly pipeline passes before merging.

mingweishih:
I didn't see any update on the pass/fail list. It's unclear what's the affecting test cases that this PR tries to fix.

PR #765, opened 2021-09-10
Add support for the get_mempolicy syscall ocall

salsal97:


kernel/syscall.c
```diff
@@ -3169,6 +3169,20 @@ long myst_syscall_mbind(
...
 }
 
+long myst_syscall_get_mempolicy(
+    int* mode,
+    unsigned long* nodemask,
+    unsigned long maxnode,
+    void* addr,
+    unsigned long flags)
+{
+    long params[6] = {
```
bodzhang, line 3179: From Linux manual:

If nodemask is not NULL, then the nodemask associated with the policy will be
stored in the location pointed to by this argument. maxnode specifies the number
of node IDs that can be stored into nodemask—that is, the maximum node ID plus
one. The value specified by maxnode is always rounded to a multiple of
sizeof(unsigned long)*8.

The app might provide a maxnode value that's not a multiple of `sizeof(unsigned long)*8`. Round it up according to the Linux manual before calling into the target will simplify OCALL implementation.

kernel/syscall.c
```diff
@@ -3169,6 +3169,20 @@ long myst_syscall_mbind(
...
+        (long)mode, (long)nodemask, (long)maxnode, (long)addr, (long)flags};
+    long ret = myst_tcall(SYS_get_mempolicy, params);
+
```
bodzhang, line 3182: From Linux Manual:

       If flags specifies MPOL_F_MEMS_ALLOWED (available since Linux
       2.6.24), the mode argument is ignored and the set of nodes
       (memories) that the thread is allowed to specify in subsequent
       calls to mbind(2) or set_mempolicy(2) (in the absence of any mode
       flags) is returned in nodemask.

As Mystikos does not support mbind or set_mempolicy, it seems that the get_mempolicy() implementation should override nodemask returned by the target to indicate the thread is not allow to specify policy on any node in mbind or set_mempolicy.

bodzhang (changes requested):

kernel/syscall.c
salsal97: Thank you for your suggestion, I have rounded up the value here

kernel/syscall.c
salsal97: Added the overriding - this creates an assert failure in the `make native` target though that I have left there with a comment of explanation, since both are right in what they do

tools/myst/myst.edl
```diff
@@ -513,5 +513,13 @@ enclave
             [out] uint32_t* rbx,
             [out] uint32_t* rcx,
             [out] uint32_t* rdx);
+        
+        long myst_get_mempolicy_ocall(
+            [out] int* mode,
+            [out] unsigned long* nodemask,
+            unsigned long maxnode,
+            [out] void* addr,
```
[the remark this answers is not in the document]
salsal97: Changed the configuration in the edl, this also needed me to either 
i)nodemask/=8 which is what I did, or 
ii) introduce a new parameter in the ocall which I wanted to avoid to keep it in line with the prototype in the linux manpage

let me know if you think option ii) is better in order to pass the size to the ocall

kernel/syscall.c
```diff
@@ -3169,6 +3169,33 @@ long myst_syscall_mbind(
...
+        (long)mode, (long)nodemask, (long)size, (long)addr, (long)flags};
+    long ret = myst_tcall(SYS_get_mempolicy, params);
+
+    // As Mystikos does not support mbind or set_mempolicy, we override nodemask
+    // returned by the target to indicate the thread is not allowed to specify a
+    // policy on any node in mbind or set_mempolicy when
+    // flags = MPOL_F_MEMS_ALLOWED
+    if (flags == 1 << 2)
+    {
+        *nodemask = 0;
```
bodzhang, line 3193: if `maxnode` > 64, more than one unsigned long element pointed to by `nodemask` need to be zeroed.

tools/myst/myst.edl
```diff
@@ -513,5 +513,13 @@ enclave
             [out] uint32_t* rbx,
             [out] uint32_t* rcx,
             [out] uint32_t* rdx);
+        
+        long myst_get_mempolicy_ocall(
+            [out] int* mode,
+            [out, size=maxnode] unsigned long* nodemask,
```
bodzhang, line 519: `maxnode` indicates the num of bits within `nodemask` buffer. EDL syntax `size` is probably in bytes. Might need to convert `maxnode` in bits to a different parameter, for example, number of unsigned long elements in the `nodemask` buffer.

tools/myst/myst.edl
```diff
@@ -513,5 +513,13 @@ enclave
...
+            unsigned long maxnode,
+            [in] void* addr,
```
bodzhang, line 521: EDL `in` syntax instruct the OCALL framework to copy a buffer referenced by a pointer from inside the enclave to outside of the enclave, so the outside enclave code can deference the pointer to the buffer (outside of the enlcave) and access the buffer. In get_mempolicy syscall, `addr` pointer will not be dereferenced. The eDL should not have `[in]`

tools/myst/myst.edl
salsal97: I added a line maxnode/=8 to go around this in the caller function for this, but I changed that approach to add a different variable in my update, thanks

tools/myst/myst.edl
salsal97: Got it, updated this. Will out be accurate then in this case? Earlier comment says `should not have [out]`

kernel/syscall.c
salsal97: updated implementation

tools/myst/myst.edl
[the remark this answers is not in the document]
mingweishih: As @bodzhang pointed out, the `addr` seems to be an input (not sure if there is in/out case, please double check). Also, the syscall only uses the `addr` as an address value instead of a pointer to a memory buffer, you can simply remove the `[]` annotation.

kernel/syscall.c
```diff
@@ -3126,6 +3126,37 @@ long myst_syscall_mbind(
...
+    int* mode,
+    unsigned long* nodemask,
+    unsigned long maxnode,
+    void* addr,
+    unsigned long flags)
+{
+    unsigned long size = 0;
+    int ROUND_UP_VAL = sizeof(unsigned long) * 8;
+    myst_round_up(maxnode, ROUND_UP_VAL, &size);
+
```
mingweishih, line 3139: Suggest adding some input validation (the case that returns `EINVAL` as follows):

>        EINVAL The value specified by maxnode is less than the number of
>              node IDs supported by the system.  Or flags specified
>              values other than MPOL_F_NODE or MPOL_F_ADDR; or flags
>              specified MPOL_F_ADDR and addr is NULL, or flags did not
>              specify MPOL_F_ADDR and addr is not NULL.  Or, flags
>              specified MPOL_F_NODE but not MPOL_F_ADDR and the current
>              thread policy is not MPOL_INTERLEAVE.  Or, flags specified
>              MPOL_F_MEMS_ALLOWED with either MPOL_F_ADDR or
>              MPOL_F_NODE.  (And there are other EINVAL cases.)

tests/getmempolicy/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+native:
+	gcc $(CFLAGS) getmempolicy.c $(LDFLAGS) -L. -lnuma
+	./a.out
+
+myst:
+	$(MAKE) -C $(TOP)/tools/myst
+
+clean:
+	rm -rf $(APPDIR) rootfs export ramfs
```
mingweishih, line 36: nit: new line

tests/getmempolicy/getmempolicy.c
```diff
@@ -0,0 +1,104 @@
...
+    assert(maxnode == 64);
+    assert(addr == 0);
+}
+
+int main(int argc, const char* argv[])
+{
+    test1();
+
+    test2();
+
```
mingweishih, line 101: Can you also add following cases
- `addr` is not zero
- larger `nodemask` (can be allocated on heap)

kernel/syscall.c
mingweishih: Thinking out load --- given we're rounding up the size, I'm wondering would there be a case of `nodemask` being overflowed. This may also apply to the behavior of the syscall itself --- how does the syscall behave if the user is passing unrounded `maxnode`?  Maybe this have been guarded by the memory alignment? Can we confirm this behavior?

tools/myst/myst.edl
```diff
@@ -513,5 +513,14 @@ enclave
             [out] uint32_t* rbx,
             [out] uint32_t* rcx,
             [out] uint32_t* rdx);
+        
+        long myst_get_mempolicy_ocall(
+            [out] int* mode,
+            [out, size=maxnode_byte] unsigned long* nodemask,
+            unsigned long maxnode,
+            size_t maxnode_byte,
```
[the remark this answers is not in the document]
salsal97: Thanks, that does make it easier to read

tools/myst/enc/syscall.c
```diff
@@ -900,9 +900,9 @@ static long _get_mempolicy(
     unsigned long flags)
 {
     long ret;
-    size_t maxnode_byte = maxnode / 8;
+    size_t nodemask_size = maxnode / 8;
```
bodzhang, line 903: Using sizeof(...) instead of hardcoding of 8 would be better

tools/myst/host/syscall.c
```diff
@@ -524,8 +524,8 @@ long myst_fsync_ocall(int fd)
 long myst_get_mempolicy_ocall(
     int* mode,
     unsigned long* nodemask,
+    size_t nodemask_size,
```
bodzhang, line 527: `maxnode` parameter one line 532 can be derived from `nodemask_size`, no need to include 'maxnode` as an OCALL parameter.

kernel/syscall.c
salsal97: Thank you for helping me with this case ming-wei, have added our added testcases here

kernel/syscall.c
salsal97: Added more testcases, thank you for pointing this out

tests/getmempolicy/getmempolicy.c
salsal97: Added

tools/myst/enc/syscall.c
salsal97: using CHAR_BIT here since 8 is the translation of bits to byte, and is used as `8` in the spec as well

tools/myst/host/syscall.c
salsal97: since nodemask_size is only needed for the myst.edl, I like the idea of keeping the parameters separate from each other. In the event I take out maxnode in the OCALL implementation, there would be double computation of first calculating nodemask_size from maxnode (maxnode/8) and then calculating maxnode from nodemask size, which seems unideal.

tests/getmempolicy/Makefile
salsal97: added, thanks

kernel/syscall.c
```diff
@@ -3126,6 +3126,57 @@ long myst_syscall_mbind(
...
+    void* addr,
+    unsigned long flags)
+{
+    unsigned long rounded_maxnode = 0;
+    int ROUND_UP_VAL = sizeof(unsigned long) * 8;
+    myst_round_up(maxnode, ROUND_UP_VAL, &rounded_maxnode);
+
+    unsigned long* rounded_nodemask;
+    // allocate higher buffer
+    if (rounded_maxnode > maxnode)
```
bodzhang, line 3142: `nodemask` buffer size is supposed to be big enough to hold `maxnode` of bits. The API does not provide any parameter to specify the size of the `nodemask` buffer. Therefore the Linux manual clarified that the implementation of API can round up `maxnode` to the next multiple of `sizeof(unsigned long) * 8` (bits in one element of the `nodemask` buffer), and use it to derive the buffer size. No need to allocate an internal buffer if `maxnode` provided is not a multiple of `sizeof(unsigned long) * 8`

kernel/syscall.c
```diff
@@ -3126,6 +3126,57 @@ long myst_syscall_mbind(
...
+                      (long)rounded_maxnode,
+                      (long)addr,
+                      (long)flags};
+    long ret = myst_tcall(SYS_get_mempolicy, params);
+
+    // As Mystikos does not support mbind or set_mempolicy, we override nodemask
+    // returned by the target to indicate the thread is not allowed to specify a
+    // policy on any node in mbind or set_mempolicy when
+    // flags = MPOL_F_MEMS_ALLOWED
+    if (flags == 1 << 2)
```
bodzhang, line 3162: If `ret` indicate error, skip any operation on nodemask.

tools/myst/host/syscall.c
```diff
@@ -520,3 +520,14 @@ long myst_fsync_ocall(int fd)
...
+
+long myst_get_mempolicy_ocall(
+    int* mode,
+    unsigned long* nodemask,
+    size_t nodemask_size,
+    unsigned long maxnode,
+    void* addr,
+    unsigned long flags)
+{
+    RETURN(syscall(SYS_get_mempolicy, mode, nodemask, maxnode, addr, flags));
```
bodzhang, line 532: OE outcall implementation does NOT zero-set [OUT] parameter. Does Linux syscall implementation zero out nodemask before setting relevant bits? If not, the returned nodemask might be a mix of SYSCALL set bits and the relic of nodemask allocated by the OE OCALL infrastructure code. Setting nodemask as [IN, OUT] in the EDL will guarantee that nodemask is zeroed before the Linux syscall.

kernel/syscall.c
```diff
@@ -3126,6 +3126,57 @@ long myst_syscall_mbind(
...
+    // allocate higher buffer
+    if (rounded_maxnode > maxnode)
+    {
+        rounded_nodemask = malloc(rounded_maxnode / CHAR_BIT);
+    }
+    else
+    {
+        rounded_nodemask = nodemask;
+    }
+
```
bodzhang, line 3150: zero out `nodemask` before calling lower level function, in case the app code didn't do that.

tests/getmempolicy/getmempolicy.c
```diff
@@ -0,0 +1,191 @@
...
+    printf("1. MPOL_DEFAULT, null addr");
+    print(&mode, &nodemask, maxnode, addr, flags);
+    long ret = get_mempolicy(&mode, &nodemask, maxnode, addr, flags);
+    print(&mode, &nodemask, maxnode, addr, flags);
+    printf("\nRet = %ld", ret);
+    printf("\nerrno = %d", errno);
+
+    assert(ret == 0);
+    assert(mode == 0);
```
bodzhang, line 71: default policy on Linux target might not be 0

tests/getmempolicy/getmempolicy.c
```diff
@@ -0,0 +1,191 @@
...
+    assert(nodemask == 0);
```
bodzhang, line 72: nodemask associated with default policy on Linux target might not be 0

tests/getmempolicy/getmempolicy.c
```diff
@@ -0,0 +1,191 @@
...
+    assert(maxnode == 64);
```
bodzhang, line 73: The check has no effect, as maxnode is an input set to 64. Same comments on other assert code on input variables.

tests/getmempolicy/getmempolicy.c
```diff
@@ -0,0 +1,191 @@
...
+
+int test4()
+{
+    int mode = 5;
+    unsigned long nodemask = 0xffffffffffffffff;
+    unsigned long maxnode = 64;
+    void* addr = NULL;
+    unsigned long flags = MPOL_F_ADDR;
+
+    printf("\n\n4. MPOL_F_ADDR, NULL addr, expecting EFAULT");
```
bodzhang, line 134: Should also test valid addr case. If addr points to a buffer allocated on heap or stack, it's part of the Mystikos app EPC.

tools/myst/host/syscall.c
salsal97: Added comment & void statement

kernel/syscall.c
salsal97: @mingweishih and I actually tried this case out, and got a segmentation fault in the case where we pass a non multiple, since we are rounding the maxnode value before passing it to the ocall. If we are passing it a higher value, it seems to also be expecting a larger buffer. My guess is if we remove the rounding we also remove the need for a larger buffer, since it would all be taken care of by the ocall/linux call itself

kernel/syscall.c
salsal97: Good catch, adding in next commit

tests/getmempolicy/getmempolicy.c
salsal97: Works on usual + icelake VM and works for both targets, is there anything I should change here to make it future proof?

radhikaj:

PR #767, opened 2021-09-10
Removing broken sub-module (third_party/ltp/ltp)

mrragava:
Removing broken sub-module (third_party/ltp/ltp) and Adding /dev/isgx in sgx driver installation check

Signed-off-by: Ragavan Dasarathan <mrragava@gmail.com>


mingweishih:
The removal of ltp submodule is part of #756. Please double check.

PR #768, opened 2021-09-10
EPOLL as a first-class feature

mikbras:
This PR implements ``epoll`` as a first-class feature. Formerly, it was implemented on top of ``poll``, which does not support edge-triggered events. Now ``epoll`` is implemented directly with ocalls. For this to work **pipes** and **eventfds** were reimplemented to utilize host-side events rather than kernel-side events (implementing kernel-side edge-triggering for these would have been very difficult to get right).

The **pipe** implementation employs a host-side pipe merely for synchronization and a kernel-side buffer for exchanging data. The host-side pipe can be added to the host-side ``epoll`` set. The pipe implementation manages the host-side pipe so that it has three states: write-enabled (empty), read-write-enabled (half-full), and read-enabled (full). The algorithm is described [here](https://github.com/deislabs/mystikos/blob/f32c23f55bb0d9a00342c4fb042730e35eed4ee8/kernel/pipedev.c#L42).

This PR also makes ``eventfd`` a host-side object and eliminates the concept of kernel-side events.


mikbras:
https://github.com/deislabs/mystikos/blob/f32c23f55bb0d9a00342c4fb042730e35eed4ee8/kernel/pipedev.c#L39

tools/myst/enc/syscall.c
```diff
@@ -1637,6 +1603,132 @@ static long _fsync(int fd)
...
+
+    if (myst_epoll_wait_ocall(&retval, epfd, events, maxevents, timeout) !=
+        OE_OK)
+    {
+        ret = -EINVAL;
+        goto done;
+    }
+
+    /* guard against possible read past end of buffer */
+    if (retval > maxevents)
```
bodzhang, line 1669: Encountered a failure here with a .NET coreclr test case. Looks like the compiler converted int retval to size_t before the comparison and caused negative retval, such a -EINTR, to trigger the error condition.

```suggestion
    if ((retval > 0) && (retval > maxevents))
```

mikbras: Thank you Bo for finding this. The comparison between `retval` and `maxevents` should have reported a conversion error but unfortunately that compiler option is not enabled. We need to enable checks for conversions across Mystikos but for now, I have applied the following fix.

```C++
    /* check for error */
    if (retval < 0)
    {
        ret = retval;
        goto done;
    }

    /* guard against possible read past end of buffer */
    if ((size_t)retval > maxevents)
    {
        ret = -EINVAL;
        goto done;
    }
```

paulcallen (approved):

PR #770, opened 2021-09-11
Use duplicated fd for file mappings

vtikoo:
### Summary
- For file mappings, mman duplicates the fd and uses that for msync operations.
- Mechanism for notifying mman of fd close - myst_mman_close_notify, is removed. As mman now has its own copy of the fd.
- munmap and myst_release_process_mappings now need to do additional cleanup of closing duplicated file handles. This is done in a 2 pass manner. In the first pass, the list of fd's to be closed is collected. And once outside the mman lock, the close operation on the fd list is performed. This is to avoid deadlock situation between the mman lock and the file system lock.

### DIFF
--- a/hostfs/hostfs.c
+++ b/hostfs/hostfs.c
@@ -910,7 +910,6 @@ static int _fs_dup(
     ECHECK((tret = myst_tcall(SYS_dup, params)));
 
     new_file->fd = tret;
-    ret = tret;
 
     *file_out = new_file;
     new_file = NULL;
--- a/kernel/mmanutils.c
+++ b/kernel/mmanutils.c
@@ -527,13 +527,30 @@ int myst_mprotect(const void* addr, const size_t len, const int prot)
 }
 
 /* release msync mappings that are contained in the range [addr:addr+length] */
-static int _remove_file_mappings(void* addr, size_t length)
+typedef struct fdlist
+{
+    int fd;
+    struct fdlist* next;
+} fdlist_t;
+
+fdlist_t* get_tail(fdlist_t* node)
+{
+    fdlist_t* prev = NULL;
+    while (node)
+    {
+        prev = node;
+        node = node->next;
+    }
+    return prev;
+}
+
+static int _remove_file_mappings(void* addr, size_t length, fdlist_t** head_out)
 {
     int ret = 0;
     size_t index;
     vectors_t v = _get_vectors();
     bool locked = false;
-
+    fdlist_t* head = NULL;
     if (!addr || !length)
         ERAISE(-EINVAL);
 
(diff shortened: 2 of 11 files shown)


hostfs/hostfs.c
```diff
@@ -910,7 +910,6 @@ static int _fs_dup(
     ECHECK((tret = myst_tcall(SYS_dup, params)));
 
     new_file->fd = tret;
-    ret = tret;
```
vtikoo, line 913: ramfs and ext2fs `fs_dup` returns 0 on success. The fd allocation and assignment is done in `myst_syscall_dup`.

kernel/mmanutils.c
```diff
@@ -538,12 +561,31 @@ static int _remove_file_mappings(void* addr, size_t length)
...
             /* remove any fd-mapping (it is okay if it does exist) */
+            myst_fdmapping_t* p = &v.fdmappings[i];
+
+            /* add fd's for in-use mappings to a list. add only for the first
+             * page of
+             * interval mapped with same fd. The fd list will be closed by the
+             * caller outside the mman lock. */
+            if (p->used == MYST_FDMAPPING_USED && p->fd != prev_cleared_fd)
+            {
+                fdlist_t* fd_node = calloc(1, sizeof(fdlist_t));
```
paulcallen, line 577: possible memory allocation failure Make sure the unlock happen!

kernel/mmanutils.c
```diff
@@ -553,15 +595,37 @@ static int _remove_file_mappings(void* addr, size_t length)
     }
     _runlock(&locked);
 
+    *head_out = head;
+    head = NULL;
+
 done:
-    _runlock(&locked);
+
+    if (head)
```
paulcallen, line 603: the if is not needed because the while loop takes care of this

kernel/mmanutils.c
```diff
@@ -570,9 +634,25 @@ int myst_munmap(void* addr, size_t length)
...
+    _runlock(&locked);
+    return ret;
+}
+
+int myst_munmap(void* addr, size_t length)
+{
+    int ret = 0;
+    fdlist_t* head = NULL;
+
+    ECHECK(__myst_munmap(addr, length, &head));
```
paulcallen, line 652: any reason the file handles arent returned from the remove_file_mappings instead? At least the handles are related to that, right?
Not a big deal, but could possibly make it more obvious.

paulcallen (approved):

kernel/mmanutils.c
vtikoo: Good point, fixed.

kernel/mmanutils.c
vtikoo: The file handles are returned by remove_file_mappings, but I added a `__myst_munmap` wrapper so that the munmap and clearing of fdmappings happens atomically.

radhikaj (approved):

PR #781, opened 2021-09-15
Fallback to use pathname from fdmappings entry for msync

vtikoo:
### Summary
- fdmappings are not cleaned up when fd is closed. This enables msync operations after file close.
- on `close(fd)`, the fd field in the corresponding file mapping entries is invalidated. Any later msync uses `fdmapping.pathname` to open the file. 

### DIFF
--- a/kernel/mmanutils.c
+++ b/kernel/mmanutils.c
@@ -964,31 +964,41 @@ int myst_msync(void* addr, size_t length, int flags)
         bool consistent;
         const size_t n = length / PAGE_SIZE;
         uint8_t* page = addr;
-        int prev_fd = -1;
+        char* prev_pathname = NULL;
 
         for (size_t i = index; i < index + n; i++)
         {
             myst_fdmapping_t* p = &v.fdmappings[i];
-
-            if (p->used == MYST_FDMAPPING_USED)
+            /* sync only for writable mappings */
+            ECHECK(myst_mman_get_prot(
+                &_mman, page, PAGE_SIZE, &prot, &consistent));
+            if (prot & PROT_WRITE && p->used == MYST_FDMAPPING_USED)
             {
-                /* sync only for writable mappings */
-                ECHECK(myst_mman_get_prot(
-                    &_mman, page, PAGE_SIZE, &prot, &consistent));
-
-                if (prot & PROT_WRITE)
+                if (p->fd != -1) // valid fd
                 {
-                    /* fetch file handle via pathname if new fd encountered */
-                    if (prev_fd != p->fd)
+                    // file may have been truncated, ignore failure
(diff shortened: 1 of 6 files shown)


kernel/mmanutils.c
```diff
@@ -960,15 +960,46 @@ int myst_msync(void* addr, size_t length, int flags)
...
 
         for (size_t i = index; i < index + n; i++)
         {
             myst_fdmapping_t* p = &v.fdmappings[i];
-
-            if (p->used == MYST_FDMAPPING_USED)
-                ECHECK(_sync_file(p->fd, p->offset, page, PAGE_SIZE));
+            /* sync only for writable mappings */
+            ECHECK(myst_mman_get_prot(
+                &_mman, page, PAGE_SIZE, &prot, &consistent));
```
mingweishih, line 974: `consistent` is not used. Do we need a check for it as well?

kernel/mmanutils.c
```diff
@@ -960,15 +960,46 @@ int myst_msync(void* addr, size_t length, int flags)
...
+                if (p->fd != -1) // valid fd
+                {
+                    // file may have been truncated, ignore failure
+                    _sync_file(p->fd, p->offset, page, PAGE_SIZE);
+                }
+                // invalidated fd, fall back on pathname
+                else
+                {
+                    /* Get file handle the first time, and when the pathname
+                     * field changes*/
```
mingweishih, line 986:
```suggestion
                     * field changes */
```

kernel/mmanutils.c
```diff
@@ -960,15 +960,46 @@ int myst_msync(void* addr, size_t length, int flags)
...
+                    if (!prev_pathname ||
+                        !strcmp(prev_pathname, p->pathname->data))
```
mingweishih, line 988: Does this mean that we only sync on mappings that have the same pathname as the first entry? Just to help me understand, can you explain a bit more on this?

kernel/mmanutils.c
vtikoo: Protection is being checked for a single page here, so it should always be consistent.

mingweishih: Ah, I see. In this case, you can simply pass NULL instead of `&consistent`

kernel/mmanutils.c
vtikoo: The sync happens always, this if block checks whether we need to get a new handle. We fetch a file handle once for the mapping interval which belongs to the same pathname. I have expanded the comment as well.

kernel/mmanutils.c
vtikoo: `myst_mman_get_prot` does not allow a nullable `consistent` argument currently.

vtikoo:
Overruled by #770

Issue #534, opened 2021-07-02
LTP test: getsockname01 fails 


salsal97:
Creating this issue for tracking this syscall since it is widely used

```
$ make one TEST=/ltp/testcases/kernel/syscalls/getsockname/getsockname01
sudo -E /home/sagoel/multitask/mystikos/build/bin/myst exec-sgx ext2fs /ltp/testcases/kernel/syscalls/getsockname/getsockname01
getsockname01 1 TPASS : bad file descriptor successful
getsockname01 2 TFAIL : getsockname01.c:127: bad file descriptor ; returned -1 (expected -1), errno 9 (expected 88)
getsockname01 3 TFAIL : getsockname01.c:127: invalid socket buffer ; returned -1 (expected -1), errno 22 (expected 14)
getsockname01 4 TFAIL : getsockname01.c:127: invalid aligned salen ; returned -1 (expected -1), errno 22 (expected 14)
Makefile:183: recipe for target 'one' failed
make: *** [one] Error 255
```

vtikoo:
It looks like we are not returning the right errno in the failing cases.

radhikaj:
Closing since #783 reenabled this test. @salsal97 , please reopen this test if I am mistaken.

PR #783, opened 2021-09-16
fix errno values and enable getsockname ltp test

salsal97:


salsal97:
This PR fixes https://github.com/deislabs/mystikos/issues/534

include/myst/kernel.h
```diff
@@ -226,7 +226,7 @@ void myst_start_shell(const char* msg);
 
 const char* myst_syscall_str(long n);
 
-MYST_INLINE bool myst_is_bad_addr(const void* ptr)
+MYST_INLINE bool myst_is_good_addr(const void* ptr)
```
mingweishih, line 229: I have similar fix in #724. Suggest a better naming
`myst_is_within_kernel`
or
`myst_is_within_enclave` (to avoid the confusion that user memory is also within this range)

kernel/syscall.c
```diff
@@ -2671,8 +2671,26 @@ long myst_syscall_getsockname(
...
+    void* device = NULL;
+    void* object = NULL;
+
+    if (!myst_is_good_addr(addrlen) || !myst_is_good_addr(addr))
+    {
+        ret = -EFAULT;
+        goto done;
+    }
+
+    ECHECK(myst_fdtable_get_any(fdtable, sockfd, &type, &device, &object));
```
mingweishih, line 2684: I think it's fine to just pass `sd` and `sock` as-is. Do you see any problem with it?

kernel/syscall.c
```diff
@@ -2671,8 +2671,26 @@ long myst_syscall_getsockname(
     myst_fdtable_t* fdtable = myst_fdtable_current();
     myst_sockdev_t* sd;
     myst_sock_t* sock;
+    myst_fdtable_type_t type;
+    void* device = NULL;
+    void* object = NULL;
+
+    if (!myst_is_good_addr(addrlen) || !myst_is_good_addr(addr))
+    {
+        ret = -EFAULT;
```
mingweishih, line 2680: use ERAISE(..)

kernel/syscall.c
```diff
@@ -2671,8 +2671,26 @@ long myst_syscall_getsockname(
...
+        goto done;
+    }
+
+    ECHECK(myst_fdtable_get_any(fdtable, sockfd, &type, &device, &object));
+    sd = (myst_sockdev_t*)device;
+    sock = (myst_sock_t*)object;
+
+    if (type != MYST_FDTABLE_TYPE_SOCK)
+    {
+        ret = -ENOTSOCK;
```
mingweishih, line 2690: use ERAISE(..)

kernel/syscall.c
salsal97: yeah - it expects void**, and this is how other syscalls in this file were structured

mingweishih: Given that you will casting them back anyway, you can cast them with (void**) here.
See https://github.com/deislabs/mystikos/blob/beba4ea27622a64d1698dfd1c9d997f391c59dd5/kernel/syscall.c#L2530

include/myst/kernel.h
mikbras: This check is more restrictive than just "within-enclave" so I like your first suggestion: `myst_is_within_kernel`.

mikbras: Also, the kernel should not know about the concept of an enclave. :)

kernel/syscall.c
mikbras, line 2680: I'd replace these two lines with:

```C++
ERAISE(-EFAULT);
```

mikbras: Oh, Ming-Wei already commented on that. :)

kernel/syscall.c
mikbras: I agree that you can just cast them an not have the temporaries.

mikbras:
Other than Ming-Wei's suggestion, it looks great.

kernel/syscall.c
salsal97: using ERAISE causes the tests to all fail - is there a way to do this other than ERAISE(EFAULT);?
```
getsockname01    1  TPASS  :  bad file descriptor successful
getsockname01    2  TFAIL  :  getsockname01.c:127: bad file descriptor ; returned -1 (expected -1), errno 2 (expected 88)
getsockname01    3  TFAIL  :  getsockname01.c:127: invalid socket buffer ; returned -1 (expected -1), errno 22 (expected 14)
getsockname01    4  TFAIL  :  getsockname01.c:127: invalid aligned salen ; returned -1 (expected -1), errno 22 (expected 14)
```

salsal97: Added, thank you

kernel/syscall.c
salsal97: Added, thank you

kernel/syscall.c
salsal97: Added, thank you!

kernel/syscall.c
salsal97: Ah, I was trying &((void*)sd) which failed and I see why, this makes more sense. Thank you!

include/myst/kernel.h
salsal97: Added this, thank you!

salsal97:
Successful execution of this test https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10734&view=logs&j=ab477ad6-9781-5b54-82ca-a0141b26726e&t=a1e294d4-8895-5a45-41ff-022cede0621b&l=5394

mingweishih (approved):

PR #785, opened 2021-09-16
adding native targets to ltp to aid debugging

salsal97:


tests/ltp/Makefile
```diff
@@ -66,6 +66,24 @@ build:
...
+	cp build.mak native_appdir/Makefile
+	$(MAKE) -C native_appdir native-all
+	$(MAKE) password-file
+	$(MAKE) group-file
+	$(MAKE) add_host_users_and_groups
+
+native-build-one: native-clone
+	cp -p build.mak native_appdir/Makefile
+	$(MAKE) -C native_appdir native-one TEST=$(TEST)
+	sudo chown -R $(USER).$(USER) appdir
```
radhikaj, line 82: Why is this not required for the native-build case but required for the native-build-one case?

tests/ltp/Makefile
```diff
@@ -144,6 +167,9 @@ clean:
 	sudo rm -rf $(APPDIR) ext2fs appdir $(UPPERDIR) $(WORKDIR) $(HOSTFS)
 	$(MAKE) del_host_users_and_groups
 
+native-clean:
+	sudo rm -rf native_appdir ext2fs
+
```
radhikaj, line 172: Is this line not needed in native-clean because you are adding host users and groups in native_build $(MAKE) del_host_users_and_groups

tests/ltp/Makefile
salsal97: Hi Radhika, thank you for noticing this, removed some unnecessary commands in my revision

tests/ltp/Makefile
salsal97: Removed the addition of host user and groups from these targets since they are run natively

radhikaj (approved):

PR #789, opened 2021-09-17
EXPERIMENTAL: Thread interrupt by signal

mikbras:
This PR is a first step towards interrupting threads that are blocking in the target.

General approach
--------------------

For every thread that enters the kernel, we install an empty signal handler for **SIGUSR2**, block that signal, and then enter the kernel.

For every thread that makes a blocking target call (out-call for SGX), we unblock **SIGUSR2**, perform the blocking call, and then re-block SIGUSR2. The blocking call may return **EINTR** if a **SIGUSR2** was encountered while blocking.

The kernel can interrupt a thread that is blocked on the target by sending it a SIGUSR2 signal with ``myst_tcall_tkill()``, which performs a ``tkill syscall`` on the target, breaking the blocking thread out of the current operation.

Examples of blocking calls include, ``read()``, ``write()``, ``poll()``, ``epoll()``, etc..

**Status:** There may be race condition between the time the signal is unblocked to the time the blocking call is initiated. We are currently investigating this.


tools/myst/host/exec.c
```diff
@@ -79,9 +81,20 @@ static void* _thread_func(void* arg)
...
+    sigemptyset(&set);
+    sigaddset(&set, SIGUSR2);
+
+    /* block SIGUSR2 when inside the enclave */
+    sigprocmask(SIG_BLOCK, &set, NULL);
 
     res = myst_run_thread_ecall(_enclave, &retval, cookie, event, target_tid);
 
+    /* unblock SIGUSR2 when outside the enclave */
+    sigprocmask(SIG_UNBLOCK, &set, NULL);
```
mingweishih, line 96: Wouldn't current pattern block the signals during the thread lifetime?

kernel/syscall.c
```diff
@@ -3166,6 +3167,24 @@ long myst_syscall_get_process_thread_stack(void** stack, size_t* stack_size)
     return ret;
 }
 
+long myst_interrupt_thread(int tid)
```
mingweishih, line 3170: I'd suggest using different name (e.g., `myst_interrupt_host_thread`) or something else to distinguish with future support on interrupting thread while executing inside the enclave.

tools/myst/host/exec.c
mikbras: Yes, the goal is to block all signals except when inside ``ppoll()`` or ``epoll_pwait()``. These functions take a mask to atomically unblock signals during their operation. This is the only window in which we want to get ``SIGUSR2``.

kernel/syscall.c
mikbras: Good idea!

PR #795, opened 2021-09-20
Run make clean after running tests in pipeline

asvrada:
Run make clean after tests in pipeline, also re-enable Pytorch solution

This commit passes nightly pipeline
https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10818&view=logs&j=ab477ad6-9781-5b54-82ca-a0141b26726e&t=2d9b6f4f-266f-5cad-a875-41dc232b2771


.azure_pipelines/ci-pipeline-makefile.yml
```diff
@@ -58,6 +58,7 @@ jobs:
       # run all tests
       - script: |
           make -j tests ALLTESTS=1
+          make -C tests clean
```
jxyang, line 61: The 'solutions' are the bigger spender of disk spaces. Also the clean up should be only enabled for pipeline runs. We do want to keep the intermediate files for local runs in case we need to investigate failures.

asvrada: `make clean` will only run for pipeline. It shouldn't affect local runs.

In this case, cleaning tests makes enough room for solution tests

asvrada:
Ok it didn't pass this time... 
https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10864&view=results

Will modify this PR to delete solution tests after each one's completion.

asvrada:
Also deleted 2 files under solutions/pytorch_inference/misc/ as they are not used.

CyanDevs (approved):

asvrada:
This commit passes nightly pipeline: https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10907&view=results

solutions/Makefile
```diff
@@ -42,8 +41,9 @@ $(DIRS):
 clean:
 	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) clean $(NL) )
 
+# Run each solution tests, and clean each test after its completion to free up space 
 run:
-	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) run $(NL) )
+	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) run && $(MAKE) -C $(i) clean $(NL) )
```
CyanDevs, line 46: This would only make clean if the test passed. Though it makes a small impact, make clean should be done regardless test results. I would recommend using a bash for loop instead of the foreach function.

asvrada: I think it's reasonable to keep the file (don't run make clean) if a test failed so that we can investigate if we want.  And as for bash loop vs foreach, I think foreach is simpler to understand here, do I miss something?

CyanDevs: Actually, I'm going to go back on what I said last time as I did not consider others investigating manual builds. Cleanups are great on automated builds, but not on manual builds. This change would affect both.

@jxyang would you be fine with this clean up for solution tests during manual builds?

jxyang: I'd prefer the clean up is only activated for pipeline runs. See my comment above.

Is there an environment variable indicating the test/solution is being run on the pipeline? If not we should introduce one.

jxyang: That's acceptable to me. Maybe there is a way (defs.mak?) you can do that without touching hundreds of make files.

rs-- (approved):

jxyang (changes requested):
Please limit the changes to pipeline runs.

asvrada:
I realize we need to build, run and clean each solution test, rather than build everything and clean after running each test. The original problem here is we don't have enough space to build every test. So I am working on something like

```
build_run_clean:
	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) && $(MAKE) -C $(i) run && $(MAKE) -C $(i) clean $(NL) )

# Run each solution tests, and clean each test after its completion to free up space
tests_then_clean:
	@ $(MAKE) -s clean
	@ $(MAKE) -j -s build_run_clean
```

But this doesn't have parallel build.

asvrada:
Nightly pipeline passes

https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10961&view=results

asvrada:
This failed PR pipeline twice, with a similar error. This might be caused by parallel running of tests (conflicting port,etc)

asvrada:
I don't think the current fix will work, because running tests in parallel can cause random conflicts. 

#### Summary
Problem: Currently the pipeline would build everything before running tests/solutions, however, the VM doesn't have a big enough disk to store all the generated files.

Solutions tried so far:
1.  Increase VM disk size
    - This is not trivial to do given the current setup of VM
2. Build everything, then delete test after ran
    - This approach still builds everything at once and there isn't enough space
3. Build, run then clean the test parallelly (current code in this PR does this)
    - This would fail because of conflicts between tests

### Proposed solutions:
1. Disable solution tests like pytorch as building them takes too much space (>6G). We will re-enable them after moving to the new system
    - Pros: we don't have to change anything to the current pipeline, and this test is not very crucial, we could afford not to check it regularly in the pipeline
    - Cons: Not running some tests, me wasting few days on this PR
2. Build, run then clean the solution test sequentially
    - Pros: no conflicts between tests, uses less space as previous test is cleaned
    - Cons: Pipeline would take much much longer to finish because no parallel build. Will definitely break the 60min constraint on the PR pipeline

jxyang:
> uild, run then clean the solution test sequentially
> * Pros: no conflicts between tests, uses less space as previous test is cleaned
> * Cons: Pipeline would take much much longer to finish because no parallel build. Will definitely exceed the 60min timeout value on the PR pipeline

I thought Pytorch is added to the nightly pipeline which doesn't have a 1-hour limit?

asvrada:
> > uild, run then clean the solution test sequentially
> > 
> > * Pros: no conflicts between tests, uses less space as previous test is cleaned
> > * Cons: Pipeline would take much much longer to finish because no parallel build. Will definitely exceed the 60min timeout value on the PR pipeline
> 
> I thought Pytorch is added to the nightly pipeline which doesn't have a 1-hour limit?

I am assuming we change all solution execution to sequential for every pipeline, that way code change to all the pipelines would be consistent. But yes we can leave PR pipeline to run with parallel build while changing the rest.

jxyang:
> > > uild, run then clean the solution test sequentially
> > > 
> > > * Pros: no conflicts between tests, uses less space as previous test is cleaned
> > > * Cons: Pipeline would take much much longer to finish because no parallel build. Will definitely exceed the 60min timeout value on the PR pipeline
> > 
> > 
> > I thought Pytorch is added to the nightly pipeline which doesn't have a 1-hour limit?
> 
> I am assuming we change all solution execution to sequential for every pipeline, that way code change to all the pipelines would be consistent. But yes we can leave PR pipeline to run with parallel build while changing the rest.

With the solution tests only active in the PR pipeline, does parallel building makes a big difference?

solutions/Makefile
```diff
@@ -37,8 +36,15 @@ DIRS += redis
 
 build: $(DIRS)
 
+# If BUILD_RUN_CLEAN is set, then build/run/clean each test
```
rs--, line 39: Instead of having an if/else statement if ​BUILD_RUN_CLEAN​ is set, could you just define the ​tests_clean​ stage as:

```
tests_clean: tests
@ $(MAKE) -s clean
```

asvrada:
The Jenkins nightly pipeline passed, with the Pytorch test enabled (and with no cleanup after running tests)
https://oe-jenkins-dev.westeurope.cloudapp.azure.com/job/Mystikos/job/Nightly-Pipeline/63/
This means the issue can be solved simply by using more disk space.

asvrada:
Will close this PR for now. Might re-open if problem still exist after moving to Jenkins

PR #802, opened 2021-09-21
Added code coverage pipeline

rs--:


.jenkins/scripts/global/init-access.sh
```diff
@@ -0,0 +1,10 @@
+echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ bionic main" | sudo tee /etc/apt/sources.list.d/azure-cli.list
+wget https://packages.microsoft.com/keys/microsoft.asc
+sudo apt-key add microsoft.asc
+sudo apt-get update
+while sudo lsof /var/lib/dpkg/lock-frontend | grep dpkg; do sleep 3; done
```
CyanDevs, line 5: I've been thinking if this should load the OE library just to use `helpers.WaitForAptLock()` as this only covers one potential lock.

.jenkins/scripts/global/init-docker.sh
```diff
@@ -0,0 +1,15 @@
...
+  case "${flag}" in
+    u) BUILD_USER=${OPTARG};;
+  esac
+done
+
+if [ -z $BUILD_USER ]; then
+echo "BUILD_USER required. Run script with '-u <BUILD_USER>' flag"; exit 1; fi
+
+sudo usermod -aG docker \${BUILD_USER}
+sudo chmod 666 /var/run/docker.sock
```
CyanDevs, line 12: Please do not do this as it is a big security vulnerability

.jenkins/scripts/global/init-docker.sh
```diff
@@ -0,0 +1,15 @@
+while getopts o: flag
```
CyanDevs, line 1: In my opinion, while this is fine, it is an awkward bash script to have. It does not fully encapsulate the docker setup (e.g. missing docker install) and complicates a series of simple commands with the need for passing args and parsing params.

.jenkins/scripts/global/init-install.sh
```diff
@@ -0,0 +1,3 @@
+sudo apt-get install build-essential lcov python3-setuptools python3-pip llvm-7 libmbedtls-dev docker-ce python3-pip lldb-10 -y
```
CyanDevs, line 1: Could improve readability by having a line per package.

.jenkins/scripts/global/make-world.sh
```diff
@@ -0,0 +1,3 @@
+sudo rm -rf `git ls-files --others --directory`
```
CyanDevs, line 1: Use `$(command)` instead for POSIX compatibility

.jenkins/scripts/solutions/azure-config.sh
```diff
@@ -0,0 +1,5 @@
+sudo apt-get install -y azure-cli
+az login --service-principal -u $SERVICE_PRINCIPAL_ID -p $SERVICE_PRINCIPAL_PASSWORD --tenant $TENANT_ID >> /dev/null
+az account set -s $AZURE_SUBSCRIPTION_ID
+az vm identity assign -g $JENKINS_RESOURCE_GROUP -n `hostname` --identities $MYSTIKOS_MANAGED_ID >> /dev/null
```
CyanDevs, line 4: Use `$(command)` for POSIX compatibility

.jenkins/scripts/solutions/azure-config.sh
```diff
@@ -0,0 +1,5 @@
+sudo apt-get install -y azure-cli
```
CyanDevs, line 1: If you separate this script from the az-cli repo setup, then perhaps this line would be better in the bash script that initializes the apt repo. Then in this script you either assume azure-cli is installed, or do a check to make sure it is installed before proceeding.

.jenkins/scripts/solutions/init-config.sh
```diff
@@ -0,0 +1,4 @@
+echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ bionic main" | sudo tee /etc/apt/sources.list.d/azure-cli.list
```
CyanDevs, line 1: Regarding this file name `solutions/init-config.sh`: this file isn't initialization for solutions test per se. It is for setting up the apt repo for azure-cli, which is a dependency only due to the fact that we're using an alternative authentication method in Jenkins for some solution tests.

.jenkins/scripts/global/init-docker.sh
rs--: It's a copy and paste from an existing pipeline, not sure if it's needed for the test to work

.jenkins/scripts/solutions/init-config.sh
rs--: good point, I'll clean up the naming

.jenkins/scripts/global/init-access.sh
rs--: The functionality of this file was already split into different scripts, I forgot to update this one.

.jenkins/scheduled-tests.Jenkinsfile
```diff
@@ -0,0 +1,234 @@
+/* A Jenkins pipeline that will handle PR and nightly tests
+*  These are the original pipelines:
+*  https://github.com/deislabs/mystikos/blob/main/.azure_pipelines/ci-pipeline-code-coverage-nightly.yml
+*/
+
+pipeline {
```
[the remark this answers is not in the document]
rs--: Given that dotnet is no longer an option, I think this will actually be a lot cleaner as there are fewer differences between the nightly and code coverage pipelines

.jenkins/scripts/global/make-tests.sh
```diff
@@ -0,0 +1 @@
+make -j tests ALLTESTS=1 VERBOSE=1 || echo "continue"
```
CyanDevs, line 1: Curious -- why echo continue on failure?

.jenkins/scripts/global/wait-dpkg.sh
```diff
@@ -0,0 +1 @@
+while sudo lsof /var/lib/dpkg/lock-frontend | grep dpkg; do sleep 3; done
```
CyanDevs, line 1: We should change this to 
 our more thorough script, but since we're out of time for this migration we should leave it to later. https://github.com/openenclave/openenclave/blob/master/.jenkins/library/vars/helpers.groovy#L13 or https://github.com/openenclave/openenclave/blob/master/scripts/ansible/install-ansible.sh#L10

CyanDevs:
As this moves the existing Jenkinsfiles we will need to make sure we change the job configurations to use the new location of the Jenkinsfiles when we merge this in.

.jenkins/scripts/global/make-tests.sh
rs--: This is a temporary change; the tests were being flakey and I wanted to bypass the failed tests

.jenkins/pipelines/scheduled-tests.Jenkinsfile
```diff
@@ -0,0 +1,189 @@
...
+                    withCredentials([string(credentialsId: 'mystikos-sql-db-name-useast', variable: 'DB_NAME'),
+                                     string(credentialsId: 'mystikos-sql-db-server-name-useast', variable: 'DB_SERVER_NAME'),
+                                     string(credentialsId: 'mystikos-maa-url-useast', variable: 'MAA_URL'),
+                                     string(credentialsId: 'mystikos-managed-identity-objectid', variable: 'DB_USERID')]) {
+                        sh "make tests -C ${WORKSPACE}/solutions"
+                    }
+                }
+            }
+        }
+        stage('Run SQL Solution - CACENTRAL') {
```
CyanDevs, line 106: Are both SQL solution for USEAST and CANADACENTRAL necessary? Nightly tests and CC tests only ran one of CANADACENTRAL or USEAST (respectively). Running both will add an hour to the test.

.jenkins/pipelines/scheduled-tests.Jenkinsfile
```diff
@@ -0,0 +1,189 @@
...
+
+                   ${JENKINS_SCRIPTS}/global/wait-dpkg.sh
+                   sudo apt install -y dotnet-sdk-5.0
+
+                   ${JENKINS_SCRIPTS}/global/wait-dpkg.sh
+                   sudo apt install -y cobertura
+
+                   # Publish Code Coverage Results
+                   ls -l lcov*
+                   cat ${WORKSPACE} lcov*
```
CyanDevs, line 179: cat on a workspace will return exit code 1

.jenkins/pipelines/scheduled-tests.Jenkinsfile
```diff
@@ -0,0 +1,189 @@
...
+                   ${JENKINS_SCRIPTS}/global/wait-dpkg.sh
+                   ${JENKINS_SCRIPTS}/code-coverage/download-dotnet.sh
+
+                   ${JENKINS_SCRIPTS}/global/wait-dpkg.sh
+                   sudo apt install -y dotnet-sdk-5.0
+
+                   ${JENKINS_SCRIPTS}/global/wait-dpkg.sh
+                   sudo apt install -y cobertura
+
+                   # Publish Code Coverage Results
```
CyanDevs, line 177: Can publish to a storage blob. An example is: https://github.com/deislabs/mystikos/blob/main/.jenkins/release.Jenkinsfile#L199-L205

.jenkins/pipelines/scheduled-tests.Jenkinsfile
```diff
@@ -0,0 +1,189 @@
...
+    }
+    environment {
+        MYST_SCRIPTS =    "${WORKSPACE}/scripts"
+        JENKINS_SCRIPTS = "${WORKSPACE}/.jenkins/scripts"
+        BUILD_USER = sh(
+            returnStdout: true,
+            script: 'echo \${USER}'
+        )
+        MYST_NIGHTLY_TEST = 1
+        MYST_ENABLE_GCOV = 1
```
CyanDevs, line 26: This would be set as well when the nightly pipeline is ran, yet the original nightly pipeline does not set this. What are the implications? If there are any concerns it would be better to set this only when running the Code Coverage pipeline.

.jenkins/pipelines/scheduled-tests.Jenkinsfile
```diff
@@ -0,0 +1,189 @@
+/* A Jenkins pipeline that will handle code coverage and nightly tests
+*  These are the original pipelines:
+*  https://github.com/deislabs/mystikos/blob/main/.azure_pipelines/ci-pipeline-code-coverage-nightly.yml
```
CyanDevs, line 3: Nit: Since your intention is to replace the nightly pipeline too, please also mention the original nightly pipeline here

CyanDevs (commented):
Looks good now, just a couple remaining comments to address before merging.

.jenkins/pipelines/scheduled-tests.Jenkinsfile
rs--: This file has the possibility of replacing the nightly pipeline; I think we should attempt that outside of this PR. I'll investigate the differences when/if we end up replacing the nightly pipeline with this file.

.jenkins/pipelines/scheduled-tests.Jenkinsfile
rs--: I'll use USEAST for this pipeline

CyanDevs (approved):
LGTM. Thanks Rob!

.jenkins/pipelines/scheduled-tests.Jenkinsfile
```diff
@@ -0,0 +1,173 @@
...
+*/
+
+pipeline {
+    agent {
+        label 'ACC-1804-DC4'
+    }
+    options {
+        timeout(time: 600, unit: 'MINUTES')
```
salsal97, line 11: we might only need a 300 min timeout - afraid if a 10 hr timeout would cause extra resource usage in case something is stuck?

PR #803, opened 2021-09-22
Update cpython test suite configuration

bodzhang:
- Enable more test cases
  - Seperate cpython3.8 and cpython3.9 test configuration, as certain cpython3.9
    test cases are not present in cpython3.8
  - Add config.json file to set current working directory in Mystikos environment
  - Add more depedency modules in the Dockerfile


bodzhang:
cpython test passed in [this](https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=10927&view=results) nightly pipeline run

tests/cpython-tests/Makefile
```diff
@@ -2,18 +2,18 @@ TOP=$(abspath ../..)
 include $(TOP)/defs.mak
 
 APPBUILDER=$(TOP)/scripts/appbuilder
-HEAP_SIZE="4G"
-OPTS += --memory-size $(HEAP_SIZE) --nobrk
+
+OPTS = --nobrk --app-config-path config.json
```
vtikoo, line 6: nit: --nobrk can also be specified in config.json

vtikoo (approved):

vtikoo:
We should reevaluate at some point if splitting of the passing tests list is still required. Originally I had split the test lists due to memory leaks.

tests/cpython-tests/Makefile
bodzhang: Added "NoBrk" in config.json and keep command line options for Linux target

tests/cpython-tests/Makefile
```diff
@@ -2,18 +2,23 @@ TOP=$(abspath ../..)
...
-OPTS += --memory-size $(HEAP_SIZE) --nobrk
+
+OPTS = --app-config-path config.json
 ifdef STRACE
 OPTS += --strace
 endif
+
+#Linux target ignores config.json, add equivalent command line options as a workaround
+HEAP_SIZE="4G"
+OPTS += --memory-size $(HEAP_SIZE) --nobrk
```
vtikoo, line 13: This can be wrapped in a `ifeq ($TARGET, linux)`.

bodzhang: In fact, Linux target does process config.json, but have slightly different logic than sgx target. Luckily, for this specific config.json, the difference in the logic does not have negative impact. Removed the command line options in the latest update to the PR.

Issue #807, opened 2021-09-22
myst exec-linux processes config.json differently compared with myst exec-sgx


bodzhang:
`myst exec-sgx` configures the app's runtime environment in Mystikos based on app related entries in config.json, for example, `MemorySize`, `CurrentWorkingDirectory`, `EnvironmentVariables`, with certain logic on whether to override command line options using the entries in config.json. The `myst exec-linux` implementation behaves differently with regard to certain config.json entries.  This difference makes the Linux target  a low fidelity simulation environment to debug issues encountered in SGX target.

bodzhang:
For Linux target, it's not possible to set `CurrentWorkingDirectory`, as config.json is ignored, and `myst exec` does not support a command line option to set `CurrentWorkingDirectory` either.

bodzhang:
- Linux target [overrides `fork_mode` as `none` even if the entry is not present in config.json](https://github.com/deislabs/mystikos/blob/84428561b96f535812df04a1cbb165360d701c3c/tools/myst/host/exec_linux.c#L371), different from the logic in SGX target;
- Linux target does not process `ApplicationPath` ,  'ApplicationParameters` entries from config.json
- Linux target does not process `EnvironmentVariables` entry from config.json;
- Linux target does not support overriding `max_affinity_cpus` command line option with entry from config.json;

bodzhang:
Addressed by PR #809

PR #809, opened 2021-09-22
Make sure that on SGX we always use config

paulcallen:
Command line is not allowed to override config.
Even items not specified in config will use the config defaults.
Allowing config from command line to override the config breaks the
attestation promise


paulcallen:
fixes #807

PR #810, opened 2021-09-23
Ignore self-loops when resolving symlinks.

anakrish:
This prevents infinite recursion.

### DIFF
--- a/ext2/ext2.c
+++ b/ext2/ext2.c
@@ -1334,15 +1334,20 @@ static int _path_to_ino_recursive(
                     }
                 }
 
-                ECHECK(_path_to_ino_recursive(
-                    ext2,
-                    locals->target,
-                    current_ino,
-                    FOLLOW,
-                    &current_ino,
-                    &locals->ino,
-                    realpath,
-                    target_out));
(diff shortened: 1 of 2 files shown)


ext2/ext2.c
```diff
@@ -1334,15 +1334,20 @@ static int _path_to_ino_recursive(
...
+                    ECHECK(_path_to_ino_recursive(
+                        ext2,
+                        locals->target,
+                        current_ino,
+                        FOLLOW,
+                        &current_ino,
+                        &locals->ino,
+                        realpath,
+                        target_out));
+                }
```
vtikoo, line 1350: Does this return an error code if its a self-loop?

vtikoo (commented):
Similar patch might work for ramfs as well.

ext2/ext2.c
anakrish: On Linux, realpath of self-loop does not seem to raise any errors.

mingweishih (approved):

mikbras (approved):
Really cool. Thank you for doing this!

vtikoo (approved):

Issue #46, opened 2021-01-26
Support legacy usage of SYS_mknod and SYS_mknodat


jxyang:
According to Linux man page:

> POSIX.1-2001 says: "The only portable use of mknod() is to create a
FIFO-special file. If mode is not S_IFIFO or dev is not 0, the
behavior of mknod() is unspecified." However, nowadays one should
never use mknod() for this purpose; one should use mkfifo(3), a
function especially defined for this purpose.

However, we don't support mkfifo in Myst either. And it seems some applications, such as dotnet runtime, still uses mknod to create named pipes.

paulcallen:
currently we only support this for unix domain sockets, but not for pipes devices.

bodzhang:
Current Mystikos implementation returns success on mknod(S_IFIFO), without creating a FIFO-special file.

mikbras:
.NET attempts a ``mknod`` system call unless the following environment variable is define.

```
COMPlus_EnableDiagnostics=0
```

vtikoo:
For tracing related dotnet tests, `COMPlus_EnableDiagnostics` needs to be turned on. One of the threads keeps [reading on a pipe created via `mknod` in a loop](https://github.com/dotnet/runtime/blob/v5.0.0-rtm.20519.4/src/coreclr/src/debug/shared/dbgtransportsession.cpp#L1319). This fails currently as the /tmp path related to the pipe is not present in the enclave rootfs, causing an indefinite loop.

PR #811, opened 2021-09-23
Add /proc/fd link on dup() for files

vtikoo:

### DIFF
--- a/kernel/fdtable.c
+++ b/kernel/fdtable.c
@@ -374,6 +374,19 @@ int myst_fdtable_dup(
         new->device = old->device;
         new->object = newobj;
 
+        // add /proc/self/fd/[fd] entry only for files
+        // ATTN: update once pipes can be accessed by pathnames, GH #46
+        if (newfd != oldfd && new->type == MYST_FDTABLE_TYPE_FILE)
+        {
+            myst_fs_t* newfs = (myst_fs_t*)new->device;
+            if ((ret = myst_add_fd_link(newfs, new->object, newfd)) < 0)
+            {
+                myst_fdtable_remove(fdtable, newfd);
+                (*newfs->fs_close)(newfs, new->object);
+                ERAISE(ret);
+            }
+        }
+
         ret = newfd;
     }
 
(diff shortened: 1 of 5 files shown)


kernel/fdtable.c
```diff
@@ -374,6 +374,12 @@ int myst_fdtable_dup(
         new->device = old->device;
         new->object = newobj;
 
+        // add /proc/self/fd/[fd] entry for files
+        if (newfd != oldfd && new->type == MYST_FDTABLE_TYPE_FILE)
+        {
+            myst_add_fd_link(new->device, new->object, newfd);
```
bodzhang, line 380: Can this operation fail? If it fails, do we need to clean up anything from fdtable?

vtikoo: Have updated to closing the new file handle and forward the error.

kernel/fdtable.c
```diff
@@ -374,6 +374,18 @@ int myst_fdtable_dup(
         new->device = old->device;
         new->object = newobj;
 
+        // add /proc/self/fd/[fd] entry for files
+        if (newfd != oldfd && new->type == MYST_FDTABLE_TYPE_FILE)
```
bodzhang, line 378: If the oldfd is a pipe, should a link also be added?

vtikoo: Adding /proc/self/fd/[fd] symlink requires the target device(the one which the fd belongs to), to support `fs_realpath`. pipdev.c does not support this currently.

bodzhang: That's unfortunate. The app exposed the issue of missing /dev/fd link created a pipe, dup2 one end of it, and attempted to access it through /dev/fd/[fd].

vtikoo: Is it creating a named pipe? We have an outstanding issue for that - #46.

bodzhang: Not sure whether it's a named pipe. Please add an ATTN comment explaining that once support for /dev/fd/[fd] for other type of descriptors are added, the dup code here need to change.

bodzhang (approved):

PR #819, opened 2021-09-24
Fix myst_cond_timedwait() recursion bug

mikbras:
Paul and I found that ``myst_cond_timedwait()`` can be entered twice by the same thread under the following scenario.
- The app calls ``futex()`` which enters ``myst_cond_timedwait()`` the first time.
- The ``myst_cond_timedwait()`` function discovers and dispatches a signal.
- The signal is dispatched to the app.
- The app calls ``futex()`` which enters ``myst_cond_timedwait()`` the second time.

Paul suggested calling ``myst_signal_has_active_signals()`` within ``myst_cond_timedwait()`` and ``myst_mutex_lock()``. This function returns true if there are signals on the given thread. We then take the return path and dispatch signals after we have reached a safe state. This patch renames the mutex and cond functions and makes them static and then defines new functions with the original names that dispatch the signals.

### DIFF
--- a/kernel/cond.c
+++ b/kernel/cond.c
@@ -56,53 +56,13 @@ typedef struct myst_cond_thread_sig_handler
 
 } myst_cond_thread_sig_handler_t;
 
-/* Our default handler just needs to remove itself from the queue and call the
- * next in line */
-static void myst_cond_sig_handler(
-    MYST_UNUSED unsigned signum,
-    void* _sig_handler)
-{
-    myst_cond_thread_sig_handler_t* sig_handler =
-        (myst_cond_thread_sig_handler_t*)_sig_handler;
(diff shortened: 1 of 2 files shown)


kernel/cond.c
```diff
@@ -320,3 +284,12 @@ int myst_cond_requeue(
 
     return 0;
 }
+
+int myst_cond_timedwait(
+    myst_cond_t* c,
+    myst_mutex_t* mutex,
+    const struct timespec* timeout)
+{
+    myst_signal_process(myst_thread_self());
```
paulcallen, line 293: why not do the same while loop that we have in mutex?
If not then we dont need this processing of signals (which should have been after the timed wait anyway)

paulcallen (approved):

PR #821, opened 2021-09-24
Check process ownership for mmap with address hint

vtikoo:


kernel/syscall.c
```diff
@@ -3627,6 +3627,31 @@ static long _syscall(void* args_)
                 myst_panic("mmap unsupported: pseudo fork process is calling "
                            "mmap when running in pseudo_wait mode");
 
+            if ((uintptr_t)addr % PAGE_SIZE || !length)
+                BREAK(_return(n, -EINVAL));
+
+            /* mman supports addr hint only if the mapping already exists. If
+             * addr hint is provided, process must own the existing mapping. */
+            if (addr && length)
```
anakrish, line 3635: According to https://www.man7.org/linux/man-pages/man2/mmap.2.html,

>If addr is not NULL, then the
       kernel takes it as a hint about where to place the mapping; on
       Linux, the kernel will pick a nearby page boundary (but always
       above or equal to the value specified by
       /proc/sys/vm/mmap_min_addr) and attempt to create the mapping
       there.  **If another mapping already exists there, the kernel picks
       a new address that may or may not depend on the hint*

Will we end up reusing the mapping if the same address is passed?

vtikoo: I think the non-null case we cared about most was from musl libc loader(more details in this [code comment](https://github.com/deislabs/mystikos/blob/main/kernel/mmanutils.c#L284-L287)). Though it looks like that usage is using MAP_FIXED. I have updated the PR to support non-null addr just for MAP_FIXED. Lets see if it causes any incompatibilities in other tests.
cc: @mikbras

vtikoo: Successful PR pipeline run [here](https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=11325&view=results).
The failure reported in the run reported in the Github UI seems to be a flaky ltp test.

anakrish (approved):
LGTM. Might want to get approval from Mike as well.

PR #825, opened 2021-09-25
Ssr sample

jxyang:
This PR adds a Secure Secret Release solution to show the end-to-end workflow of retrieving a key from MHSM.

The test case in the solution is supposed to match the design doc in another PR: https://github.com/jxyang/mystikos/blob/secret-provisioning-2/doc/design/secret-provisioning.md


jxyang:
Superseded by https://github.com/deislabs/mystikos/pull/848

PR #826, opened 2021-09-27
Add support for futex bitset and realtime clock

Sahakait:
Futex syscall now supports `FUTEX_CLOCK_REALTIME` and `FUTEX_{WAKE,WAIT}_BITSET`.

Real time clock
This is implemented by converting the timeout parameter into a relative timeout.

Bitset support
Most of the change is in `cond.c`. See [futex manpage](https://man7.org/linux/man-pages/man2/futex.2.html) for expected functionality. 

**Testing**:
Added test cases for wait/wake with bitset in `test\futex`.
Added test case for realtime clock in `test\futex`.
Added new test file in `test\thread_queue`

### DIFF
--- a/include/myst/thread.h
+++ b/include/myst/thread.h
@@ -3,6 +3,7 @@
 
 #ifndef _MYST_THREAD_H
 #define _MYST_THREAD_H
+
 #define _GNU_SOURCE
 #include <signal.h>
 #include <sys/resource.h>
@@ -212,6 +213,7 @@ struct myst_thread
 
     /* used by myst_thread_queue_t (condition variables and mutexes) */
     struct myst_thread* qnext;
+    uint32_t qbitset;
 
     /* for jumping back on exit */
     myst_jmp_buf_t jmpbuf;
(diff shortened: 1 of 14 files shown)


include/myst/thread.h
```diff
@@ -530,4 +551,79 @@ MYST_INLINE void myst_thread_sig_handler_uninstall(
...
+    myst_thread_queue_t* queue)
+{
+    return __myst_thread_queue_pop_front(queue, NULL);
+}
+
+MYST_INLINE myst_thread_t* myst_thread_queue_pop_front_bitset(
+    myst_thread_queue_t* queue,
+    uint32_t* bitset)
+{
+    return __myst_thread_queue_pop_front(queue, bitset);
```
radhikaj, line 579: Is there a need for a function wrapping this function? Can it be called directly instead?

tests/thread_queue/thread_queue.c
```diff
@@ -0,0 +1,276 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include <assert.h>
+#include <errno.h>
+#include <stdlib.h>
```
radhikaj, line 6: Are there LTP tests that can be run for FUTEX?

include/myst/thread.h
Sahakait: This was part of Mike's work to ensure that the existing flow would remain the same if bitset is not used. @mikbras, is this something that we simplify in a followup PR?

tests/thread_queue/thread_queue.c
Sahakait: `/ltp/testcases/kernel/syscalls/futex/futex_wait_bitset01` tests bitset directly but is hanging, I'm still investigating.

mikbras:
I rebased and restarted the pipeline.

PR #828, opened 2021-09-27
EXPERIMENTAL: Cleanup appdir/rootfs for pipeline

mikbras:


tests/Makefile
```diff
@@ -132,8 +132,14 @@ dirs: $(DIRS)
...
 
+ifdef MYST_REMOVE_APPDIRS
+remove_appdir = rm -rf $(1)/appdir
+else
+remove_appdir =
+endif
+
 __tests:
-	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) tests $(NL) )
+	@ $(foreach i, $(DIRS), $(MAKE) -C $(i) tests; $(call remove_appdir,$(i)) $(NL) )
```
jxyang, line 142: Can you add the additional step to `solutions` as well?

mikbras: I don't think this is going to work, which is why it is experimental. I will update soon.

radhikaj (approved):

tests/Makefile
mikbras: Okay, it seems to work and I addressed your request.

jxyang:
Nice. The pipeline shows we start from a 84%-used disk and we are still able to run through all solutions/tests without disk problem.

jxyang (approved):

PR #835, opened 2021-09-28
Add /sys/devices/system/cpu support

Francis-Liu:

### DIFF
--- a/kernel/enter.c
+++ b/kernel/enter.c
@@ -102,71 +102,6 @@ done:
     return ret;
 }
 
-static int _copy_host_etc_files()
-{
-    int ret = 0;
-    int fd = -1;
-    const char* resolv_file = "/etc/resolv.conf";
-    void* buf = NULL;
-    size_t buf_size;
-    struct stat statbuf;
-
-    ECHECK(myst_load_host_file(resolv_file, &buf, &buf_size));
-
-    if (stat(resolv_file, &statbuf) == 0)
-    {
-        if ((myst_syscall_unlink(resolv_file)) < 0)
-        {
-            myst_eprintf("kernel: failed to unlink file %s\n", resolv_file);
-            ERAISE(-EINVAL);
-        }
-    }
-    else
-    {
-        if (stat("/etc", &statbuf) == -1)
-        {
-            if ((myst_mkdirhier("/etc", 0755)) != 0)
-            {
-                myst_eprintf("kernel: failed to mkdir /etc\n");
-                ERAISE(-EINVAL);
-            }
-        }
-        else if (!S_ISDIR(statbuf.st_mode))
-        {
-            if ((myst_syscall_unlink("/etc")) < 0)
-            {
-                myst_eprintf("kernel: failed to unlink file /etc\n");
-                ERAISE(-EINVAL);
-            }
-            if ((myst_mkdirhier("/etc", 0755)) != 0)
-            {
-                myst_eprintf("kernel: failed to mkdir /etc\n");
-                ERAISE(-EINVAL);
-            }
-        }
-    }
-    if ((fd = creat(resolv_file, 0644)) < 0)
-    {
-        myst_eprintf("kernel: failed to open file %s\n", resolv_file);
-        ERAISE(-EINVAL);
-    }
-    if ((myst_write_file_fd(fd, buf, buf_size)) < 0)
-    {
-        myst_eprintf("kernel: failed to write to file %s\n", resolv_file);
-        ERAISE(-EINVAL);
-    }
-
-done:
-
-    if (fd != -1)
-        close(fd);
-
-    if (buf)
-        free(buf);
-
-    return ret;
-}
-
 static int _setup_tty(void)
 {
     int ret = 0;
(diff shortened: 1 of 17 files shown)


kernel/enter.c
```diff
@@ -808,6 +809,9 @@ int myst_enter_kernel(myst_kernel_args_t* args)
 
     ECHECK(_copy_host_etc_files());
 
+    /* Setup sysfs */
+    sysfs_setup();
```
vtikoo, line 813: Should this be wrapped in an `ECHECK`?

kernel/hostfile.c
```diff
@@ -115,3 +139,123 @@ int myst_load_host_file(const char* path, void** data_out, size_t* size_out)
...
+
+    if (!src_dir || !dst_dir)
+        ERAISE(-EINVAL);
+
+    ECHECK(fd = _host_open(src_dir, O_RDONLY | O_DIRECTORY, 0));
+
+    if (!(locals = calloc(1, sizeof(struct locals))))
+        ERAISE(-ENOMEM);
+
+    ECHECK(myst_mkdirhier(dst_dir, 07555));
```
vtikoo, line 173:
```suggestion
    ECHECK(myst_mkdirhier(dst_dir, 0755));
```

kernel/hostfile.c
```diff
@@ -115,3 +139,123 @@ int myst_load_host_file(const char* path, void** data_out, size_t* size_out)
...
+            }
+            else if (d_type == DT_REG)
+            {
+                if (myst_load_host_file(locals->src_path, &buf, &buf_size) < 0)
+                {
+                    if (ignore_errors)
+                        continue;
+                    ERAISE(-EINVAL);
+                }
+                ECHECK(new_fd = creat(locals->dst_path, 06444));
```
vtikoo, line 217:
```suggestion
                ECHECK(new_fd = creat(locals->dst_path, 0644));
```

kernel/hostfile.c
```diff
@@ -115,3 +139,123 @@ int myst_load_host_file(const char* path, void** data_out, size_t* size_out)
...
+                ECHECK(myst_write_file_fd(new_fd, buf, buf_size));
+
+                if (new_fd != -1)
```
vtikoo, line 220: The ECHECK on L217 ensures we have a valid `new_fd` here. Wondering if we need the if clause here.

kernel/hostfile.c
```diff
@@ -115,3 +139,123 @@ int myst_load_host_file(const char* path, void** data_out, size_t* size_out)
...
+                    buf = NULL;
+                }
+            }
+            else if (d_type == DT_LNK)
+            {
+                ssize_t n = _host_readlink(
+                    locals->src_path,
+                    locals->symlink_target,
+                    sizeof(locals->symlink_target));
+                ECHECK(n);
```
vtikoo, line 238: We should probably also check the truncation case, where n == `sizeof(locals->symlink_target)`.

kernel/hostfile.c
```diff
@@ -115,3 +139,123 @@ int myst_load_host_file(const char* path, void** data_out, size_t* size_out)
...
+            if (d_type == DT_DIR)
+            {
+                ECHECK(myst_copy_host_directory_recursively(
+                    locals->src_path, locals->dst_path, ignore_errors));
+            }
+            else if (d_type == DT_REG)
+            {
+                if (myst_load_host_file(locals->src_path, &buf, &buf_size) < 0)
+                {
+                    if (ignore_errors)
```
vtikoo, line 213: For my understanding, `ignore_errors` only tolerates errors during copying of regular files? Is this a scenario you saw happen locally?

kernel/sysfs.c
```diff
@@ -0,0 +1,21 @@
...
+#include <myst/eraise.h>
+#include <myst/hostfile.h>
+
+#define SYS_DEVICE_CPU_DIR "/sys/devices/system/cpu"
+#define SYS_DEVICE_NODE_DIR "/sys/devices/system/node"
+
+int sysfs_setup()
+{
+    int ret = 0;
+    ECHECK(myst_copy_host_directory_recursively(
```
vtikoo, line 13: Consider mounting a ramfs based filesystem at /sys vs adding directly to the rootfs. That way we won't write back in the case of hostfs.

kernel/hostfile.c
Francis-Liu: That's right. For example reading `/sys/devices/system/cpu/cpu0/power/autosuspend_delay_ms` will result in an Input/output error. According to https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-power

> Not all drivers support this attribute.  If it isn't supported,
> attempts to read or write it will yield I/O errors.

kernel/enter.c
Francis-Liu: Fixed.

kernel/hostfile.c
Francis-Liu: Fixed.

kernel/hostfile.c
Francis-Liu: Fixed.

kernel/hostfile.c
Francis-Liu: You're right, `new_fd` check is not necessary.

kernel/hostfile.c
Francis-Liu: Fixed.

radhikaj:

PR #837, opened 2021-09-28
Update glibC tests to disable tests that are failing due to missing f…

radhikaj:
…eature support in the ability to call atexit or tests that have a fork inside a fork

After the waitpid stability enhancements in https://github.com/deislabs/mystikos/commit/4fbed213fab3855f7f345f70fa9cb570d96c3626, certain tests in glibC related to atexit are failing. These are because of atexit not being supported in vfork currently.  tst-aio5 is a fork within a fork which is also not a supported scenario. Disabling these for now.

Signed-off-by: Radhika Jandhyala <radhikaj@microsoft.com>

### DIFF
--- a/tests/glibc/tests.failed
+++ b/tests/glibc/tests.failed
@@ -1680,10 +1680,12 @@
 /glibc/build/nptl/tst-cancelx6
 /glibc/build/nptl/tst-cancelx7
 /glibc/build/nptl/tst-cancelx9
+/glibc/build/nptl/tst-cleanup0
 /glibc/build/nptl/tst-cleanup1
 /glibc/build/nptl/tst-cleanup2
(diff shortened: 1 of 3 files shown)


tests/glibc/tests.failed
```diff
@@ -1961,6 +1964,7 @@
 /glibc/build/pwd/tst-putpwent
 /glibc/build/rt/tst-aio10
 /glibc/build/rt/tst-aio4
+/glibc/build/rt/tst-aio5
```
salsal97, line 1967: Since this test is flaky it should be in tests.remove, else next time someone runs the training script it might make its way to the passed list again

salsal97 (approved):
lgtm

PR #847, opened 2021-09-30
Implement pipe fcntl(F_GETFD)

mikbras:
This PR fixes a hang in the cpython test:

```
make one TEST=test_bz2
```

The test still fails but gets beyond the hang and runs many tests successfully first.

### DIFF
--- a/kernel/fdtable.c
+++ b/kernel/fdtable.c
@@ -73,12 +73,25 @@ int myst_fdtable_clone(myst_fdtable_t* fdtable, myst_fdtable_t** fdtable_out)
                 myst_fdops_t* fdops = entry->device;
                 void* object;
                 long r;
+                int fdflags = 0;
 
+                /* Save the file descriptor flags (for pipes only) */
+                if (entry->type == MYST_FDTABLE_TYPE_PIPE)
+                {
+                    fdflags =
+                        (*fdops->fd_fcntl)(fdops, entry->object, F_GETFD, 0);
+                }
+
+                /* Duplicate the object */
                 if ((r = (*fdops->fd_dup)(fdops, entry->object, &object)) != 0)
                 {
                     myst_spin_unlock(&fdtable->lock);
                     ERAISE(r);
                 }
+
+                /* Propagate the file descriptor flags (for pipes only) */
+                if (entry->type == MYST_FDTABLE_TYPE_PIPE && fdflags >= 0)
+                    (*fdops->fd_fcntl)(fdops, object, F_SETFD, fdflags);
 
                 new_entry->type = entry->type;
                 new_entry->device = entry->device;
@@ -141,6 +154,7 @@ int myst_fdtable_cloexec(myst_fdtable_t* fdtable)
     myst_spin_unlock(&fdtable->lock);
 
 done:
+
     return ret;
 }
 
--- a/kernel/pipedev.c
+++ b/kernel/pipedev.c
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#include <assert.h>
 #include <sys/ioctl.h>
 
 #include <myst/buf.h>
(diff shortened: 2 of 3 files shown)


kernel/pipedev.c
```diff
@@ -154,8 +173,8 @@ static int _pd_pipe2(myst_pipedev_t* pipedev, myst_pipe_t* pipe[2], int flags)
         /* Set the state */
         shared->state = STATE_WR_ENABLED;
 
-        /* Set the non-blocking flag */
-        shared->flags = flags;
+        /* Set the pipe id (for debugging) */
```
bodzhang, line 176: Should this operation be under `ENABLE_TRACE` compiling switch?

kernel/pipedev.c
```diff
@@ -169,6 +188,11 @@ static int _pd_pipe2(myst_pipedev_t* pipedev, myst_pipe_t* pipe[2], int flags)
         rdpipe->fd = fds[0];
         rdpipe->mode = O_RDONLY;
         rdpipe->shared = shared;
+        rdpipe->flags = flags;
```
bodzhang, line 191: Should the file status flag include O_CLOEXEC?

kernel/pipedev.c
```diff
@@ -580,7 +634,7 @@ static int _pd_fcntl(
     {
         case F_SETFL:
         {
-            pipe->shared->flags |= arg;
+            pipe->flags |= arg;
```
bodzhang, line 637: Should F_GETFL be serviced by the kernel, instead of invoke the host? How about other fcntl cmds? If a pipe fcntl cmd is invalid or not supported by Mystikos implementation, it's better to error out inside the kernel, instead of exposing the info to the host.

kernel/pipedev.c
bodzhang: On line 156, is it necessary to expose O_CLOEXEC bit setting to the host?

kernel/fdtable.c
```diff
@@ -73,13 +73,22 @@ int myst_fdtable_clone(myst_fdtable_t* fdtable, myst_fdtable_t** fdtable_out)
...
+                fdflags = (*fdops->fd_fcntl)(fdops, entry->object, F_GETFD, 0);
+
+                /* Duplicate the object */
                 if ((r = (*fdops->fd_dup)(fdops, entry->object, &object)) != 0)
                 {
                     myst_spin_unlock(&fdtable->lock);
                     ERAISE(r);
                 }
 
+                /* Propagate the file descriptor flags to the object */
```
bodzhang, line 88: With current Mystikos fd_dup() implementation for ext2fs/ramfs, the old_fd and new_fd of dup shares the same file object and its file descriptor flags.

kernel/pipedev.c
mikbras: Yes, I think that is cleaner. I was leaving it for gdb inspection but it is probably better to remove it.

mikbras: Fixed

kernel/pipedev.c
mikbras: Fixed. The file status no longer includes O_CLOEXEC. Also, we mask off O_CLOEXEC before passing the flag to the host now.

kernel/pipedev.c
mikbras: Fixed. F_GETFL is no completely serviced by the kernel. And now we error out if these flags are not supported.

kernel/fdtable.c
mikbras: I modified ``fdtable.c`` so that we only perform this fdflags propagation for pipes. We will still need to revisit the treatment of CLOEXEC on files.

mikbras:
I reworked the handling of **file status flags** and **file descriptor flags**. The pipe structure now has two fields to make the usage clearer: ``fl_flags`` (corresponds to ``F_GETFL``/``F_SETFL``) and ``fd_flags`` (corresponds to ``F_GETFD``/``F_SETFD``). The ``flags`` parameter passed to ``pipe2()`` is split into these two fields as appropriate.

Two other changes:
- The O_DIRECT flag passed to ``pipe2()`` only applies to the write pipe descriptor (learned through experimentation).
- The O_WRONLY flag is implicitly added to the write pipe descriptor file status flags (also gleaned from experimentation).

bodzhang (approved):

PR #851, opened 2021-09-30
Fix POLLNVAL poll() bug

mikbras:
This PR fixes a bug where the app fails to receive an POLLNVAL (invalid event) on an invalid file descriptor.


kernel/poll.c
```diff
@@ -99,7 +98,8 @@ static long _syscall_poll(struct pollfd* fds, nfds_t nfds, int timeout)
         }
 
         /* get the target fd for this object */
-        if ((tfd = (*fdops->fd_target_fd)(fdops, object)) >= 0)
+        if (tfd == INT_MAX ||
```
bodzhang, line 101: Why treat a closed/invalid fd as a target/host fd of INT_MAX?

paulcallen: This is to force the underlying to process as an invalid file handle. We were getting stuck in a loop because poll was just going around in circles getting a success from the underlying layer but nothing was actually available to be processed. This forces the underlying layer to properly handle this case and cause us to break out of the loop

paulcallen (approved):

PR #853, opened 2021-09-30
fix ttydev get_events() operation

mikbras:
This PR fixes the events returned by the get_events() operation.

### DIFF
--- a/kernel/ttydev.c
+++ b/kernel/ttydev.c
@@ -322,10 +322,19 @@ static int _td_get_events(myst_ttydev_t* ttydev, myst_tty_t* tty)
     if (!ttydev || !_valid_tty(tty))
         ERAISE(-EINVAL);
 
-    if (tty->fd == STDOUT_FILENO)
-        ret |= POLLOUT;
-    else
-        ret = -ENOTSUP;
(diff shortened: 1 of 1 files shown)


kernel/ttydev.c
```diff
@@ -322,10 +322,19 @@ static int _td_get_events(myst_ttydev_t* ttydev, myst_tty_t* tty)
...
         ERAISE(-EINVAL);
 
-    if (tty->fd == STDOUT_FILENO)
-        ret |= POLLOUT;
-    else
-        ret = -ENOTSUP;
+    switch (tty->fd)
+    {
+        case STDIN_FILENO:
+            ret |= POLLIN;
```
bodzhang, line 328: Does Mystikos support read from stdin? BTW, myst_tcall_read_console() implementation proceeds if the fd is stdout or stderr, and return error if fd is stdin. Is that intended?

mikbras: I'm sorry, I should have marked this one as experimental.

I think there is a security concern so we should abandon this PR.

mikbras:
Closing for now. Reading from stdin could pose a security concern and not supported by Mystikos yet.

PR #855, opened 2021-10-01
New script to reduce size of appdir's

mikbras:


radhikaj:
@mikbras , can the commits be squashed?

salsal97:
@mikbras I think your `git add` missed the squash-appdir.sh script - or is it being pulled from elsewhere?

mikbras:
> @mikbras I think your `git add` missed the squash-appdir.sh script - or is it being pulled from elsewhere?

Haha! Thanks for catching that; git add ignores .sh files unless you use -f aparently.

mikbras:
> @mikbras , can the commits be squashed?

Done. I usually wait until just before I commit (so I will be able to back out later commits). But since this one is so trivial, I squashed up front. :)

defs.mak
```diff
@@ -365,3 +365,11 @@ RUNTEST_COMMAND=$(TOP)/scripts/runtest
...
 
 MKROOTFS=$(TOP)/scripts/mkrootfs
+
+##==============================================================================
+##
+## SQUASH-APPDIR script
+##
+##==============================================================================
+
+SQUASH-APPDIR=$(TOP)/scripts/squash-appdir.sh
```
jxyang, line 375: I think "slim" is a better term than "squash". :-)

scripts/squash-appdir.sh
```diff
@@ -0,0 +1,48 @@
...
+
+appdir=$(realpath $1)
+
+if [ "${appdir}" == "/" ]; then
+    echo "$0: refusing to squash the system root directory"
+    exit 1
+fi
+
+# Remove files with these extensions: .c, .h, .o
+rm -rf `find ${appdir} -name '*.[cho]'`
```
jxyang, line 26: I think we can also remove *.a files.

mikbras: Yes, I removed them and I was surprised that .NET actually checks dependencies on .a files and then fails.

defs.mak
mikbras: I guess it is a matter of taste. I like "squash" better than "slim" in this context.

salsal97: Would "clean" be better? When the context was "squash" and without seeing the script, the first thing I thought was we are taking all the appdirs and combining them to reduce the build time for the rootfs by building one instead of N, but seeing the script I realize we are reducing the size of individual appdirs. Just my first impression here.

mikbras: Hmm? Clean makes me think it does away with the appdir althogether.

mikbras: Smash? Crush? Deflate?

salsal97: How about prune?

PR #860, opened 2021-10-04
Add option to enable executable stack

Sahakait:
Nested functions creates [trampolines](https://gcc.gnu.org/onlinedocs/gccint/Trampolines.html), which requires the stack to be executable. This is disabled by default as a security precaution.


third_party/musl/crt/patch.diff
```diff
@@ -243,15 +271,40 @@ index afec985a..0ac9f0c4 100644
...
++				
++				if (ph->p_flags & PF_W)
++					prot_flags |= PROT_WRITE;
++				
++				if (ph->p_flags & PF_R)
++					prot_flags |= PROT_READ;
++				
++				void *stack_p = 0;
++				size_t *stack_size = 0;
++				if (!syscall(SYS_get_process_thread_stack, &stack_p, &stack_size))
```
paulcallen, line 291: I am unsure if SYS_get_process_thread_stack is going to work in all scenarios.
This was written for use in fork(). In fork the call can be called from any thread and it first checks the pthread structure to see if it has a stack from present there. Then if that does not exist it falls back to this syscall. 
With the first process to execute the kernel allocates the stack. This stack is returned from the syscall.
I cannot remember the situation for all other situations though regarding the process thread of child processes.
You will need to check if a fork/exec or spawned child process can also access the stack with this syscall, or if you need to retrieve it from the pthread structure

PR #865, opened 2021-10-05
Pipe fixes to enable cpython test_io suite

vtikoo:
### Summary
- Support setting nonblocking mode via ioctl for pipes.
- Fix return value for lseek on a pipe. On Linux this returns an ESPIPE.
- Fix bug related to clearing of pipe file flags after `fcntl(F_SETFL)`.


vtikoo:
tests/cpython-tests successful run [here](https://openenclave.visualstudio.com/ACC-Services/_build/results?buildId=11393&view=logs&j=ab477ad6-9781-5b54-82ca-a0141b26726e&t=a1e294d4-8895-5a45-41ff-022cede0621b).

PR #924, opened 2021-10-21
Install python3-pip or python-pip

CyanDevs:


CyanDevs:
FYI - handing this over to @rs-- as part of his work on Ubuntu 20.04

PR #928, opened 2021-10-21
change exec stack from malloc to mmap

paulcallen:


paulcallen:
asserts of the returns fixed to check for 0 only.

paulcallen:
re-wrote to replace exec-stack with mmap page instead of alloced page

PR #955, opened 2021-10-28
Add PTY support

jxyang:
1) Implement PTY (master) and PTS(slave) devices under devfs
2) Allow writing to a master device and reading from the paired slave and vice versa
3) Allow PTY devices to be configured/queried with ioctl and fnctl
4) Fix a bug in MUSL's implementation of ioctl.


kernel/devfs.c
```diff
@@ -81,6 +99,134 @@ static int _urandom_read_cb(void* buf, size_t count)
...
+
+    if ((pair->path_slave = strdup(tmp)) == NULL)
+        ERAISE(-ENOMEM);
+
+    pair->file_master = file;
+    pair->next = _pty_pairs;
+    _pty_pairs = pair;
+
+    v_cb.open_cb = _open_slave_pty_cb;
+    v_cb.close_cb = _close_pty_file_cb;
```
vtikoo, line 220: When is `close_cb` used?

kernel/devfs.c
```diff
@@ -81,6 +99,134 @@ static int _urandom_read_cb(void* buf, size_t count)
...
+    int ret = 0;
+    for (struct pty_pair* tmp = _pty_pairs; tmp; tmp = tmp->next)
+    {
+        if (tmp->file_slave == file)
+        {
+            ret = myst_read_virtual_file(tmp->file_master, buf, count);
+            goto done;
+        }
+    }
+    ERAISE(-EINVAL); /* can't find the paired slave PTY device */
```
vtikoo, line 182:
```suggestion
    ERAISE(-EINVAL); /* can't find the paired master PTY device */
```

kernel/ramfs.c
```diff
@@ -2964,6 +2966,34 @@ int myst_create_virtual_file(
     return ret;
 }
 
+int myst_read_virtual_file(myst_file_t* file, void* buf, size_t buf_size)
```
vtikoo, line 2969: Is this read function generalizable? It seems like its suitable for stateful virtual files, where reads consume from `file->vbuf`.

vtikoo (commented):
Thanks for improving the virtual file support. Looks great overall.
It would be great to add vbuf locks to the existing callbacks, not necessarily as part of this PR.

kernel/ramfs.c
```diff
@@ -2276,6 +2285,17 @@ static int _fs_ioctl(
             _set_fd_flag(file, 0);
             break;
         }
+        case TIOCSPTLCK:
```
vtikoo, line 2288: Should the `devfs_is_pty_pts_device` check be present for TIOCSPTLCK and TIOCGPTN as well?

kernel/devfs.c
jxyang: It used by `_close_pty_file_cb`, where we tear down the master/slave pair when either the master or the slave device is closed.

jxyang:
Thanks for the review. The stateless virtual files, such as /proc/cpuinfo are populated just once by one thread, so I didn't use the lock for them. Are there other types of existing virtual files that require locking? We should certainly use the lock for future additions of stateful virtual files.

kernel/devfs.c
vtikoo: Gotcha. So how is that function being invoked from userspace. Is it called when the file handle is closed?

kernel/ramfs.c
vtikoo: As this is specific to stateful virtual files, can we qualify the function name to suggest that?
Maybe `myst_read_stateful_virtual_file`.

jxyang: Renamed.

kernel/devfs.c
jxyang: Thanks for catching that. Added the plumbing in _fs_close so the call back can be triggered.

vtikoo:
> The stateless virtual files, such as /proc/cpuinfo are populated just once by one thread, so I didn't use the lock for them. Are there other types of existing virtual files that require locking? 

Makes sense. The stateless /dev entries like zero, null; also don't need locks.

vtikoo (approved):

PR #962, opened 2021-11-01
Parallel Test Execution

rs--:
Parallel test execution workflow:

1. Run make build on DC8 node to improve build time from ~30 minutes to under 15 minutes. The resulting build folder will be used for the solutions tests.
2. The solutions tests (SQL, Azure SDK, and DotNet) will download the prebuilt build folder, and run the respective tests. These tests are built on DC4 nodes, as test runs on DC8 nodes did not show much improvements.
3. Due to some build bits using absolute paths, the build folder will need to rebuilt in the node running the unit tests. Tests will be built and ran on a DC8 node as the build time is reduced by ~20 minutes, and the test runtime is further reduced by another ~20 minutes. Code coverage analysis will run if enabled.

Running the equivalent of the nightly/code coverage build in the manner described above reduces build/run time from ~5 hours, to ~2 hours. Though the total compute runtime is the same, (Unit tests: ~1 hour 40 minutes, Azure SDK: ~25 minutes, DotNet: ~1 hour 30 minutes, SQL: ~1 hour 5 minutes, and the initial repo source build: ~ 15 minutes), the rate-determining step is the slowest build.

The main trade-off with running the tests in parallel is the number of cores used for a given run. Instead of using a single DC4 instance for the whole run, one DC8 instance is used for ~1 hour 40 minutes, and three DC4 instances are used between 20 minutes to 1.5 hours.

To further expedite the build process, I introduced the use of a DC8 instance which will be used for the initial build, and is kept alive as long as it's been used within the past hour (@CyanDevs I'd especially like your feedback on this). This reduces the build process by several minutes it usually takes to create an instance; this change is more focused on the user experience.

Moving forward, users could be given the option to select which tests they would like to run, reducing the number of test files we have.

### DIFF
--- a/.jenkins/pipelines/code-coverage.Jenkinsfile
+++ b/.jenkins/pipelines/code-coverage.Jenkinsfile
@@ -1,0 +1,88 @@
+/* A Jenkins pipeline that will handle code coverage and nightly tests
+*  These are the original pipelines:
+*  https://github.com/deislabs/mystikos/blob/main/.azure_pipelines/ci-pipeline-code-coverage-nightly.yml
+*/
+
+pipeline {
+    agent {
+        label 'ACC-1804-DC2'
+    }
+    options {
+        timeout(time: 300, unit: 'MINUTES')
+    }
+    parameters {
+        string(name: "REPOSITORY", defaultValue: "deislabs")
+        string(name: "BRANCH", defaultValue: "main", description: "Branch to build")
+        choice(name: "REGION", choices:['useast', 'canadacentral'], description: "Azure region for SQL test")
+    }
+    stages {
+        stage('Run Tests') {
+            parallel {
+                stage("Run Unit Tests") {
+                    steps {
+                    catchError(buildResult: 'FAILURE', stageResult: 'FAILURE') {
+                        build job: "Standalone-Pipelines/Unit-Test-Pipeline",
+                        parameters: [
+                            string(name: "REPOSITORY", value: REPOSITORY),
+                            string(name: "BRANCH", value: BRANCH),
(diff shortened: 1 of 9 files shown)


CyanDevs (approved):
Great improvements to our build & test times. Thanks for your great design and work!

.jenkins/pipelines/pipeline-tests.Jenkinsfile
```diff
@@ -0,0 +1,167 @@
+/* A Jenkins pipeline that will handle code coverage and nightly tests
+*  These are the original pipelines:
+*  https://github.com/deislabs/mystikos/blob/main/.azure_pipelines/ci-pipeline-code-coverage-nightly.yml
```
CyanDevs, line 3: Nit: comment no longer applicable

.jenkins/pipelines/pipeline-tests.Jenkinsfile
```diff
@@ -0,0 +1,185 @@
...
+            script: 'echo \${USER}'
+        )
+    }
+    stages {
+        stage("Initialize Workspace") {
+            steps {
+                sh """
+                   ${JENKINS_SCRIPTS}/global/clean-temp.sh
+                   """
+                azureDownload(
```
CyanDevs, line 36: This is the first stage of a new PR build right? What build resources is there (or should be) downloaded?

CyanDevs: Is this so one can rerun jobs and skip the build phase?

rs--: Yeah, exactly!

mikbras:
This is really cool!

PR #966, opened 2021-11-02
strace filtering

anakrish:
The following options are added to make strace more helpful
for debugging

- --strace-failing
  When specified, all syscalls that fail will be logged.
  Other syscalls will not be logged, unless specified via filter (see below).
  Set a breakpoing in `_strace_failure_hook` to stop execution whenever a
  syscall fails. Use breakpoint conditions to control the behavior of the breakpoint.

- --strace-filter "SYS_name1:SYS_name2:..."
  Specify the set of syscalls to be traced. When filters are specified, only those
  syscalls specified in the filter will be traced, in addition to failing syscalls if
  specified as described above.
  E.g: To trace open and mprotect syscalls, specify
     --strace-filter "SYS_open:SYS_mprotect"

### DIFF
--- a/include/myst/kernel.h
+++ b/include/myst/kernel.h
@@ -63,6 +63,26 @@ typedef struct myst_crt_args
     myst_wanted_secrets_t* wanted_secrets;
 } myst_crt_args_t;
 
+typedef struct _myst_strace_config
+{
+    /* Is tracing enabled or not */
+    bool trace_syscalls;
+
+    /* Should failing syscalls be traced or not*/
+    bool trace_failing;
+
+    /* Is filtering enabled or not */
+    bool filter;
+
+    /*
+        If filter is enabled, should the given syscall be traced or not.
+        Note: ausyscall --dump shows that maximum syscall value on Ubuntu 18.04
+              is 332 statx.
+              MYST_MAX_SYSCALLS value is much higher (3000).
+    */
+    bool trace[MYST_MAX_SYSCALLS];
+} myst_strace_config_t;
+
 typedef struct myst_kernel_args
 {
     /* The image that contains the kernel and crt etc. */
@@ -170,7 +190,7 @@ typedef struct myst_kernel_args
 
     /* Tracing options */
     bool trace_errors;
-    bool trace_syscalls;
+    myst_strace_config_t strace_config;
 
     /* Whether the target supports the SYSCALL instruction */
     bool have_syscall_instruction;
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -201,12 +201,63 @@ static const char* _format_timespec(
     return buf->data;
 }
 
+static bool _trace_syscall(long n)
+{
+    // Check if syscall tracing is enabled.
+    if (__myst_kernel_args.strace_config.trace_syscalls)
+    {
+        if (__myst_kernel_args.strace_config.filter)
+        {
+            // If filtering is enabled, trace only this syscall has been
+            // specified in the filter.
+            return __myst_kernel_args.strace_config.trace[n];
+        }
+        else
+        {
+            // Trace all syscalls.
+            return true;
+        }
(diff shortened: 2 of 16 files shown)


include/myst/kernel.h
```diff
@@ -58,6 +58,25 @@ typedef struct _myst_wanted_secrets_config
...
+
+    /* Is filtering enabled or not */
+    bool filter;
+
+    /*
+        If filter is enabled, should the given syscall be traced or not.
+        Note: ausyscall --dump shows that maximum syscall value is 332 statx.
+              The largest syscall value supported by mystikos is 2019.
+    */
+    bool trace[2020];
```
mikbras, line 77: ``./kernel/times.c`` defines this:

```
#define MYST_MAX_SYSCALLS 3000
```

It would be nice to have a centralized definition in ``<myst/syscall.h>`` an replace 2020 with that constant.

kernel/syscall.c
```diff
@@ -200,12 +200,63 @@ static const char* _format_timespec(
...
+        return true;
+
+    // Check if tracing failing syscalls has been enabled.
+    if (__myst_kernel_args.strace_config.trace_failing)
+    {
+        // If the syscall returns a negative value and has a valid error name,
+        // then consider it a failure and enable tracing.
+        if (ret < 0)
+        {
+            const char* error_name = myst_error_name(-ret);
```
mikbras, line 246: I guess I should have called this function something closer to the standard C name like ``myst_strerror()``.

tools/myst/host/strace.c
```diff
@@ -0,0 +1,62 @@
...
+    {
+        filter_copy = strdup(filter);
+        char* state = NULL;
+        char* p = strtok_r(filter_copy, ":", &state);
+        while (p)
+        {
+            int num = myst_syscall_num(p);
+            if (num > 0)
+            {
+                if (num < sizeof(strace_config->trace) /
```
mikbras, line 36: If you define ``MYST_MAX_SYSCALLS`` as suggested above, then you can just compare with that constant here.

tools/myst/host/strace.c
```diff
@@ -0,0 +1,62 @@
...
+    if (cli_getopt(argc, argv, "--strace-failing", NULL) == 0)
+    {
+        strace_config->trace_failing = 1;
+        strace_config->filter = 1;
+        ret = 0;
+    }
+
+    if (cli_getopt(argc, argv, "--strace-filter", &filter) == 0 && filter)
+    {
+        filter_copy = strdup(filter);
```
mikbras, line 28: Not a requirement for this PR but I sometimes find it easier to use ``myst_strsplit()`` from ``<myst/strings.h>``.

```
int myst_strsplit(
    const char* str,
    const char* delim,
    char*** toks,
    size_t* ntoks);
```

You can later pass ``toks`` to free.

mikbras:
A very clean change. Thank you Anand for doing this. I just have a couple of minor nits.

include/myst/kernel.h
anakrish: Done.

tools/myst/host/strace.c
anakrish: Done.

tools/myst/host/strace.c
```diff
@@ -0,0 +1,62 @@
...
+        char* state = NULL;
+        char* p = strtok_r(filter_copy, ":", &state);
+        while (p)
+        {
+            int num = myst_syscall_num(p);
```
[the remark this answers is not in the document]
anakrish: Fixed.

tools/myst/host/strace.c
anakrish: Nice idea. Done.

mikbras (approved):
Look great Anand. Thank you for doing this.

PR #968, opened 2021-11-02
Fix pipeline emails

CyanDevs:
* Wrap conditionals in script block
* Email report to Mystikos group only if the build is using deislabs/mystikos and main branch.


rs-- (approved):

rs--:
We should think about create a Jenkins library for Mystikos

PR #969, opened 2021-11-02
Patch to increase MUSL pthread stack size from 128K to 8M

Francis-Liu:


vtikoo (approved):

radhikaj:
Thanks Francis and Vikas

PR #970, opened 2021-11-02
Revert "Patch to increase MUSL pthread stack size from 128K to 8M"

radhikaj:
This reverts commit 43f2cf1ab39bf0799a6d0a917582ec422915aa62.

I merged https://github.com/deislabs/mystikos/pull/969 without realizing that CI had not passed. Reverting now.

Sorry about the error.

### DIFF
--- a/third_party/musl/crt/patch.diff
+++ b/third_party/musl/crt/patch.diff
@@ -623,7 +623,7 @@ index cbabde47..e78515e4 100644
  
  #endif
 diff --git a/src/internal/pthread_impl.h b/src/internal/pthread_impl.h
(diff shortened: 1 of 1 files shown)


vtikoo (approved):

jxyang:
I'd advise against increasing MUSL pthread size. It has a very wide impact on apps that don't even use 128k thread stack.

vtikoo:
> I'd advise against increasing MUSL pthread size. It has a very wide impact on apps that don't even use 128k thread stack.

@jxyang we are seeing issues with musl's 128k limit for a number of python tests built against glibc.

PR #971, opened 2021-11-02
Support relative path in SYS_execve

asvrada:


asvrada:
Passing cpython relative path tests

```
test_cwd_with_absolute_arg (Lib.test.test_subprocess.ProcessTestCase) ... ok
test_cwd_with_bytes (Lib.test.test_subprocess.ProcessTestCase) ... ok
test_cwd_with_pathlike (Lib.test.test_subprocess.ProcessTestCase) ... ok
test_cwd_with_relative_arg (Lib.test.test_subprocess.ProcessTestCase) ... ok
test_cwd_with_relative_executable (Lib.test.test_subprocess.ProcessTestCase) ... ok
test_double_close_on_error (Lib.test.test_subprocess.ProcessTestCase) ... ok
```

kernel/syscall.c
```diff
@@ -2087,11 +2088,23 @@ long myst_syscall_execve(
...
+    // Resolve relative path
+    if (filename[0] != '/')
+    {
+        if (!(resolved_path = (myst_path_t*)malloc(sizeof(myst_path_t))))
+            ERAISE(-ENOMEM);
+
+        if (ret = myst_realpath(filename, resolved_path))
+            ERAISE(ret);
+
+        filename = resolved_path->buf;
```
mingweishih, line 2105: Suggest declare a new variable for this. Reusing a `const` parameter is not a good practice.

kernel/syscall.c
```diff
@@ -2087,11 +2088,23 @@ long myst_syscall_execve(
...
+    if (!filename)
         ERAISE(-EINVAL);
 
+    // Resolve relative path
+    if (filename[0] != '/')
+    {
+        if (!(resolved_path = (myst_path_t*)malloc(sizeof(myst_path_t))))
+            ERAISE(-ENOMEM);
+
+        if (ret = myst_realpath(filename, resolved_path))
```
mingweishih, line 2102: Use `ECHECK(myst_realpath(filename, resolved_path))` instead

kernel/syscall.c
```diff
@@ -2087,11 +2088,23 @@ long myst_syscall_execve(
     long ret = 0;
     const char** argv = NULL;
     myst_thread_t* current_thread = myst_thread_self();
+    myst_path_t* resolved_path = NULL;
```
mingweishih, line 2091: `myst_path_t` is a struct with a char array. You can just declare the variable instead of a pointer.

asvrada: It would breach stack size limit if we do that

`error: stack usage might be 4272 bytes [-Werror=stack-usage=]`

mingweishih: Not sure if it's configurable, can you check? 4272 bytes does not seem to be large to me though.

kernel/syscall.c
asvrada: Adding a new variable might be too confusing? How do you think about this way? No reusing `filename` and no new variable

```
argv[0] = resolved_path ? resolved_path->buf : filename;
```

mingweishih: Would it be conflicted with the logic below? A simple way to just declare
`const char* path = filename;`
And use `path` throughout the function instead.

kernel/syscall.c
asvrada: The error could be ignored with `#pragma GCC diagnostic ignored "-Wstack-usage="`, but I am not sure if its a good idea, and couldn't find the place to configure the threshold... Do you have more info on this issue?

mingweishih: ok, I see we set the limit to 512: https://github.com/deislabs/mystikos/blob/6cdeb80aef0faf7e807b953c44d241f3543a5676/kernel/Makefile#L42

Let's keep the current implementation.

mingweishih (approved):

PR #972, opened 2021-11-02
Fix path/filename limitations in EXT2

mikbras:
This PR corrects errors EXT2 errors with:
- maximum filename length (now ``EXT2_FILENAME_MAX``)
- maximum path length (now ``PATH_MAX``)
- maximum path depth (now arbitrary)

### DIFF
--- a/include/myst/ext2.h
+++ b/include/myst/ext2.h
@@ -23,7 +23,7 @@
 **==============================================================================
 */
 
-#define EXT2_PATH_MAX 256
+#define EXT2_FILENAME_MAX 255
 
 #define EXT2_MAX_BLOCK_SIZE (8 * 1024)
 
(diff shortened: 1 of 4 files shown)


include/myst/ext2.h
```diff
@@ -23,7 +23,7 @@
 **==============================================================================
 */
 
-#define EXT2_PATH_MAX 256
+#define EXT2_FILENAME_MAX 255
```
bodzhang, line 26: Should the definition be 255 or 256?

bodzhang (approved):

bodzhang:
"EXT2_PATH_MAX" still exist in compiled out code in ext2.c and ext2dump.c. Would be good to update those code in case they will be enabled during debug or in the future. Not a blocking issue for this PR in my opinion.

PR #982, opened 2021-11-04
Multiple fixes and enhancements for ext2-fs and ramfs for issues exposed by cpython test suite.

bodzhang:
Fixes
  - Return EBADF instead of ENOENT on dup operation, if the old fd is not valid.
  - Return EINVAL on ext2-fs lseek operation, if the resulting file offset
    would be negative
  - Fix the typo bug in myst_syscall_chmod() return value
  - Return 0 on fs read() if lseek offset is beyond end of the file
  - Fix _ext2_futimens() bug of not writing the update to fs

Enhancement
  - Add O_NONBLOCK flag tracking and FIONBIO ioctl support in ext2fs and ramfs  
  - Add umask support in creat/open/mkdir syscalls 
  - Set S_IFCHR bit in file mode for devfs files such as /dev/null, dev/ptmx 
  - Allow root user to chown to uid/gid not in the passwd file

### DIFF
--- a/kernel/chown.c
+++ b/kernel/chown.c
@@ -71,12 +71,16 @@ long myst_syscall_chown(const char* pathname, uid_t owner, gid_t group)
 
     ECHECK(_path_checks(pathname));
 
-    if (((owner != (uid_t)-1) &&
-         (myst_valid_uid_against_passwd_file(owner) < 0)) ||
-        ((group != (gid_t)-1) &&
-         (myst_valid_gid_against_group_file(group) < 0)))
+    /* if not root, check target uid/gid validity */
+    if (thread->euid != 0)
     {
-        ret = -EINVAL;
+        if (((owner != (uid_t)-1) &&
+             (myst_valid_uid_against_passwd_file(owner) < 0)) ||
+            ((group != (gid_t)-1) &&
+             (myst_valid_gid_against_group_file(group) < 0)))
+        {
+            ret = -EINVAL;
+        }
     }
 
     if (!(locals = malloc(sizeof(struct locals))))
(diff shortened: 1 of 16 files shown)


jxyang (approved):

kernel/ramfs.c
```diff
@@ -916,8 +916,13 @@ static int _fs_open(
             ramfs, locals->dirname, true, NULL, &parent, NULL, NULL));
 
         /* Create the new file inode */
-        ECHECK(_inode_new(
-            ramfs, parent, locals->basename, (S_IFREG | mode), &inode));
+        /* ATTN: Current ramfs only support S_IFREG or S_ISCHR (virtual /dev) */
```
jxyang, line 919: Are we allowing creating virtual files with `open`?

kernel/chown.c
```diff
@@ -121,10 +123,12 @@ long myst_syscall_fchown(int fd, uid_t owner, gid_t group)
     if (fd < 0)
         return -EINVAL;
 
-    if (((owner != (uid_t)-1) &&
-         (myst_valid_uid_against_passwd_file(owner) < 0)) ||
-        ((group != (gid_t)-1) &&
-         (myst_valid_gid_against_group_file(group) < 0)))
+    /* if not root, check target uid/gid validity */
+    if ((thread->euid != 0) &&
+        (((owner != (uid_t)-1) &&
```
jxyang, line 128: Please write nested if-statements to make the conditions more clear. Also on Linux I do get errors with `sudo chmod xxxx <file>` when `xxxx` is a string and doesn't map to an existing user/group. It seems Ok when `xxxx` is a number.

kernel/ramfs.c
bodzhang: Yes. `myst_create_virtual_file()` use `fs_open()`

PR #983, opened 2021-11-04
Allow the pthread stack size to be configurable

jxyang:
* The default pthread stack size can be configured either with config.json or from "myst" command line.
* Parse the desired pthread stack size from either of the two sources and pass it into the kernel.
* The kernel patches the PT_GNU_STACK header of the CRT elf image to the correct stack size.
* We don't have to patch MUSL as it already has logics to use the bigger value in PT_GNU_STACK or the default 128k.
* Rewrite tests/pthread to verify that the command line route is open.
* Enlarge thread stack size to 2 MB for cpython test suites via config.json.

### DIFF
--- a/crt/enter.c
+++ b/crt/enter.c
@@ -29,7 +29,6 @@
 static myst_wanted_secrets_t* _wanted_secrets;
 
 void _dlstart_c(size_t* sp, size_t* dynv);
-int __pthread_enlarge_default_stack_size(size_t size);
 
 typedef long (*syscall_callback_t)(long n, long params[6]);
 
--- a/crt/enter.c
+++ b/crt/enter.c
@@ -29,6 +29,7 @@
 static myst_wanted_secrets_t* _wanted_secrets;
 
 void _dlstart_c(size_t* sp, size_t* dynv);
+int __pthread_enlarge_default_stack_size(size_t size);
 
 typedef long (*syscall_callback_t)(long n, long params[6]);
 
@@ -167,10 +168,13 @@ void myst_enter_crt(
     void* stack,
     void* dynv,
     syscall_callback_t callback,
-    myst_wanted_secrets_t* wanted_secrets)
+    myst_crt_args_t* args)
 {
     _syscall_callback = callback;
-    _wanted_secrets = wanted_secrets;
+    if (args)
+    {
+        _wanted_secrets = args->wanted_secrets;
+    }
     _dlstart_c((size_t*)stack, (size_t*)dynv);
 }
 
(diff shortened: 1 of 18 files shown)


crt/enter.c
```diff
@@ -29,6 +29,7 @@
 static myst_wanted_secrets_t* _wanted_secrets;
 
 void _dlstart_c(size_t* sp, size_t* dynv);
+int __pthread_enlarge_default_stack_size(size_t size);
```
mikbras, line 32: Is this dead code?

tools/myst/host/exec.c
```diff
@@ -535,6 +535,30 @@ int exec_action(int argc, const char* argv[], const char* envp[])
             }
         }
 
+        /* Get --thread-stack-size */
+        {
+            const char* opt = "--thread-stack-size";
```
mingweishih, line 540: Would need to update the help message accordingly.

mikbras (approved):

kernel/exec.c
```diff
@@ -1012,7 +1018,20 @@ int myst_exec(
...
+    phdr = (Elf64_Phdr*)((const uint8_t*)crt_data + ehdr->e_phoff);
+
+    /* Patch the PT_GNU_STACK entry */
+    {
+        ehdr = crt_data;
+        for (size_t i = 0; i < ehdr->e_phnum; i++)
+        {
+            if (phdr[i].p_type == PT_GNU_STACK)
+            {
+                if (_thread_stack_size > phdr[i].p_memsz)
```
mingweishih, line 1030: what's the default value?

doc/sign-package.md
```diff
@@ -104,6 +104,7 @@ Settings | Type | Description
 ~~UserMemSize~~ | `int \| string` | Deprecated, use `MemorySize` instead
 MemorySize | `int \| string` | Amount of memory your application needs to run. Try not to make this just a very large number as the larger this number needs to be the slower load time will be. Value can be bytes (just a number), **k**ilobytes (for example `"128k"`), **m**egabytes (for example `"512m"`), or **g**igabytes (for example `"1g"`)
 MainStackSize | `int \| string` | Stack size of your application's main process. Defaults to 1536k (or 1.5M) bytes. Normally, you do not need to customize this. If running an application generates a OOM error like in [#612](https://github.com/deislabs/mystikos/issues/612), try tuning this value, e.g. to 8M. Value can be bytes (just a number), **k**ilobytes (for example `"128k"`), **m**egabytes (for example `"512m"`), or **g**igabytes (for example `"1g"`)
+ThreadStackSize | `int \| string` | The default stack size of pthreads created by the application.
```
mingweishih, line 107: The description does not seem to be accurate; the size set by this configuration will only be applied if it's greater than a default value (defined in musl?)

crt/enter.c
jxyang: Thanks for catching that. Removed.

kernel/exec.c
jxyang: 0

mingweishih (approved):

PR #985, opened 2021-11-05
EXPERIMENTAL: Unix-domain sockets (AF_LOCAL and AF_UNIX)

mikbras:
This PR implements Unix-domain sockets, which provide a way for two processes within the same kernel image to communicate securely using host-side sockets, where the kernel internally adds a layer of encryption. See the documentation under [./kernel/udsdev.c](https://github.com/deislabs/mystikos/blob/bed02f53fa5e875e5d4e608fda16685674b59532/kernel/udsdev.c) for details of the algorithm.

This PR also adds the ``myst_syslog()`` function for writing message to the console in different colors. This was added so that warning could be produced when apps attempt to use unsupported Unix-domain socket features.

### DIFF
--- a/include/myst/crypt.h
+++ b/include/myst/crypt.h
@@ -1,0 +1,29 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#ifndef _MYST_CRYPT_H
+#define _MYST_CRYPT_H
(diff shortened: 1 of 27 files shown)


mikbras:
Marked as experimental again due to certain issues.

PR #986, opened 2021-11-05
set kernel-mode thread td properly when forking child

paulcallen:


crt/fork.c
```diff
@@ -30,7 +30,15 @@ static void _set_fsbase(void* p)
 {
 #if 0
```
mikbras, line 31: It's probably safe to remove this now.

crt/fork.c
```diff
@@ -30,7 +30,15 @@ static void _set_fsbase(void* p)
...
-#else
+#endif
+#if 1
+    if (syscall(SYS_set_thread_area, p) < 0)
+    {
+        fprintf(stderr, "syscall(SYS_set_thread_area, p) failed\n");
+        abort();
+    }
+#endif
+#if 0
```
mikbras, line 41: And we could remove this too I think.

mikbras (approved):

PR #989, opened 2021-11-08
Add tcalls sched_setscheduler, sched_getscheduler, sched_getparam, sched_get_priority_max, sched_get_priority_min

RRathna:
To-do:
1. enable ltp tests
 - sched_get_priority_min02/max02 pass. All the other tests fail.
2. Explore security implication of sending out sched_setscheduler call to the host.

### DIFF
--- a/kernel/procfs.c
+++ b/kernel/procfs.c
@@ -27,6 +27,10 @@ static int _stat_vcallback(
     myst_file_t* file,
     myst_buf_t* vbuf,
     const char* entrypath);
+static int _sys_kernel_pid_max_vcallback(
+    myst_file_t* self,
+    myst_buf_t* vbuf,
+    const char* entrypath);
 
 static myst_fs_t* _procfs;
 static char* _cpuinfo_buf = NULL;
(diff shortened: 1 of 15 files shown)


paulcallen (changes requested):
There will need tom be some parameter validation on the structure parameters in the kernel, and need to enable the LTP tests for these if possible.

kernel/procfs.c
```diff
@@ -97,6 +101,25 @@ int procfs_pid_setup(pid_t pid)
         ECHECK(myst_mkdirhier(locals->tmp_path, 777));
     }
 
+    /* Create /proc/sys/kernel directory */
```
vtikoo, line 104: /proc/sys/kernel is not a pid specific entry. `procfs_pid_setup` is called on process creation.

vtikoo: Recommend moving this to `create_proc_root_entries`.

kernel/procfs.c
```diff
@@ -97,6 +101,25 @@ int procfs_pid_setup(pid_t pid)
...
+    {
+        ECHECK(myst_snprintf(
+            locals->tmp_path, sizeof(locals->tmp_path), "/proc/sys/kernel"));
+
+        ECHECK(myst_mkdirhier(locals->tmp_path, 777));
+
+        /* Create /proc/sys/kernel/pid_max entry */
+        ECHECK(myst_snprintf(
```
vtikoo, line 112: You don't need the string formatting, you could directly use the string literal in `myst_create_virtual_file`.

kernel/procfs.c
```diff
@@ -97,6 +101,25 @@ int procfs_pid_setup(pid_t pid)
...
+            locals->tmp_path,
+            sizeof(locals->tmp_path),
+            "/proc/sys/kernel/pid_max"));
```
vtikoo, line 115: This should be a path within /proc mount. So just "/sys/kernel/pid_max".

radhikaj:

PR #991, opened 2021-11-08
added feature to get the envvars from the docker container

salsal97:


kernel/enter.c
```diff
@@ -820,6 +939,104 @@ int myst_enter_kernel(myst_kernel_args_t* args)
     if (__myst_kernel_args.perf)
         _print_boottime();
 
+    // open the appenv.json
+    char* data;
+    size_t size;
+    char* path = "appenv.json";
```
mingweishih: If this is a hard-coded string, can you use `#define`?

kernel/enter.c
```diff
@@ -820,6 +939,104 @@ int myst_enter_kernel(myst_kernel_args_t* args)
...
+    if ((r = json_parser_init(
+             &parser,
+             data,
+             size,
+             _json_read_callback,
+             &callback_data,
+             &allocator,
+             &options)) != JSON_OK)
+    {
+        myst_eprintf("json_parser_init() failed: %d\n", r);
```
mingweishih: Shouldn't we return on failure?

kernel/enter.c
```diff
@@ -820,6 +939,104 @@ int myst_enter_kernel(myst_kernel_args_t* args)
...
+    }
+
+    parser.trace = _trace;
+
+    if ((r = json_parser_parse(&parser)) != JSON_OK)
+    {
+        myst_eprintf("json_parser_parse() failed: %d\n", r);
```
mingweishih: same here

kernel/enter.c
```diff
@@ -820,6 +939,104 @@ int myst_enter_kernel(myst_kernel_args_t* args)
...
+    }
+
+    if (parser.depth != 0)
+    {
+        myst_eprintf("unterminated objects\n");
```
mingweishih: same here

kernel/enter.c
```diff
@@ -653,6 +662,116 @@ static void _print_boottime(void)
...
+    size_t nv_len = strlen(new_envp.data[pos]);
+    char* tuple = malloc(nv_len + 1 + val_len);
+    strncat(tuple, new_envp.data[pos], nv_len);
+    strncat(tuple, ":", 1);
+    strncat(tuple, value, val_len);
+    if (myst_args_remove(&new_envp, pos, 1) != 0)
+    {
+        myst_eprintf("path variable not found in envp");
+    }
+    myst_args_append1(&new_envp, tuple);
```
mingweishih: I think `tuple` can/should be freed after append. Can you double check?

kernel/enter.c
```diff
@@ -653,6 +662,116 @@ static void _print_boottime(void)
...
+
+static void _trace(
+    json_parser_t* parser,
+    const char* file,
+    unsigned int line,
+    const char* func,
+    const char* message)
+{
+    (void)parser;
+    printf("%s(%u): %s(): %s\n", file, line, func, message);
```
mingweishih: To be safe, check pointers are not NULL before printf

kernel/enter.c
```diff
@@ -820,6 +939,104 @@ int myst_enter_kernel(myst_kernel_args_t* args)
     if (__myst_kernel_args.perf)
         _print_boottime();
 
+    // open the appenv.json
+    char* data;
+    size_t size;
+    char* path = "appenv.json";
+    if (myst_load_file(path, (void**)&data, &size) != 0)
+    {
+        goto skip_docker_env;
```
mingweishih: Instead of adding a goto flow, can you put these code (942 - 1038) in a function? That is, the function can do early return if myst_load_file() != 0

kernel/enter.c
```diff
@@ -653,6 +662,116 @@ static void _print_boottime(void)
...
 static myst_thread_t _main_thread;
 
+static void concat_env_variables(
+    const char* value,
+    myst_args_t new_envp,
+    int pos)
+{
+    size_t val_len = strlen(value);
+    size_t nv_len = strlen(new_envp.data[pos]);
+    char* tuple = malloc(nv_len + 1 + val_len);
```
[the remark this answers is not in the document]
salsal97: Added this check, thank you

kernel/enter.c
```diff
@@ -653,6 +662,116 @@ static void _print_boottime(void)
...
+    strncat(tuple, new_envp.data[pos], nv_len);
+    strncat(tuple, ":", 1);
+    strncat(tuple, value, val_len);
```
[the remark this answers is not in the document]
salsal97: Changed this implementation to use asprintf, thank you for the suggestion

kernel/enter.c
salsal97: Added an ERAISE to propogate the failure

kernel/enter.c
salsal97: Added ERAISE

kernel/enter.c
salsal97: Added ERAISE

kernel/enter.c
salsal97: looking into the frees

kernel/enter.c
salsal97: Implemented this as a different function, thank you for the suggestion

kernel/enter.c
salsal97: added this, thank you

kernel/appenv.c
```diff
@@ -0,0 +1,252 @@
...
+            }
+            free(nv);
+        }
+    }
+
+    myst_args_release(&callback_data.env);
+
+    args->envc = new_envc;
+    args->envp = new_envp.data;
+
```
jxyang, line 245: Free the original arguments here.

kernel/appenv.c
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /* store the appenv variables in the env vector according to rules */
+    for (size_t i = 0; i < callback_data.env.size; i++)
+    {
+        char* nv = NULL;
+        if (asprintf(&nv, "%s", callback_data.env.data[i]) == -1)
+            ERAISE(-ENOMEM);
+
+        const char* value = strchr(nv, '=');
```
jxyang, line 207: Suggest a better variable name such as `index`

kernel/appenv.c
```diff
@@ -0,0 +1,252 @@
...
+
+        if (myst_args_append1(&new_envp, nv) != 0)
+            ERAISE(-EINVAL);
+    }
+
+    /* store the appenv variables in the env vector according to rules */
+    for (size_t i = 0; i < callback_data.env.size; i++)
+    {
+        char* nv = NULL;
+        if (asprintf(&nv, "%s", callback_data.env.data[i]) == -1)
```
jxyang, line 204: Why do we need to make a copy of the string?

kernel/appenv.c
```diff
@@ -0,0 +1,252 @@
...
+        {
+            if (myst_args_append1(&new_envp, nv) != 0)
+                ERAISE(-EINVAL);
+            new_envc++;
+        }
+        else
+        {
+            /* if the environment variable is already defined, skip or concat to
+             * its current value if it is PATH or LD_LIBRARY_PATH */
+            if (strncmp(new_envp.data[pos], "PATH=", 5) == 0)
```
jxyang, line 229: We can combine this if-statement and the next one. Also try not to use literal integer for literal string lengths otherwise we will easily screw up when later we modify the string but forget the length. The better way is:

```c
const char* pattern = "....";
if (strncmp(longer_string, pattern, strlen(pattern)) == 0)
   ...
```

kernel/enter.c
```diff
@@ -938,6 +948,16 @@ int myst_enter_kernel(myst_kernel_args_t* args)
         free(process);
         process = NULL;
 
+        if (create_appenv_ret == 0)
```
jxyang, line 951: Suggest doing the clean up in `myst_create_appenv` instead (only when successful)

kernel/enter.c
```diff
@@ -663,6 +665,7 @@ int myst_enter_kernel(myst_kernel_args_t* args)
     myst_process_t* process = NULL;
     myst_fstype_t fstype;
     int tmp_ret;
+    int create_appenv_ret;
```
jxyang, line 668: I think you can reuse `tmp_ret` instead of creating a new variable.

kernel/appenv.c
```diff
@@ -0,0 +1,252 @@
...
+    myst_args_t* new_envp,
+    int pos)
+{
+    char* tuple;
+    const char* curr = new_envp->data[pos];
+
+    if (!curr)
+        return -EINVAL;
+
+    /* Append the APPENV_FILENAME value of the var to the config.json value */
```
jxyang, line 34: The comment about `APPENV_FILENAME` and `config.json` is too specific.

kernel/appenv.c
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    result = JSON_OK;
+
+done:
+    return result;
+}
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstack-usage="
```
jxyang, line 126: Why the stack size warning has to be turned off? Any huge stack variable should be allocated on the heap using the `locals` structure. 

This function is only called once during the entry of kernel so maybe it's fine to keep the variable on the stack. But I am curious which variable is costing the stack.

kernel/appenv.c
salsal97: Thank you for the suggestion, added

kernel/enter.c
salsal97: I think it might be better to use a separate variable in case the value of tmp_ret is altered by some other part of the code

kernel/appenv.c
salsal97: I agree, modified it a little

kernel/enter.c
salsal97: We use the value after the call to that function, so the free needs to be after the exec statement in enter.c

kernel/appenv.c
salsal97: Was able to turn this off, thank you for your help with the suggestions

jxyang (approved):

Issue #990, opened 2021-11-08
Only first SIGSEGV is handled for dotnet program with multiple SIGSEGVs (Linux target only)


vtikoo:
If a dotnet program throws multiple SIGSEGV's, only the first SIGSEGV is delivered by Mystikos. The second SIGSEGV causes the host process to abort.

This scenario comes up in a dotnet library test - https://github.com/dotnet/runtime/blob/v5.0.0-rtm.20519.4/src/libraries/System.ValueTuple/tests/ValueTupleTests.cs

Repro program -

```C#
using System;
using System.Diagnostics;

namespace hello
{
    class Program
    {
        static void nullstrlen(string s)
        {
            try {
                int len = s.Length;
            } catch (NullReferenceException e) {
                Console.WriteLine("Exception caught: {0}", e);
            }
        }
        static void Main(string[] args)
        {
            string s = null;
            nullstrlen(s);

            string s1 = null;
            nullstrlen(s1);

            // unreachable
            //Environment.Exit(1);
        }  
    }
}

```

Output on Mystikos - 
```
_handle_one_signal(11 SIGSEGV) pid=101 tid=20992 <-- first SIGSEGV
Exception caught: System.NullReferenceException: Object reference not set to an instance of an object.
   at hello.Program.nullstrlen(String s) in /src/hello/Program.cs:line 11
Segmentation fault (core dumped) <-- second SIGSEGV causes segfault and core dump on host
Makefile:26: recipe for target 'tests' failed
make: *** [tests] Error 139
```

mingweishih:
I'm wondering after the first SIGSEGV if the execution ever returns and trigger the second SIGSEGV.
Does this test run in package mode?

vtikoo:
Following C program causes SIGSEGVs to be sent from the signal handler functions twice before exiting - 
This behaves differently on SGX target vs host Linux. Linux target matches host Linux behavior.

```C
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

#include <signal.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

void one()
{
    printf("one\n");
    int* p = NULL;
    int b = *p;
}

void two()
{
    printf("two\n");
    int* p = NULL;
    int b = *p;
}

int segv_count = 0;
static void _sigsegv_handler(int sig, siginfo_t *si, void *ctxt)
{
    if (segv_count == 0)
    {
        segv_count++;
        one();
    }
    if (segv_count == 1)
[...]

PR #993, opened 2021-11-08
Linux target: Unmask signal on host before delivering it to Mystikos …

vtikoo:
…kernel

For scenarios where the signal handler may never return,
and same signal may be generated again(for eg: dotnet runtime).

### DIFF
--- a/tools/myst/host/exec_linux.c
+++ b/tools/myst/host/exec_linux.c
@@ -318,6 +318,10 @@ static void _install_signal_handlers()
 {
     struct sigaction sa;
     sa.sa_flags = SA_SIGINFO;
+    // Specify SA_NODEFER, so that if same signal is thrown from the handler, it
+    // can be caught. This is encountered in .net runtime, where the signal
+    // handler never returns.
+    sa.sa_flags |= SA_NODEFER;
     sigemptyset(&sa.sa_mask);
     sa.sa_sigaction = _sigaction_handler;
 
--- a/tests/dotnet-ubuntu/hello/Program.cs
+++ b/tests/dotnet-ubuntu/hello/Program.cs
@@ -5,18 +5,34 @@ namespace hello
 {
     class Program
     {
-        static void Main(string[] args)
+        static void test_divide_by_zero()
         {
-            Console.WriteLine("Hello World!");
             int q, a = 5, b = 0;
             try {
                 q = a / b;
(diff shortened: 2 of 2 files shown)


tools/myst/host/exec_linux.c
```diff
@@ -309,6 +309,15 @@ void cleanup_alt_stack()
 
 static void _sigaction_handler(int sig, siginfo_t* si, void* context)
 {
+    /* Unblock signal */
+    sigset_t mask;
+    sigemptyset(&mask);
+    sigaddset(&mask, sig);
+
+    int ret = pthread_sigmask(SIG_UNBLOCK, &mask, NULL);
```
jxyang, line 317: Not sure I get this correctly: if we are here, that means the signal has already been delivered to *this* thread, that implied the mask is off for this thread. Why unmask again?

jxyang: Is this for the 2nd firing of SIGSEGV? We are supposed to turn on the mask for SIGSEGV when handing SIGSEGV, and turn it off when finished handling.  We need to understand why the mask stays on after the handler is done.

vtikoo: We are unblocking the signal for this thread. So assuming the signal handler doesn't return, and there are more instances of this signal; we are able to deliver those signals.

jxyang: I am still not sure why the signal handler didn't return from the 1st SIGSEGV. But if we want nested handling of SIGSEGV, we could use SA_NODEFER while calling `sigaction`

vtikoo: From dotnet team engineer:

>  the SIGSEGV handler never returns. We always start the exception handling directly from the handler. To execute the catch handler, we set the CPU context using our the RtlRestoreContext function.

jxyang: I am not sure about the "never returns". We only have one thread in the test case, if the signal handler never returns, the test should hang, right?

It should be easy to find out by debugging `myst_handle_host_signal`

jxyang: Thanks @vtikoo  for verifying this. The exception handling on Linux indeed doesn't return.

jxyang (approved):

jxyang (approved):

vtikoo:
Fixes #990

PR #994, opened 2021-11-09
Add execveat syscall

asvrada:
Add syscall execveat and fix myst_get_absolute_path_from_dirfd

### DIFF
--- a/kernel/chown.c
+++ b/kernel/chown.c
@@ -229,14 +229,15 @@ long myst_syscall_fchownat(
     if ((flags & ~(AT_EMPTY_PATH | AT_SYMLINK_NOFOLLOW)) != 0)
         ERAISE(-EINVAL);
 
-    ECHECK(myst_get_absolute_path_from_dirfd(dirfd, pathname, flags, &abspath));
-
-    if (*abspath == '\0')
+    if (*pathname == '\0' && (flags & AT_EMPTY_PATH) && dirfd != AT_FDCWD)
     {
         ECHECK(myst_syscall_fchown(dirfd, owner, group));
     }
     else
     {
+        ECHECK(myst_get_absolute_path_from_dirfd(
+            dirfd, pathname, flags, &abspath, FB_PATH_NOT_EMPTY));
+
         if (flags & AT_SYMLINK_NOFOLLOW)
         {
             ECHECK(myst_syscall_lchown(abspath, owner, group));
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -504,50 +504,50 @@ done:
     return ret;
 }
 
-// Given a dir fd and a pathname, return a concatenated absolute path.
-// This function allocates buffer for the concatenated abspath. The
-// caller doesn't need to preallocate buffer, but should free the
-// abspath buffer after using it.
-//
-// Caveats:
-//
-// - Sometimes abspath is set to equal to
-//   pathname, in which case no free is needed. The caller should check
-//   *abspath_out before free:
-//   e.g.
-//       if (*abspath_out != pathname)
-//           free(*abspath_out)
-//
-// - If this function hits an error and doesn't return SUCCESS, it will
-//   free the abspath buffer by itself. Also, *abspath_out == NULL.
-//   Thus, the check above is still valid.
-//
-// - Some flags allow the pathname to be an empty string. Caller
-//   passing in those flags should check when *abspath == '\0'. If
-//   true, then the caller should apply the fd version of the syscall
-//   using dirfd.
+/*
+Given a dirfd and a pathname, return a concatenated absolute path.
+
+Caveats:
+
+- Sometimes abspath is set to equal to
+  pathname, in which case no free is needed. The caller should check
+  *abspath_out before free:
+  e.g.
+      if (*abspath_out != pathname)
+          free(*abspath_out)
+
+Arguments:
+ - abspath_out: will point to the final concatenated absolute path.
+    This method will allocate memory for abspath_out if necessary
+    Caller should free the memory if abspath_out != pathname
+ - flags_behavior: a bit mask that can be one or more of the following flags
+FB_PATH_NOT_EMPTY: indicate pathname cannot be empty and type check on dirfd is
+skipped
+FB_TYPE_FILE: indicate file pointed by dirfd can be file
+FB_TYPE_DIRECTORY: indicate file pointed by dirfd can be directory
+FB_THROW_ERROR_NOFOLLOW: if set, throw ELOOP when abspath is a symlink
+*/
 long myst_get_absolute_path_from_dirfd(
     int dirfd,
     const char* pathname,
     int flags,
-    char** abspath_out)
+    char** abspath_out,
+    const int flags_behavior)
 {
     long ret = 0;
+    myst_path_t* resolved_path = NULL;
     char* path_out = NULL;
-    struct locals
-    {
-        char dirname[PATH_MAX];
-    }* locals = NULL;
 
     if (!pathname || !abspath_out)
         ERAISE(-EINVAL);
 
-    /* If pathname is absolute, then ignore dirfd */
-    if (*pathname == '/' || dirfd == AT_FDCWD)
+    // Absolute pathname or AT_FDCWD
+    if (pathname[0] == '/' || dirfd == AT_FDCWD)
     {
-        *abspath_out = (char*)pathname;
+        path_out = (char*)pathname;
     }
-    else if (*pathname == '\0')
+    // Empty pathname
+    else if (pathname[0] == '\0')
     {
         if (!(flags & AT_EMPTY_PATH))
             ERAISE(-ENOENT);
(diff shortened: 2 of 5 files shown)


tests/sys_execve/src/test_sys_execveat.c
```diff
@@ -0,0 +1,376 @@
...
+    syscall(SYS_execveat, dirfd, pathname, argv, 0, AT_SYMLINK_NOFOLLOW);
+#else
+    printf("Skipping "
+           "test_execveat_symlink_in_path_and_symlink_nofollow_flag_then_ok\n");
+    exit(0);
+#endif
+}
+
+// last component of pathname is symlink
+void test_execveat_symlink_end_path_and_symlink_nofollow_flag_then_eloop()
```
jxyang, line 291: I have never seen a function name that is so descriptive. :-)

In practice, keep the name short yet descriptive. If you need to explain more what does it do, write a comment above the function declaration.

kernel/syscall.c
```diff
@@ -2080,32 +2080,127 @@ static size_t _count_args(const char* const args[])
...
-    const char* resolved_filename = filename;
+    char* resolved_pathname = NULL;
 
-    if (!resolved_filename)
+    if (!pathname || (flags & ~(AT_SYMLINK_NOFOLLOW | AT_EMPTY_PATH)))
         ERAISE(-EINVAL);
 
-    // Resolve relative path
-    if (resolved_filename[0] != '/')
+    if (!(resolved_pathname = (char*)malloc(PATH_MAX)))
```
jxyang, line 2103: We have a helper function `myst_get_absolute_path_from_dirfd` that resolves `dirfd + pathname` to the real path. No need to repeat in this function.

kernel/syscall.c
```diff
@@ -3917,9 +4025,9 @@ static long _syscall(void* args_)
             /* the kstack is freed later by next exec or by exit */
             thread->exec_kstack = args->kstack;
 
-            long ret = myst_syscall_execve(filename, argv, envp);
-            /* myst_syscall_execve() only returns on failure */
-            /* when myst_syscall_execve() returns on failure, kstack will be
+            long ret = myst_syscall_execveat(AT_FDCWD, filename, argv, envp, 0);
+            /* myst_syscall_execveat() only returns on failure */
```
jxyang, line 4029: I would keep both `myst_syscall_execve` and `myst_syscall_execveat`, with the latter calling `myst_get_absolute_path_from_dirfd` and then `myst_syscall_execve`.

tests/sys_execve/src/test_sys_execveat.c
```diff
@@ -0,0 +1,376 @@
+// Copyright (c) Microsoft Corporation.
```
jxyang, line 1: Why rename the file? It still essentially tests `execv` (`execvat` is just a derivation). Additionally, renaming the file make it much harder to diff with the previous version.

kernel/syscall.c
asvrada: I think it's more reasonable to only use myst_syscall_execveat, since execve is one of the special case in execveat.

tests/sys_execve/src/test_sys_execveat.c
asvrada: Thanks. Will keep the current names as is, but will use better naming in the future :)

tests/sys_execve/src/test_sys_execveat.c
asvrada: Will revert this

kernel/syscall.c
asvrada: I tried to reuse `myst_get_absolute_path_from_dirfd` but I found it more confusing/repetitive to use that and add additional logic to make it work, rather than start fresh. 

> For example, when the pathname is empty, `myst_get_absolute_path_from_dirfd` would behave differently if flag AT_SYMLINK_NOFOLLOW is set, and won't check if `dirfd` points to a directory, while `execveat` doesn't check the flag at this stage, and needs to check if `dirfd` points to a directory.

jxyang: I think the original implementation of `AT_SYMLINK_NOFOLLOW` is wrong in `myst_get_absolute_path_from_dirfd`. We should fix it instead of working around it. The man page says: 
> If pathname is a symbolic link, do not dereference it: instead operate on the link itself.

This should be consistent across all xxxat syscalls including fchmodat and exeaveat.

kernel/syscall.c
jxyang: I think it's the other way around. This applies to virtually all syscalls with a "at" variant, such as openat, chmodat, etc.

asvrada:
Before settling with the current implementation, I tried to use existing `myst_get_absolute_path_from_dirfd` and follow other familiar `_at` syscalls, the code is supposed to look like this:

```
long myst_syscall_execveat(...) {
...
    myst_get_absolute_path_from_dirfd(dirfd, pathname, flag, abspath_out);
    myst_syscall_execve(abspath_out, xxx);
...
}
```

This could work, but (to my best knowledge, please correct me if I'm wrong) not without many, many additional logic to address the difference in behavior between `execveat` and other `_at` syscalls.

kernel/syscall.c
asvrada: Please see my new comment.

kernel/syscall.c
asvrada: 😢 So I did more reading and still think the current implementation is the most suitable... Here's my reasoning

1. __The original implementation of AT_SYMLINK_NOFOLLOW is wrong in `myst_get_absolute_path_from_dirfd`__

I couldn't figure out what part is wrong, possibly because I don't have enough context on this function (like what's its expected behavior). Could you create a new issue to describe that?

2. __Should use `myst_get_absolute_path_from_dirfd` in execveat__

Please see my new comment

tests/libcxx/Makefile
```diff
@@ -53,7 +53,7 @@ $(APPDIR):
 $(ROOTFS): run.c
 	$(MYST) mkext2 $(APPDIR) $(ROOTFS)
 
-OPTS += --memory-size=64m
+OPTS += --memory-size=128m
```
salsal97, line 56: why was this change necessary? Is it intentional?

asvrada: With 64mb and this code change, these test would fail with memory allocation issue.

asvrada:
After discussion with Xuejun, we agreed to change the code to
1. Fix `myst_get_absolute_path_from_dirfd` to have the following behavior
    1. For absolute pathname, return itself
    2. For empty pathname, return `abspath(dirfd)` (dirfd should not point to directory)
    3. For relative pathname, return `abspath(dirfd) + pathname` (dirfd should ONLY point to directory)
    4. After getting abspath, check flag `AT_SYMLINK_NOFOLLOW`, and act accordingly (For example, throw ELOOP for `execveat`) 
2. Call `execve` inside `execveat`

kernel/syscall.c
jxyang: Talked to Zijie offline. We found some problems with the current `myst_get_absolute_path_from_dirfd` implementation:

* The name is wrong, the function returns a relative path when dirfd is the current directory
* The checking of `AT_SYMLINK_NOFOLLOW` should be done regardless `pathname` is null or not.
* It should add additional parameter to indicate whether to throw an error if `dirfd + pathname` is a symbolic link and `AT_SYMLINK_NOFOLLOW`  flag is specified. 

I think we should address these problems in this PR rather than creating a clone of the original and later on removing it.

asvrada:
Will replace old method and usage of that method in another commit

kernel/syscall.c
```diff
@@ -504,102 +504,108 @@ long myst_syscall_open(const char* pathname, int flags, mode_t mode)
...
+    char* const abspath_out,
+    const int flags_behavior)
 {
-    long ret = 0;
-    char* path_out = NULL;
-    struct locals
-    {
-        char dirname[PATH_MAX];
-    }* locals = NULL;
+    int ret = 0;
```
asvrada, line 530: Will use `long`

kernel/chown.c
```diff
@@ -229,14 +229,18 @@ long myst_syscall_fchownat(
     if ((flags & ~(AT_EMPTY_PATH | AT_SYMLINK_NOFOLLOW)) != 0)
         ERAISE(-EINVAL);
 
-    ECHECK(myst_get_absolute_path_from_dirfd(dirfd, pathname, flags, &abspath));
```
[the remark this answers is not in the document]
jxyang: While your changes to `myst_get_absolute_path_from_dirfd` easier to follow, but:

1. it forces the caller to allocate memory first, which resulted in a lot of changes to the many callers below. When a function is called many times, we tend to add complex logics to it rather than the call sites.
2. Some memory allocations could be otherwise saved now become mandatory.

kernel/syscall.c
```diff
@@ -5804,7 +5856,41 @@ static long _syscall(void* args_)
...
+
+            /* free the previous kernel stack from the SYS_execve syscall */
+            if (thread->exec_kstack)
+            {
+                myst_put_kstack(thread->exec_kstack);
+                thread->exec_kstack = NULL;
+            }
+
+            /* the kstack is freed later by next exec or by exit */
+            thread->exec_kstack = args->kstack;
```
jxyang, line 5883: `SYS_execveat` and `SYS_execve` both handles kstacks the same way before and after the call to `myst_syscall_execveat`. Consider moving the handling code into `myst_syscall_execveat` to avoid duplication.

kernel/syscall.c
```diff
@@ -504,102 +504,111 @@ long myst_syscall_open(const char* pathname, int flags, mode_t mode)
...
-//   Thus, the check above is still valid.
-//
-// - Some flags allow the pathname to be an empty string. Caller
-//   passing in those flags should check when *abspath == '\0'. If
-//   true, then the caller should apply the fd version of the syscall
-//   using dirfd.
+/*
+Given a dirfd and a pathname, return a concatenated absolute path.
+
+This function DOES NOT allocate buffer for the concatenated abspath.
```
jxyang, line 510: This comment is outdated.

kernel/syscall.c
```diff
@@ -608,24 +617,50 @@ long myst_get_absolute_path_from_dirfd(
                 ERAISE(-ENOTDIR);
         }
 
-        /* get the full path of dirfd */
-        ECHECK((*fs->fs_realpath)(
-            fs, file, locals->dirname, sizeof(locals->dirname)));
+        if (!(buf = (char*)malloc(PATH_MAX)))
+            ERAISE(-ENOMEM);
+        // get the full path of dirfd
+        ECHECK((*fs->fs_realpath)(fs, file, buf, PATH_MAX));
```
jxyang, line 623: We don't need the temporary `buf`. We can call `fs_realpath` on `path_out` and then `strcpy` or `memcpy` the `pathname`

kernel/syscall.c
```diff
@@ -655,7 +690,8 @@ static long _openat(
     if (!(locals = malloc(sizeof(struct locals))))
         ERAISE(-ENOMEM);
 
-    ECHECK(myst_get_absolute_path_from_dirfd(dirfd, pathname, 0, &abspath));
+    ECHECK(myst_get_absolute_path_from_dirfd(dirfd, pathname, 0, &abspath, 0));
```
jxyang, line 693: Please use the defined macros instead of `0` as the last parameter. Same applies to the following myst_get_absolute_path_from_dirfd calls.

jxyang (approved):

PR #996, opened 2021-11-09
tests/dotnet-sos fix

vtikoo:
- Explicitly copy data access module(libmscordaccore.so) to /tmp symbol files path. This is required for SOS to function.
- Base prereqs image on Ubuntu 18.04.

### DIFF
--- a/tests/dotnet-sos/Dockerfile-prereqs
+++ b/tests/dotnet-sos/Dockerfile-prereqs
@@ -1,6 +1,35 @@
-FROM mcr.microsoft.com/dotnet/sdk:5.0 AS build
+FROM ubuntu:18.04
+ARG PSW_VERSION
+
+RUN apt-get update && apt-get install -y \
+    wget \
+    gnupg
+
+RUN if [ -z "$PSW_VERSION" ]; then echo "Please set PSW_VERSION (e.g. 2.15.100)." >&2; exit 1; fi
 
 WORKDIR /tmp
+
+# Use the APT preference file to pin sgx packages to specific versions
+# Reference https://manpages.debian.org/buster/apt/apt_preferences.5.en.html
+# Download the pref file from https://download.01.org/intel-sgx/sgx_repo/ubuntu/apt_preference_files/
+# Assuming file name to follow *sgx_<PSW_VERSION>_bionic_custom_version.cfg convention
+RUN ["/bin/bash", "-c", "wget -r -l1 --no-parent -nd -A *sgx_$(echo ${PSW_VERSION//./_})_bionic_custom_version.cfg https://download.01.org/intel-sgx/sgx_repo/ubuntu/apt_preference_files/"]
+RUN ["/bin/bash", "-c", "mv *sgx_$(echo ${PSW_VERSION//./_})_bionic_custom_version.cfg /etc/apt/preferences.d/intel-sgx.pref"]
+
+# Add the repository to sources, and add the key to the list of
+# trusted keys used by the apt to authenticate packages
+RUN echo "deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu bionic main" | tee /etc/apt/sources.list.d/intel-sgx.list \
+    && wget -qO - https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | apt-key add -
+# Add Microsoft repo for az-dcap-client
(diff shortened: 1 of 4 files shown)


mikbras (approved):

tests/dotnet-sos/Dockerfile-prereqs
```diff
@@ -1,20 +1,55 @@
...
+    az-dcap-client
+
 # Install lldb-10
 RUN apt update && \
      apt install -y lsb-release software-properties-common gnupg
 RUN wget https://apt.llvm.org/llvm.sh && \
         chmod +x llvm.sh && \
         ./llvm.sh 10
 
+RUN wget https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb && \
```
CyanDevs, line 40: Is this the same as installing the Microsoft apt repo (which is done earlier in line 24/25)?

tests/dotnet-sos/Dockerfile-prereqs
```diff
@@ -1,20 +1,55 @@
...
+    dpkg -i packages-microsoft-prod.deb && \
+    rm packages-microsoft-prod.deb
+
+#RUN apt update \
```
CyanDevs, line 44: Commented out on purpose or kept for testing? If not needed can it be removed?

tests/dotnet-sos/Dockerfile-prereqs
```diff
@@ -1,20 +1,55 @@
...
+#    apt install -y apt-transport-https
+RUN apt update && \
```
CyanDevs, line 46: Would suggest running all apt update & install commands in a single layer so as to minimize the amount of layers / image size (and slightly faster builds!)

tests/dotnet-sos/Makefile
```diff
@@ -26,12 +29,13 @@ runner-image:
...
 clean:
-	sudo rm -fr appdir ext2fs rootfs stdouterr.txt
+	sudo rm -fr appdir ext2fs rootfs stdouterr.txt libmscordaccore.so
 
 DOCKER_SGX_DRIVER_OPTIONS = -v /dev/sgx:/dev/sgx
 DOCKER_SGX_DRIVER_OPTIONS += --device /dev/sgx/enclave:/dev/sgx/enclave
 DOCKER_SGX_DRIVER_OPTIONS += --device /dev/sgx/provision:/dev/sgx/provision
-
+DOCKER_SGX_DRIVER_OPTIONS += --volume /var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket
+DOCKER_SGX_DRIVER_OPTIONS += --env SGX_AESM_ADDR=1
```
CyanDevs, line 38: If you have the ENV directive in your Dockerfile you won't actually need this

tests/dotnet-sos/sos_test.py
```diff
@@ -3,32 +3,62 @@
...
 
+def copy_libmscordaccore(ci, res):
+    ci.HandleCommand("finish", res)
+    assert(res.Succeeded())
+
+    ci.HandleCommand("p tmpdir", res)
+    assert(res.Succeeded())
+    tmpdir_match = re.search('"(.*)"', res.GetOutput())    
+    assert(tmpdir_match)
+    tmpdir_path = tmpdir_match.group(0)[1:-1]
```
CyanDevs, line 18: Assuming tmpdir_match returns a path, is this trying to get the path of the tmpdir? If that's the case it seems more readable to use os.path.basedir

tests/dotnet-sos/sos_test.py
```diff
@@ -3,32 +3,62 @@
...
+
+    bashCommand = f"cp libmscordaccore.so {tmpdir_path}"
```
CyanDevs, line 20: Since f-strings is a feature on Python 3 - do we need to keep Python 2 support?

tests/dotnet-sos/Dockerfile-prereqs
```diff
@@ -1,20 +1,55 @@
-FROM mcr.microsoft.com/dotnet/sdk:5.0 AS build
+FROM ubuntu:18.04
```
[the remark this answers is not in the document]
vtikoo: Thanks for sharing that. Have updated to the OE base image.

tests/dotnet-sos/Dockerfile-prereqs
vtikoo: I realized that too, removed the duplicate.

PR #1002, opened 2021-11-10
Dup no longer inherits FD_CLOEXEC

### LINKED ISSUE #503: filesystem dup() needs to have different flag on duplicate than the original
When a filesystem handle is duplicated it uses a reference count on the whole object and so both copies of the handle can be used interchangeably. This is the defined behavior of dup().
Unfortunately there is one omission in the current implementation to do with the close-on-exec (FD_CLOEXEC) flag. This flag should be turned off in the duplicated file handle.
The current file object can stay mostly the same, but it needs to be wrapped by another object that points to the original refcounted object. Then a new field in the outer layer needs to track if the FS_CLOEXEC flag is set or not, such that in dup this flag can be reset.
This needs to be fixed in all three filesystems (ramfs, ext2 and hostfs)

Sahakait:
closes #503 

Moved shared file objects to `myst_file_shared` in ramfs and ext2fs. `myst_file` is now a file descriptor level object, allowing each dup'd file descriptor to have its own file descriptor flag (currently only FD_CLOEXEC).

### DIFF
--- a/ext2/ext2.c
+++ b/ext2/ext2.c
@@ -45,7 +45,7 @@ struct ext2_dir
 
 #define FILE_MAGIC 0x0e6fc76762264945
 
-struct myst_file
+struct myst_file_shared
 {
     uint64_t magic;
     ext2_ino_t ino;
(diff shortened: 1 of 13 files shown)


paulcallen (approved):

paulcallen:
looks good. Thanks for adding the test.

PR #1011, opened 2021-11-11
Fix a couple of PTY bugs

jxyang:

### DIFF
--- a/kernel/ramfs.c
+++ b/kernel/ramfs.c
@@ -2326,6 +2326,15 @@ static int _fs_ioctl(
             ret = devfs_get_pts_id(file, id);
             break;
         }
+        case TCGETS:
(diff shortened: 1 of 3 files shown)


kernel/ramfs.c
```diff
@@ -2326,6 +2326,15 @@ static int _fs_ioctl(
             ret = devfs_get_pts_id(file, id);
             break;
         }
+        case TCGETS:
+        case TCSETS:
+        case TCSETS + 1:
```
mingweishih, line 2331: any reason why not using TCSETSW / TCSETF directly?

jxyang: I don't think those are defined in MUSL.

mingweishih (approved):

PR #1013, opened 2021-11-12
[do not merge] Add /proc/sys/kernel/pid_max

RRathna:
The file /proc/sys/kernel/pid_max is required for ltp tests for syscall sched_getparam, sched_getscheduler and sched_setscheduler to function correctly. 

To cycle pids, introduced a 4096 byte bitmap to track pid/tids in use. The pid/ tid are assigned starting from 100 in increments of 1, and set the bit corresponding to the pid/tid on the bitmap. Concurrently, when a thread has finished execution, the bit corresponding to pid/ tid is unset in bitmap. When pid_max is reached, the bitmap is checked for the first unset bit starting from value 104.

### DIFF
--- a/include/myst/thread.h
+++ b/include/myst/thread.h
@@ -29,6 +29,13 @@
 /* Typical default value in host operating system's /proc/sys/kernel/pid_max */
 #define MYST_PID_MAX 0x8000
 
+/*
+ * Bitmap to track pids in use. If bit 'n' (starting from 0) is set, then pid
+ * 'n' is assigned to an active process or thread. When cycling, we cycling pid
+ * between [104, 32677].
+ */
+extern uint8_t myst_bitmap_pid[4096];
+
 /* this signal is used to interrupt threads blocking on host in syscalls */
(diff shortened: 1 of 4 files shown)


include/myst/thread.h
```diff
@@ -29,6 +29,13 @@
 /* Typical default value in host operating system's /proc/sys/kernel/pid_max */
 #define MYST_PID_MAX 0x8000
 
+/*
+ * Bitmap to track pids in use. If bit 'n' (starting from 0) is set, then pid
+ * 'n' is assigned to an active process or thread. When cycling, we cycling pid
+ * between [104, 32677].
+ */
+MYST_UNUSED static uint8_t bitmap_pid[4096] = {0};
```
paulcallen, line 37: this is wrong. This should be a global variable in something like thread.c itself, not defined in the header. This is adding this arrary to every source file that includes thread.h which is cauing a lot of bloat.
By all means make this an extern variable in the header, but only define it once.

paulcallen (changes requested):

paulcallen (approved):

radhikaj:

PR #1014, opened 2021-11-12
Add tcall sched_getparam

RRathna:
To test /ltp/testcases/kernel/syscalls/sched_getparam/sched_getparam03 passes with,the file /proc/sys/kernel/pid_max, when added. But that introduces pid/tid cycling and the requirements/need has to be clarified - tracked by #1013

### DIFF
--- a/include/myst/thread.h
+++ b/include/myst/thread.h
@@ -25,6 +25,9 @@
 #define MYST_THREAD_MAGIC 0xc79c53d9ad134ad4
 #define MYST_MAX_MUNNAP_ON_EXIT 5
 
+/* Typical default value in host operating system's /proc/sys/kernel/pid_max */
+#define MYST_PID_MAX 0x8000
+
 /* this signal is used to interrupt threads blocking on host in syscalls */
 #define MYST_INTERRUPT_THREAD_SIGNAL (SIGRTMIN + 1)
 
--- a/kernel/sched.c
+++ b/kernel/sched.c
@@ -1,0 +1,49 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include <errno.h>
+#include <sched.h>
+#include <string.h>
+#include <syscall.h>
+#include <unistd.h>
+
+#include <myst/eraise.h>
+#include <myst/errno.h>
+#include <myst/kernel.h>
+#include <myst/thread.h>
+
+long myst_syscall_sched_getparam(pid_t pid, struct sched_param* param)
+{
+    long ret = 0;
+    long params[6] = {0};
+
+    if (pid < 0)
+        ERAISE(-EINVAL);
+    else if (pid >= MYST_PID_MAX)
+        ERAISE(-ESRCH);
+
+    if (!param || !myst_is_addr_within_kernel(param))
+    {
+        ERAISE(-EFAULT);
+    }
+
+    memset(param, 0, sizeof(struct sched_param));
+
+    if (pid == 0)
+        params[0] = (long)pid;
+    else
+    {
+        /* We need to send the target pid to the tcall */
+        myst_process_t* process = myst_find_process_from_pid(pid, true);
+        if (process)
+            params[0] = (long)process->main_process_thread->target_tid;
+        else
+            ERAISE(-ESRCH);
+    }
+    params[1] = (long)param;
+
+    ret = myst_tcall(SYS_sched_getparam, params);
+
+done:
+    return ret;
+}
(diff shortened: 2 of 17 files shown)


include/myst/thread.h
```diff
@@ -24,6 +24,7 @@
 
 #define MYST_THREAD_MAGIC 0xc79c53d9ad134ad4
 #define MYST_MAX_MUNNAP_ON_EXIT 5
+#define MYST_PID_MAX 0x8000
```
paulcallen, line 27: where did this magic number come from? Is this really the case? If more than this number of processes is created will it fail?

kernel/sched.c
```diff
@@ -0,0 +1,46 @@
...
+
+    if (pid < 0)
+        ERAISE(-EINVAL);
+    else if (pid >= MYST_PID_MAX)
+        ERAISE(-ESRCH);
+
+    if (param && !myst_is_addr_within_kernel(param))
+    {
+        ERAISE(-EFAULT);
+    }
```
paulcallen, line 27: if it exists and is valid memory please please memset here to zero

tests/ltp/ramfs_tests_other_errors.txt
```diff
@@ -558,7 +558,6 @@
 /ltp/testcases/kernel/syscalls/sched_get_priority_min/sched_get_priority_min02
 /ltp/testcases/kernel/syscalls/sched_getaffinity/sched_getaffinity01
 /ltp/testcases/kernel/syscalls/sched_getattr/sched_getattr02
-/ltp/testcases/kernel/syscalls/sched_getparam/sched_getparam01
 /ltp/testcases/kernel/syscalls/sched_getparam/sched_getparam03
```
paulcallen, line 561: what is to stop test 03 from being enabled too?

tools/myst/myst.edl
```diff
@@ -123,7 +135,13 @@ enclave
             uint64_t self_event,
             [in] const struct myst_timespec* timeout);
 
-        long myst_sched_yield_ocall();
+        long myst_sched_yield_ocall()
+            transition_using_threads;
+
+        long myst_sched_getparam_ocall(
+            pid_t pid,
+            [out] struct myst_sched_param* param)
```
paulcallen, line 143: why can this not be just `struct sched_param`? You can include the correct header in the top of the file I believe?

paulcallen (commented):
looking good. A few comments/questions though

tests/ltp/ramfs_tests_other_errors.txt
RRathna: test /ltp/testcases/kernel/syscalls/sched_getparam/sched_getparam03 passes with the file /proc/sys/kernel/pid_max (can verify by applying changes in PR - #1013). But this PR still needs to address tid cycling as noted by comment in threads.c. Once we figure that out, we can enable it too.

tools/myst/myst.edl
RRathna: I tried adding #include "sched.h", "myst/sched.h" and "sys/sched.h" and none of them worked as expected. Do you know the right header?

kernel/sched.c
RRathna: I memset param to 0 here - https://github.com/RRathna/mystikos/blob/41d8e8da122231c9c50f3a4e1044de3e9f3e6ea9/crt/sched.c#L10-L11, should I move it here?

include/myst/thread.h
RRathna: On the host /proc/sys/kernel/pid_max is set to 32678 (0x8000) and I thought keeping the lowest possible would be a good place to start. Currently we don't have a solution for wraping around for tid/pids once the pid_max is reached (reason why #1013 is a draft).

From man proc - 
/proc/sys/kernel/pid_max (since Linux 2.5.34)
              This file specifies the value at which PIDs wrap around
              (i.e., the value in this file is one greater than the
              maximum PID).  PIDs greater than this value are not
              allocated; thus, the value in this file also acts as a
              system-wide limit on the total number of processes and
              threads.  The default value for this file, 32768, results
              in the same range of PIDs as on earlier kernels.  On
              32-bit platforms, 32768 is the maximum value for pid_max.
              On 64-bit systems, pid_max can be set to any value up to
              2^22 (PID_MAX_LIMIT, approximately 4 million).

kernel/sched.c
paulcallen: I know! But the kernel trusts no one so needs to be done here.

tests/ltp/ramfs_tests_other_errors.txt
paulcallen: thanks.

tools/myst/myst.edl
paulcallen: man page suggests just `#include <sched.h>`

include/myst/thread.h
paulcallen: ok. If this is going to be handled properly in the other PR then I will accept that. Can you put a comment in here that this is the typical default used in the host operating system.

tools/myst/myst.edl
RRathna: Tried, didn't work either..

paulcallen (approved):

Issue #880, opened 2021-10-07
exit() called from child pthread causes different exit status in Mystikos vs Linux


vtikoo:
If a child thread calls `exit(num)`, Linux shuts down the process with exit status `num`. On Mystikos, the parent receives a SIGKILL(signum 9) from the exit of the child pthread, and the process returns an exit status of 137(128 + SIGKILL).

Sample program reproducing this issue -

```C
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>

void* thread_func(void* args)
{
    exit(0);
}

int main()
{
    pthread_t child;
    pthread_create(&child, NULL, thread_func, NULL);
    pthread_join(child, NULL);
    printf("Parent exiting.\n");
    return 0;
}
```

PR #1016, opened 2021-11-13
Setting process status not atomic and set once

paulcallen:
it is set atomically between SYS_exit, SYS_exit_group and a terminating
signal.
On top of this, SYS_exit_group now terminates the current thread in the
same way that SYS_exit does.

### DIFF
--- a/kernel/signal.c
+++ b/kernel/signal.c
@@ -278,6 +278,19 @@ static long _default_signal_handler(unsigned signum)
     myst_process_t* process = myst_process_self();
     myst_thread_t* process_thread = process->main_process_thread;
 
+    bool process_status_set = false;
+    if (__atomic_compare_exchange_n(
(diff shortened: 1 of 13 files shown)


paulcallen:
fixes #880

kernel/signal.c
```diff
@@ -278,6 +278,19 @@ static long _default_signal_handler(unsigned signum)
...
+    bool process_status_set = false;
+    if (__atomic_compare_exchange_n(
+            &process->exit_status_signum_set,
+            &process_status_set,
+            true,
+            false,
+            __ATOMIC_RELEASE,
+            __ATOMIC_ACQUIRE))
+    {
+        process->exit_status = 128 + signum;
```
mikbras, line 290: Too bad there isn't a macro for this 128 value.

mikbras (approved):

PR #1018, opened 2021-11-15
[sched_getparam] bug fix  - look for pid in caller's threadpool too

RRathna:
sched_getparam should look in all processes and all threads in caller's threadpool for matching pid, change struct myst_sched_param to match glibc definition, and add memset protection for sched_param in kernel/sched.c


vtikoo (approved):

vtikoo:
@RRathna pls wait for the pipelines to finish before merging. Thanks.

PR #1025, opened 2021-11-17
itimer multi-process, update wait and getitimer

paulcallen:
add multi-process support
getitimer calculates based off of wait start time, more accurate
wait does wait for full amount, not pulsing every millisecond
wait timer does not overrun if got woken with interrupt

### DIFF
--- a/crt/enter.c
+++ b/crt/enter.c
@@ -55,11 +55,19 @@ int myst_pre_launch_hook()
 
 long myst_syscall(long n, long params[6])
 {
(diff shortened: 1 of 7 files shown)


mikbras (approved):
Look good. I was wondering whether the itimer thread shuts down before freeing the process itimer object.

paulcallen:
main thread does a couple of blocks before. Child process does it as almost the last thing that it does so all threads and processes have gone.

Issue #556, opened 2021-07-08
Environment variables from host should be disabled by default


jxyang:
In the [document](https://github.com/deislabs/mystikos/blob/main/doc/sign-package.md) it says:

>  HostEnvironmentVariables | A list of environment variables that can be imported from the insecure host

Without an explicit entry in config.json that's measured, we shouldn't rely on any environment variable from the untrusted host, even though that's convenient. 

The proper way here is to create a config.json and put these two variables in there and run myst with `--app-config-path`

bodzhang:
I assume that in the release mode (packaged app), if the config.json file does not contain a "HostEnvironmentVariables" entry, any environment variable provided through the app execution command line will be ignored. @paulcallen, can you confirm?

jxyang:
Because the application could use host environment variables in unpredicted ways, allowing the enclave to take dependency on a host environment variable is against the fundamental idea that the enclave should be an isolated execution environment. Much like containers allow application to access only environment variables specified with `ENV`.

bodzhang:
@jxyang , doo you mean with packaged mode, the host environment variables are still exposed to the Mystikos environment in the Enclave?

paulcallen:
by default with target=sgx, no environment variables are inherited from the host.
When we are not signing and packaging we may not have a config.json with an allow list and so we have no config when we get into the enclave and with that all environments variables are used.  
We should have a consistent story without the config.json file, which also is required for linux target

PR #1027, opened 2021-11-17
dont export host env vars unless there is a config

salsal97:


salsal97:
This PR addresses https://github.com/deislabs/mystikos/issues/556

PR #1040, opened 2021-11-22
Fix a bug in O_NOFOLLOW check in fs_open, for ext2fs and ramfs.

bodzhang:

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -466,16 +466,6 @@ long myst_syscall_open(const char* pathname, int flags, mode_t mode)
     if (!(locals = malloc(sizeof(struct locals))))
         ERAISE(-ENOMEM);
 
-    if (flags & O_NOFOLLOW)
-    {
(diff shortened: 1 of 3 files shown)


mingweishih (approved):

kernel/syscall.c
```diff
@@ -466,16 +466,6 @@ long myst_syscall_open(const char* pathname, int flags, mode_t mode)
     if (!(locals = malloc(sizeof(struct locals))))
         ERAISE(-ENOMEM);
 
-    if (flags & O_NOFOLLOW)
```
vtikoo, line 469: Is there any reason to move this check into the filesystems themselves?

bodzhang: myst_syscall_open() can be used a create a file. lstat() on a file not created yet false-trigger the ELOOP error related to the O_NOFOLLOW/O_PATH check.

mikbras (approved):
Looks right to me.

PR #1044, opened 2021-11-23
Lower kstack size, change locking in kstack

paulcallen:


mikbras (approved):
The disabling of switchless mprotect is temporary. Should create issue.

PR #1045, opened 2021-11-23
Fix __fwritex() console write crash

mikbras:
The ``tests/pthread`` test crashes in OE's **musl libc** ``__fwritex()`` function. This happens when **strace** is enabled. This fix bypasses ``__fwritex()`` by introducing``myst_write_console_ocall()`` as defined below.

```C++
        long myst_write_console_ocall(
            int fd, /* STDOUT_FILENO or STDERR_FILENO */
            [in, size=count] const void* buf,
            size_t count);
```

Since the crash only occurs when ``__fwritex()`` is bombarded by hundreds of threads, we suspect a file-stream locking bug in **OE's musl libc**.

This fix is a better approach for two reasons.
- Mystikos in general rarely relies on OE musl libc.
- The new OCALL consumes less of the kernel stack than **musl libc fwrite()** (which is unpredictable), and since we are trying to optimize the kernel stack, this approach is at least predictable and probably minimal.

### DIFF
--- a/target/sgx/enclave/tcall.c
+++ b/target/sgx/enclave/tcall.c
@@ -305,21 +305,8 @@ long myst_tcall(long n, long params[6])
             int fd = (int)x1;
             const void* buf = (const void*)x2;
             size_t count = (size_t)x3;
-            FILE* stream = NULL;
 
-            if (fd == STDOUT_FILENO)
-                stream = stdout;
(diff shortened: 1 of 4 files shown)


mikbras:
The bug in OE may have been fixed by the following PR. https://github.com/openenclave/openenclave/pull/4286.

paulcallen (approved):

PR #1048, opened 2021-11-23
Invoke OE syscalls on bigger stack

mikbras:
This PR introduces `myst_invoke()` so that OE syscalls can be invoked on a dynamically allocated stack that is larger than the default kernel-stack. This allows the kernel stack size to be reduced.

This will help optimize `tests/pthread`.


mikbras:
Closing this as infeasible.

PR #1049, opened 2021-11-24
strace log crash with bad ptr

paulcallen:

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -4178,7 +4178,10 @@ static long _syscall(void* args_)
         {
             const char* path = (const char*)x1;
 
-            _strace(n, "path=\"%s\"", path);
+            if (path && !myst_is_bad_addr_read(path, 1))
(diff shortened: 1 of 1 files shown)


mingweishih (approved):

kernel/syscall.c
```diff
@@ -4178,7 +4178,10 @@ static long _syscall(void* args_)
         {
             const char* path = (const char*)x1;
 
-            _strace(n, "path=\"%s\"", path);
+            if (path && !myst_is_bad_addr_read(path, 1))
```
vtikoo, line 4181: Should we error out if `path` is bad?

paulcallen: we do, but all proper handling is done in the syscall handler function. We want to do a log before moving on though so this is making sure we can at least log first. Just returning an error here can cause other tests to fail because there is often an order to parameter validation and returning this error can break those rules. 
This is why I have chosen this way of handling the problem for logging in the past and here.

PR #1068, opened 2021-11-30
Install az cli

CyanDevs:
Azure SDK tests require Azure CLI to be installed which was erroneously removed in #1062. 
This PR will install `az` as part of the CI setup for the Azure SDK pipeline only.

### DIFF
--- a/.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
+++ b/.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
@@ -79,6 +79,19 @@ pipeline {
                    """
             }
         }
+        stage('Setup AZ CLI') {
+            steps {
+                withCredentials([string(credentialsId: 'Jenkins-ServicePrincipal-ID', variable: 'SERVICE_PRINCIPAL_ID'),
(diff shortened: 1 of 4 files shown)


salsal97 (approved):

.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
```diff
@@ -79,6 +79,13 @@ pipeline {
                    """
             }
         }
+        stage('Setup AZ CLI') {
+            steps {
+                sh """
+                    ${JENKINS_SCRIPTS}/azure-sdk/install-azure-cli.sh
```
rs--, line 85: I assume that because we're using the Jenkins plugins to authenticate/interact with Azure, Azure CLI is only needed for the SDK tests?

rs-- (approved):

.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
CyanDevs: It looks like the VM will not be authenticated as part of the Azure SDK tests. I've added that in as part of setup now. Thanks!

PR #1069, opened 2021-11-30
Enable full 32-bit uid/gid support in ext2fs implementation

bodzhang:

This PR addresses three issues:

1. Existing ext2fs implementation does not support full 32-bit uid/gid in ext2fs inode. The issue was exposed by cpython test suite's test_posix test set. In ext2fs inode definition, an OS specific data structure for Linux holds the high 16 bits of uid/gid. The PR follows the Linux definition to support full 32b-bit uid/gid
2. Existing sched_getaffinity and sched_setaffinity implementation's error code implementation does not match Linux syscall manual specification. The issue was also exposed by cpython test suite's test_posiz test set. The PR changed the error code implementation to match the Linux syscall manual. The PR also add address validity check.
3. Add missing FIOCLEX/FIONCLEX ioctl support for pipe fd.

The cpython test_posix test set is not enabled in this PR, pending resolution to other issues affecting that test set.

### DIFF
--- a/kernel/pipedev.c
+++ b/kernel/pipedev.c
@@ -737,25 +737,39 @@ static int _pd_ioctl(
         arg,
         myst_getppid()));
 
-    if (request == TIOCGWINSZ)
+    switch (request)
     {
-        ERAISE(-EINVAL);
-    }
-    else if (request == FIONBIO)
-    {
-        int* val = (int*)arg;
-
-        if (!val)
+        case TIOCGWINSZ:
+        {
             ERAISE(-EINVAL);
+            break;
+        }
+        case FIONBIO:
+        {
+            int* val = (int*)arg;
 
-        if (*val)
-            pipe->fl_flags |= O_NONBLOCK;
-        else
-            pipe->fl_flags &= ~O_NONBLOCK;
-    }
-    else
-    {
-        ERAISE(-ENOTSUP);
+            if (!val)
+                ERAISE(-EINVAL);
+
+            if (*val)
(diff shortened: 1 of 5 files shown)


kernel/pipedev.c
```diff
@@ -737,25 +737,39 @@ static int _pd_ioctl(
...
-            pipe->fl_flags &= ~O_NONBLOCK;
-    }
-    else
-    {
-        ERAISE(-ENOTSUP);
+            if (!val)
+                ERAISE(-EINVAL);
+
+            if (*val)
+                pipe->fl_flags |= O_NONBLOCK;
```
Sahakait, line 755: `pipedev.c` has `#define FL_FLAGS (O_APPEND | O_ASYNC | O_DIRECT | O_NOATIME | O_NONBLOCK)`, which is used for `fnctl`. Is `O_NONBLOCK` the only applicable flag here.

mikbras (approved):
Looks good, but is there any test for ids bigger than 16 bits?

kernel/pipedev.c
bodzhang: Currently, Mystikos only partially supports ioctl (for all supported dev types). Further enhancement will be evaluated and prioritized based on targeted programing language runtime needs.

bodzhang:
> Looks good, but is there any test for ids bigger than 16 bits?

cpython test_posix test set covers full 32-bit uid/gid. It will be enabled once the failures on other tests inside the test set are addressed.

PR #1071, opened 2021-11-30
Option to run tests using simplified runner.

anakrish:
make one-runner TEST=... runs a single test using the runner.
make all-runner TESTS_FILE=... runs tests specified in given file using the runner.

Both TEST and TESTS_FILE can be omitted to use defaults.


tests/aspnetcore/runner/Program.cs
```diff
@@ -0,0 +1,155 @@
...
+	    {
+		Console.WriteLine("===== {0}! {1}", failed ? "Failed" : "Passed", assemblyName);
+	    }
+        }
+
+        static void Main(string[] args)
+        {
+            if (args.Length != 2 && args.Length != 1)
+            {
+                Console.WriteLine("Usage: runner <file-with-list-of-tests> <path-prefix>");
```
vtikoo, line 127: Whats the path-prefix option for?

vtikoo:
Should we switch to use the runner version for running the aspnetcore tests? It would be good validate how this works in the pipeline.

tests/aspnetcore/runner/Program.cs
anakrish: Xuejun added it. It shortens the test paths by eliminating the prefix. We can retain this for now.

radhikaj (approved):

PR #1072, opened 2021-11-30
dont export host env vars unless there is a config

salsal97:
provide explicit config to tests needing host envv


tests/myst/exec-not-signed/Makefile
```diff
@@ -33,7 +33,6 @@ test-a:
 	grep -E "argv\[2]=green" result
 	grep -E "argv\[3]=blue" result
 	grep -E "argv\[4]=yellow" result
-	grep -E "TESTNAME=tests/myst/exec-not-signed" result
```
paulcallen, line 36: We are expecting no environment variables, so create a result file for this test and then do a compare of the results to the expected output... which is probably just the args at this point

paulcallen (commented):
Minor change in the test so we validate the non-existence of environment variables.
Other than that it looks good.

tests/myst/exec-not-signed/Makefile
salsal97: Added negative testing for this

mikbras (approved):

mikbras:
Looks good. Thank you for doing this Saloni.

PR #1078, opened 2021-12-02
Update release pipeline

rs--:
- parallelize tests
- test Debian and tarball packages on 18.04 and 20.04


.jenkins/pipelines/release.Jenkinsfile
```diff
@@ -0,0 +1,60 @@
...
+        label 'Jenkins-Shared-DC2'
+    }
+    options {
+        timeout(time: 300, unit: 'MINUTES')
+    }
+    parameters {
+        string(name: "REPOSITORY", defaultValue: "deislabs")
+        string(name: "BRANCH", defaultValue: "main", description: "Branch to build")
+        choice(name: "REGION", choices:['useast', 'canadacentral'], description: "Azure region for the SQL solutions test")
+        string(name: "PACKAGE_NAME", description: "optional - release package to install (do not include extension)")
```
CyanDevs, line 12: Is this really optional? If it is you need to put a defaultValue.
Also, what tests are ran when PACKAGE_NAME is not defined, in the context of the matrix and the TEST_PACKAGE axis?

.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
```diff
@@ -99,8 +102,49 @@ pipeline {
...
+                           rm -rf mystikos
+                           tar xzf ${PACKAGE_INSTALL}
+                           """
+                    }
+                }
+            }
+        }
         stage('Run Azure SDK tests') {
+            environment {
+                MYST_INSTALLED_BIN = "${ params.PACKAGE_EXTENSION == 'deb' ? '/opt/mystikos/bin' : '\${WORKSPACE}/mystikos/bin' }"
```
CyanDevs, line 132: May be cleaner to instead use PATH when installing from deb / tar so you don't need to do this linking? Unless there are issues other than calling the binaries

Example `PATH = "/opt/mystikos/bin:\$PATH"`

CyanDevs: Also, it would be better to have put this in it's own stage so you don't need to echo titles out and be slightly easier to navigate in Jenkins Blue Ocean

.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
```diff
@@ -99,8 +102,49 @@ pipeline {
                    """
             }
         }
+        stage("Install package") {
+            when {
+                expression { params.PACKAGE_NAME != "" }
+            }
+            steps {
+
```
CyanDevs, line 110: Nit: remove whitespace before and after azureDownload

.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
```diff
@@ -99,8 +102,49 @@ pipeline {
...
+                azureDownload(
+                    downloadType: 'container',
+                    containerName: 'mystikosreleases',
+                    includeFilesPattern: "${PACKAGE_INSTALL}",
```
CyanDevs, line 114: If `PACKAGE_NAME` isn't defined (which is a condition of running this stage) then `PACKAGE_INSTALL` would produce an invalid name? E.g. `Ubuntu-2004_.deb`

.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
rs--: The debian/tar packages are built with three binaries: `myst`, `myst-gdb`, and `myst-appbuilder`. But tests actually use four binaries, the three mentioned above and `myst-lldb`, which isn't included in the released packages.

Tests allow paths to the `myst`, `myst-gdb`, and `myst-lldb` binaries, but not to `myst-appbuilder`.

So because `myst-lldb` is being referenced in the `build/bin` folder, and `myst-appbuilder` is being referenced in the `scripts` folder, I figured the most consistent way to reference different binaries is to let the tests reference the original source locations, but just have symbolic links to the actual binaries that are to be used.

.jenkins/pipelines/release.Jenkinsfile
rs--: Good catch, `PACKAGE_NAME` isn't optional in this pipeline

.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
rs--: Also, yes, moved to the stage above.

.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
rs--: I moved the package install step into a stage requiring `PACKAGE_NAME` to be a non-empty string

.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
```diff
@@ -12,6 +12,8 @@ pipeline {
         string(name: "PULL_REQUEST_ID", defaultValue: "", description: "If you are building a pull request, enter the pull request ID number here. (ex. 789)")
         choice(name: "TEST_CONFIG", choices: ['None', 'Nightly', 'Code Coverage'], description: "Test configuration to execute")
         string(name: "COMMIT_SYNC", defaultValue: "", description: "optional - used to sync outputs of parallel jobs")
+        string(name: "PACKAGE_NAME", description: "optional - release package to install (do not include extension)")
```
CyanDevs, line 15: Remove "optional" from param description

CyanDevs (approved):

.jenkins/pipelines/standalone/azure-sdk-tests.Jenkinsfile
rs--: `PACKAGE_NAME` is optional for standalone pipelines

CyanDevs: Ok, could you make sure all optional parameters have a `defaultValue` please? Either set to blank or null. Thanks!

PR #1088, opened 2021-12-07
Fix realpath given empty path

asvrada:
Fix an issue that `myst_realpath("")` would return `/`

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -1223,8 +1223,12 @@ static const char* _trim_trailing_slashes(
     if (len >= size)
         return NULL;
 
+    /* if empty pathname or equal to "/" */
+    if (len == 0 || (pathname[0] == '/' && pathname[1] == '\0'))
+        return pathname;
+
     /* remove trailing slashes from the pathname if any */
-    if ((len = strlen(pathname)) && pathname[len - 1] == '/')
+    if (pathname[len - 1] == '/')
     {
         memcpy(buf, pathname, len + 1);
 
--- a/kernel/realpath.c
+++ b/kernel/realpath.c
@@ -30,6 +30,9 @@ int myst_realpath(const char* path, myst_path_t* resolved_path)
     if (!path || !resolved_path)
         ERAISE(-EINVAL);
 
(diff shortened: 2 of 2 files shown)


kernel/syscall.c
```diff
@@ -1228,7 +1228,7 @@ static const char* _trim_trailing_slashes(
     {
         memcpy(buf, pathname, len + 1);
 
-        for (char* p = buf + len; p != buf && p[-1] == '/'; *--p = '\0')
```
mikbras, line 1231: This changes the meaning of the function so that it no longer trims training slashes if there is only one.

So _trim_trailing_slashes("/") => "/" and the trailing slash is not trimmed.

kernel/syscall.c
```diff
@@ -1228,7 +1228,7 @@ static const char* _trim_trailing_slashes(
...
+        for (char* p = buf + len; (p - buf) > 1 && p[-1] == '/'; *--p = '\0')
```
mikbras, line 1231: Maybe it would be clearer to precheck at the start of the function whether the path is "/", like this:

```
    if (pathname[0] == '/' && pathname[1])
        return pathname;
```

mikbras: Also, this function already obtained the length of the pathname twice.

mikbras: Here is an alternative implementation.

```
static const char* _trim_trailing_slashes(
    const char* pathname,
    char* buf,
    size_t size)
{
    size_t len = strlen(pathname);

    if (len >= size)
        return NULL;

    /* if empty pathname or equal to "/" */
    if (len == 0 || (pathname[0] == '/' && pathname[1] == '\0'))
        return pathname;

    /* remove trailing slashes from the pathname if any */
    if (pathname[len - 1] == '/')
    {
        memcpy(buf, pathname, len + 1);

        for (char* p = buf + len; p != buf && p[-1] == '/'; *--p = '\0')
            ;

        pathname = buf;
    }

    return pathname;
}
```

mikbras (approved):
Thank you so much for doing this.

PR #1090, opened 2021-12-07
Performance optimizations for Redis

mikbras:
This PR fixes the performance of the Redis benchmarks. SGX is now on par with native Linux. We made three changes:
- Made ``myst_epoll_wait_ocall()`` switchless
- Removed the ``in`` annotation of the  ``events`` ocall parameter (``events`` is strictly an output parameter)
- Reduced ``maxevents`` to **64** before calling ``myst_epoll_wait_ocall()`` (to avoid copying back large output buffers)

The final change was required because Redis passes **10,128** as the ``maxevents`` parameter (the number of elements in the ``events`` output buffer), which incurs a large copy (**121,536** bytes) on every call to ``myst_epoll_wait_ocall()``. We shorten ``maxevents`` to **64** which limits the number of events that can be returned in a single call (even a handful of simultaneous events is unlikely, let alone **64**). If there are more than **64** events, they will be reported across more than one call (again, very unlikely).


bodzhang:
Did the latest push mix up the PR with other changes?

mikbras:
> Did the latest push mix up the PR with other changes?

Oh my! Probably someone tried to merge this branch and ended up merging into this branch.

PR #1091, opened 2021-12-08
.Net support and limitations documentation

vtikoo:

### DIFF
--- a/doc/dotnet-support.md
+++ b/doc/dotnet-support.md
@@ -1,7 +1,7 @@
 # Mystikos Support for .Net
 
 ### Supported .Net and ASP.Net version
-We recommend using .net 3.1 and .net 5.
+We recommend using .net/ASP.Net 3.1 and .net/ASP.Net 5.0.
 
 ### Supported Distros:
 - Alpine
@@ -14,6 +14,7 @@ We recommend using official images from [mcr.microsoft.com/dotnet](https://hub.d
 #### dotnet configurations
 - Run with dotnet diagnostics disabled. This can be controlled via the `COMPlus_EnableDiagnostics` environment variable set to 0.
 - If running with low memory settings, you can control dotnet's managed heap size by [`COMPlus_GCHeapHardLimit` environment variable](https://docs.microsoft.com/en-us/dotnet/core/run-time-config/garbage-collector#heap-limit).
+- `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` - Some applications/tests using Unicode might have to set this to 0. In many cases it could be 1. If set to 0, the docker file has to ensure lib icu is included in the image.
 
 ##### Mystikos specific configurations
 - `ThreadStackSize` of 256k. The default pthread stack size in Mystikos is 128k. We found System.Linq.Expressions can cause application to hang with the default 128k thread stack. More details [here](https://github.com/dotnet/runtime/issues/61757).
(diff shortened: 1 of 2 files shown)


vtikoo:
@jxyang do we have a recommendation for `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT`?

doc/dotnet-support.md
```diff
@@ -0,0 +1,36 @@
+# Mystikos Support for .Net
+
+### Supported .Net and ASP.Net version
+We recommend using .net 3.1 and .net 5.
```
jxyang, line 4: What about the asp.net version support?

doc/dotnet-support.md
```diff
@@ -0,0 +1,36 @@
...
+myst-lldb, a lldb extension is distributed with Mystikos. This can be used to debug high level C# code.
+More details [here](
+https://github.com/deislabs/mystikos/tree/main/tests/dotnet-sos#readme
+).
+
+### Known limitations
+- GC suspension signals are not delivered in a real-time fashion. Actively being addressed.
+
+### Library Support
+TODO
```
jxyang, line 36: I think we should provide a sample docker file and a sample config.json to put the recommendations together.

vtikoo: We have a sample docker file in the [getting started dotnet document](https://github.com/deislabs/mystikos/blob/main/doc/user-getting-started-docker-dotnet.md#containerize-the-program). Its however only for 3.1.
I can add the rest there and link it up here.

doc/dotnet-support.md
```diff
@@ -0,0 +1,39 @@
...
+- Alpine
+- Ubuntu
+
+We recommend using official images from [mcr.microsoft.com/dotnet](https://hub.docker.com/_/microsoft-dotnet/).
+
+### Recommended configurations
+
+#### dotnet configurations
+- Run with dotnet diagnostics disabled. This can be controlled via the `COMPlus_EnableDiagnostics` environment variable set to 0.
+- If running with low memory settings, you can control dotnet's managed heap size by [`COMPlus_GCHeapHardLimit` environment variable](https://docs.microsoft.com/en-us/dotnet/core/run-time-config/garbage-collector#heap-limit).
```
jxyang, line 16: As discussed in the meeting, we should also experiment with  building the dotnet app with or without ``self-contained` flag and see the impacts on the memory requirement.

anakrish (commented):
It is important to mention that `dotnet test` does not work well due to the use of two levels for vforking.

doc/dotnet-support.md
jxyang: Ignore the previous comment. My experiment shows `self-contained` is not relevant to memory requirement.

vtikoo: Resolving based on your Teams message that `self-contained` doesn't impact memory requirement.

vtikoo:
> It is important to mention that `dotnet test` does not work well due to the use of two levels for vforking.

Do you have any specific issues in mind? Given that your CRT fix should hopefully address the memory corruption failures.
I can think of only larger memory requirement(~6G) in a two-level fork.

doc/dotnet-support.md
vtikoo: Added

anakrish:
There are few issues I can think of with `dotnet test`
- dotnet-sos doesn't work 
- Currently `dotnet test` is broken since dup is not supported for AF_LOCAL sockets
- We don't lock `dotnet test` anymore in CI. Therefore it is not tested.

jxyang (approved):

vtikoo:
> * dotnet-sos doesn't work

Added a line about this.

> * Currently `dotnet test` is broken since dup is not supported for AF_LOCAL sockets

Is this going to be addressed soon? If its a fundamental limitation going forward, we should definitely document it in more detail.

PR #1093, opened 2021-12-10
epoll maxevents optimization

mikbras:
This PR makes the following optimizations to ``epoll_wait()``.

- Makes``myst_epoll_wait_ocall()`` switchless
- Removes the ``in`` annotation of the  ``events`` ocall parameter (``events`` is strictly an output parameter)
- Reduces ``maxevents`` to the number of file descriptors registered with ``epoll_ctl(EPOLL_CTL_ADD)`` (since there can never be more events than the number of file descriptors).

The final change was required because Redis passes **10,128** as the ``maxevents`` parameter (the number of elements in the ``events`` output buffer), which incurs a large copy (**121,536** bytes) on every call to ``myst_epoll_wait_ocall()``.

### DIFF
--- a/kernel/epolldev.c
+++ b/kernel/epolldev.c
@@ -25,10 +25,16 @@
 
 typedef struct epoll_entry epoll_entry_t;
 
+/* comment this out to disable the "maxevents optimization" */
+#define ENABLE_MAXEVENTS_OPTIMIZATION
+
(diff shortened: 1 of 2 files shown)


bodzhang (approved):

mikbras:
Thank you Bo!

PR #1095, opened 2021-12-13
Fix CRT corruption caused by (v)fork

anakrish:
In MUSL's vfork implementation, the child process clears libc variables.
This is OK since after a (v)fork, the child gets its own memory space and any changes
done by the child does not affect the parent.

In Mystikos, after a (v)fork, the child still uses the parent's libc/CRT. Therefore, any changes done
by the child to CRT is also visible to the parent. A new CRT is created for the child only after
execve call.

The child, immediately after a vfork, goes on to set libc.threds_minus_1 to zero. This causes MUSL
to think that there is only one thread in the program, and subsequently it skips locking.
In Mystikos, this has the effect that the CRT becomes thread-unsafe after a fork. This leads to
various issues. The most common manifestation is strange crashes in malloc's bin/unbin functions.

### DIFF
--- a/third_party/musl/crt/patch.diff
+++ b/third_party/musl/crt/patch.diff
@@ -933,6 +933,41 @@ index eeae1d0a..7e13faa8 100644
  #endif
  }
 +weak_alias (sendmmsg, __sendmmsg);
+diff --git a/src/process/fork.c b/src/process/fork.c
+index fb42478a..2d0921f8 100644
+--- a/src/process/fork.c
++++ b/src/process/fork.c
+@@ -22,15 +22,28 @@ pid_t fork(void)
+ #else
+ 	ret = __syscall(SYS_clone, SIGCHLD, 0);
+ #endif
+-	if (!ret) {
++
++        if (!ret) {
+ 		pthread_t self = __pthread_self();
+ 		self->tid = __syscall(SYS_gettid);
+ 		self->robust_list.off = 0;
+ 		self->robust_list.pending = 0;
+ 		self->next = self->prev = self;
++
++		// In Linux, the child process can safely modify the libc internal variables as below
++		// since (v)fork creates a new memory space for the child; any changes done by the child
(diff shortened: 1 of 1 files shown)


Sahakait:
Change looks good. Is there a sample that was previously failing/unstable that is now fixed by this? Out of curiosity, I tried the change with my [vfork hang repro](https://github.com/Sahakait/mystikos/tree/vfork_hang_repro) but it looks like the issue is unrelated (it still hangs).

third_party/musl/crt/patch.diff
```diff
@@ -933,6 +933,35 @@ index eeae1d0a..7e13faa8 100644
...
++
++#if 0
++       // In Linux, the child process can safely modify the libc internal variables as below
++       // since (v)fork creates a new memory space for the child; any changes done by the child
++       // does not affect the parent.
++       // However, in mystikos, after (v)fork, the child still shares the same libc/crt as the parent.
++       // Therefore, it is not safe to modify the libc state. It is only after execve that the child
++       // gets its own crt/libc.
++       if (!ret) {
+ 		pthread_t self = __pthread_self();
```
vtikoo, line 954: Setting pthread state in the child should work if __pthread_self() fetches the right thread descriptor.

vtikoo:
This is a very exciting find!
We can probably still keep the setting of thread specific state. Otherwise looks good.

third_party/musl/crt/patch.diff
anakrish: Turns out this line is needed. LTP tests use (sometimes) use fork. Retaining the logic to set the pthread_t members allow these tests to continue to pass. 

Although I'm not sure how robust the fork implementation is in the first place. Earlier, after the fork, the system was incorrectly put in a "single-threaded" mode and for some tests, a different code path was taken.

vtikoo (approved):

PR #1100, opened 2021-12-16
Add cpython v3.10 tests

asvrada:


tests/cpython-tests/Makefile
```diff
@@ -26,7 +26,7 @@ FS=ext2fs$(VER)
 # Set timeout to 50 mins (to run both the python3.8 test suite and the python3.9 test suite)
 export TIMEOUT=3000
```
jxyang, line 27: Please update the comment to include 3.10.

Do we need to increase the time out?

asvrada: The PR pipeline finished successfully for this PR, so I think that might not be necessary? But will add a few minutes just in case

jxyang (approved):

PR #1112, opened 2022-01-05
Do not return to OE after signal was delivered to kernel

mingweishih:

### DIFF
--- a/kernel/signal.c
+++ b/kernel/signal.c
@@ -756,9 +756,17 @@ done:
     return ret;
 }
 
-long myst_handle_host_signal(siginfo_t* siginfo, mcontext_t* mcontext)
+void myst_handle_host_signal(siginfo_t* siginfo, mcontext_t* mcontext)
 {
-    return _handle_one_signal(siginfo->si_signo, siginfo, mcontext);
+    _handle_one_signal(siginfo->si_signo, siginfo, mcontext);
+
+    /* The signal handler in the programming language runtime or the application
+     * might not return. If execution returns here, set the CPU context as the
+     * returned mcontext to continue execution. */
+    myst_sigreturn(mcontext);
+
+    // Unreachable
+    assert(0);
 }
 
 int myst_signal_altstack(const stack_t* ss, stack_t* old_ss)
--- a/tools/myst/enc/enc.c
+++ b/tools/myst/enc/enc.c
@@ -43,7 +43,6 @@
 #define IRETFRAME_Rsp IRETFRAME_EFlags + 8
 
 static myst_kernel_args_t _kargs;
-static mcontext_t _mcontext;
 
 struct myst_final_options final_options = {0};
 
@@ -58,7 +57,6 @@ long _exception_handler_syscall(long n, long params[6])
 
 static void _oe_context_to_mcontext(oe_context_t* oe_context, mcontext_t* mc)
 {
-    memset(mc, 0, sizeof(mcontext_t));
     mc->gregs[REG_R8] = (int64_t)(oe_context->r8);
     mc->gregs[REG_R9] = (int64_t)(oe_context->r9);
     mc->gregs[REG_R10] = (int64_t)(oe_context->r10);
@@ -80,31 +78,11 @@ static void _oe_context_to_mcontext(oe_context_t* oe_context, mcontext_t* mc)
     mc->gregs[REG_RDX] = (int64_t)(oe_context->rdx);
 
     mc->gregs[REG_EFL] = (int64_t)(oe_context->flags);
-}
 
-static void _mcontext_to_oe_context(mcontext_t* mc, oe_context_t* oe_context)
-{
-    oe_context->r8 = (uint64_t)(mc->gregs[REG_R8]);
-    oe_context->r9 = (uint64_t)(mc->gregs[REG_R9]);
-    oe_context->r10 = (uint64_t)(mc->gregs[REG_R10]);
-    oe_context->r11 = (uint64_t)(mc->gregs[REG_R11]);
-    oe_context->r12 = (uint64_t)(mc->gregs[REG_R12]);
-    oe_context->r13 = (uint64_t)(mc->gregs[REG_R13]);
-    oe_context->r14 = (uint64_t)(mc->gregs[REG_R14]);
-    oe_context->r15 = (uint64_t)(mc->gregs[REG_R15]);
+    // Ensure the definitions between musl and OE are aligned
+    assert(sizeof(struct _fpstate) == sizeof(oe_basic_xstate_t));
 
-    oe_context->rsi = (uint64_t)(mc->gregs[REG_RSI]);
-    oe_context->rdi = (uint64_t)(mc->gregs[REG_RDI]);
-    oe_context->rbp = (uint64_t)(mc->gregs[REG_RBP]);
-    oe_context->rsp = (uint64_t)(mc->gregs[REG_RSP]);
-    oe_context->rip = (uint64_t)(mc->gregs[REG_RIP]);
-
-    oe_context->rax = (uint64_t)(mc->gregs[REG_RAX]);
-    oe_context->rbx = (uint64_t)(mc->gregs[REG_RBX]);
-    oe_context->rcx = (uint64_t)(mc->gregs[REG_RCX]);
-    oe_context->rdx = (uint64_t)(mc->gregs[REG_RDX]);
-
-    oe_context->flags = (uint64_t)(mc->gregs[REG_EFL]);
+    memcpy(mc->fpregs, &(oe_context->basic_xstate), sizeof(struct _fpstate));
 }
 
 static uint64_t _forward_exception_as_signal_to_kernel(
@@ -112,8 +90,14 @@ static uint64_t _forward_exception_as_signal_to_kernel(
 {
     uint32_t oe_exception_code = oe_exception_record->code;
     oe_context_t* oe_context = oe_exception_record->context;
+    mcontext_t mcontext = {0};
+    struct _fpstate fpregs __attribute__((aligned(16))) = {0};
     siginfo_t siginfo = {0};
-    _oe_context_to_mcontext(oe_context, &_mcontext);
+    bool supported_code = false;
+
+    mcontext.fpregs = &fpregs;
+
+    _oe_context_to_mcontext(oe_context, &mcontext);
 
     // Kernel should be the ultimate handler of #PF, #GP, #MF, and #UD.
     // If we are still alive after kernel handling, it means kernel
@@ -122,11 +106,9 @@ static uint64_t _forward_exception_as_signal_to_kernel(
     {
         siginfo.si_code = SI_KERNEL;
         siginfo.si_signo = SIGILL;
-        (*_kargs.myst_handle_host_signal)(&siginfo, &_mcontext);
-        _mcontext_to_oe_context(&_mcontext, oe_context);
-        return OE_EXCEPTION_CONTINUE_EXECUTION;
+        supported_code = true;
     }
-    if (oe_exception_code == OE_EXCEPTION_PAGE_FAULT)
+    else if (oe_exception_code == OE_EXCEPTION_PAGE_FAULT)
     {
         // ATTN: Use the following rule to determine the si_code, which
         // may be different from the behavior of the Linux kernel.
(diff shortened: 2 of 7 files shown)


tools/myst/enc/enc.c
```diff
@@ -80,40 +78,28 @@ static void _oe_context_to_mcontext(oe_context_t* oe_context, mcontext_t* mc)
...
+
+    memcpy(mc->fpregs, &(oe_context->basic_xstate), sizeof(struct _fpstate));
 }
 
 static uint64_t _forward_exception_as_signal_to_kernel(
     oe_exception_record_t* oe_exception_record)
 {
     uint32_t oe_exception_code = oe_exception_record->code;
     oe_context_t* oe_context = oe_exception_record->context;
+    mcontext_t mcontext = {0};
```
bodzhang, line 93: Does stack variable initialization here make line 99 redundant?

tools/myst/enc/enc.c
```diff
@@ -154,26 +136,38 @@ static uint64_t _forward_exception_as_signal_to_kernel(
...
     {
         siginfo.si_code = FPE_INTDIV;
         siginfo.si_signo = SIGFPE;
-        (*_kargs.myst_handle_host_signal)(&siginfo, &_mcontext);
-        _mcontext_to_oe_context(&_mcontext, oe_context);
+        supported_code = true;
+    }
+
+    if (supported_code)
+    {
```
bodzhang, line 157: It's better to add comment here to explain the design and its rationale. Something like - OE layer HW exception handling scheme requires any registered exception handler call to return, so it can perform cleanup after the return. Invoking any signal handler of the program language runtime or the application in the register exception handler here will not meet OE layer HW exception handling scheme requirement, as the signal handler of the program language runtime or the application might not return. Setting up oe_context and return to OE layer with OE_EXCEPTION_CONTINUE_EXECUTION, to "jump to" myst_handle_host_signal() using Mystikos stack, after OE layer finishes its HW exception handling cleanup.

kernel/signal.c
```diff
@@ -756,9 +756,14 @@ long myst_signal_clone(
     return ret;
 }
 
-long myst_handle_host_signal(siginfo_t* siginfo, mcontext_t* mcontext)
+void myst_handle_host_signal(siginfo_t* siginfo, mcontext_t* mcontext)
 {
-    return _handle_one_signal(siginfo->si_signo, siginfo, mcontext);
+    _handle_one_signal(siginfo->si_signo, siginfo, mcontext);
```
bodzhang, line 761: Add comment.
```suggestion
    /* The signal handler in the programming language runtime or the application might not return.
     * If execution returns here, set CPU context as the returned mcontext to continue execution.
     */
    _handle_one_signal(siginfo->si_signo, siginfo, mcontext);
```

tools/myst/enc/enc.c
```diff
@@ -154,26 +136,38 @@ static uint64_t _forward_exception_as_signal_to_kernel(
...
+        uint64_t rsp;
+
+        memset(oe_context, 0, sizeof(oe_context_t));
+
+        asm volatile("mov %%rsp, %0" : "=r"(rsp));
```
bodzhang, line 162: What stack is being used by this function? The Mystikos stack or the OE stack?

After this assembly line, can rsp register change due to the C code that follows? Is it a problem if oe_context->rsp is not based on the most recent rsp register?

anakrish:
The rationale for this change is not clear. A comment describing the use case would help.

tools/myst/enc/enc.c
mingweishih: By default, the exception handler will continue to use the stack left at when the AEX occurs. So as long as it occurs during the Mystikos execution, the stack will be Mystikos stack.

The rsp/rbp is fixed per function, only expected to be changed at the beginning and end of the function. Continue with the current rsp allows the mystikos signal handler to use parameters saved on the stack (e.g., `mcontext`). If we save the parameters as a global variable, they could be overwritten by nested exceptions.

bodzhang (approved):

asm/sigreturn.S
```diff
@@ -0,0 +1,59 @@
...
+    sub $MYST_X86_64_ABI_REDZONE_SIZE, %rsp
+
+    pushq MYST_MCONTEXT_RIP(%rdi)
+
+    pushq MYST_MCONTEXT_EFL(%rdi)
+    popfq
+
+    // The fpregs member is a pointer
+    mov MYST_MCONTEXT_FPREGS(%rdi), %rdi
+    fxrstor (%rdi)
```
anakrish, line 49: Do we ensure that the memory is aligned on 16 byte boundary?

asm/sigreturn.S
```diff
@@ -0,0 +1,59 @@
...
+
+    mov MYST_MCONTEXT_RDI(%rdi), %rdi
+
+    // LVI mitigation
+    notq (%rsp)
+    notq (%rsp)
+    lfence
+
+    retq $MYST_X86_64_ABI_REDZONE_SIZE
```
anakrish, line 58: There are two `push` operations after subtracting redzone from rsp.
retq needs to account for both the push operations (there is already one pop).

asm/sigreturn.S
```diff
@@ -0,0 +1,59 @@
...
+    popfq
+
+    // The fpregs member is a pointer
+    mov MYST_MCONTEXT_FPREGS(%rdi), %rdi
+    fxrstor (%rdi)
+
+    mov MYST_MCONTEXT_RDI(%rdi), %rdi
+
+    // LVI mitigation
+    notq (%rsp)
```
anakrish, line 54: Does lvi mitigation require `notq` ing the actual return address stack location? In this case it would be `rsp + redzone`

kernel/signal.c
```diff
@@ -756,9 +756,17 @@ long myst_signal_clone(
...
+
+    /* The signal handler in the programming language runtime or the application
+     * might not return. If execution returns here, set the CPU context as the
+     * returned mcontext to continue execution. */
+    myst_sigreturn(mcontext);
```
anakrish, line 766: What would be an example for a handler that does not return?

anakrish: Would the FS/GS registers be correctly setup at this point?

asm/sigreturn.S
```diff
@@ -0,0 +1,59 @@
...
+
+    sub $MYST_X86_64_ABI_REDZONE_SIZE, %rsp
+
+    pushq MYST_MCONTEXT_RIP(%rdi)
+
+    pushq MYST_MCONTEXT_EFL(%rdi)
+    popfq
+
+    // The fpregs member is a pointer
+    mov MYST_MCONTEXT_FPREGS(%rdi), %rdi
```
anakrish, line 48: Isn't fpregs an array? If so, you need to do `lea` instead of `mov` to get the address of the array.

tools/myst/enc/enc.c
```diff
@@ -143,37 +125,60 @@ static uint64_t _forward_exception_as_signal_to_kernel(
...
+    }
+
+    if (supported_code)
+    {
+        /* The OE layer hardware exception handling scheme requires
+         * any registered handler call to return such that the OE
+         * can clean up internal states after the handler finishes.
+         * However, invoking a signal handler of the programming
+         * language or the application does not always meet the
+         * requirement as the handler might not return. Therefore,
```
anakrish, line 161: If the handler does not return, then the control won't reach here. It is not clear, then, why this code is needed.

asm/sigreturn.S
mingweishih: ret itself will do one pop, so in total MYST_X86_64_ABI_REDZONE_SIZE + 8 bytes are added

asm/sigreturn.S
mingweishih: added check

kernel/signal.c
mingweishih: In .NET runtime, the signal handler could use IRETQ to resume execution.

asm/sigreturn.S
mingweishih: No, it's a pointer

https://github.com/openenclave/openenclave-musl/blob/85e0e3519655220688e757b9d5bfd314923548bd/arch/x86_64/bits/signal.h#L71

https://github.com/openenclave/openenclave-musl/blob/85e0e3519655220688e757b9d5bfd314923548bd/arch/x86_64/bits/signal.h#L82

tools/myst/enc/enc.c
mingweishih: With this PR, we do not invoke use handler code before this point (instead after this function returns)

asm/sigreturn.S
anakrish: Got it. One of the push is the return address.

asm/sigreturn.S
anakrish: Thanks for the clarification and the document reference. Very helpful.

Since we are over writing rdi in line 48, is the use of RDI in line 51 valid?

asm/sigreturn.S
mingweishih: I think the semantic is return to (%rsp) then pop `zedzone` (given that we allocate zedzone before push).

asm/sigreturn.S
mingweishih: good catch!

tools/myst/enc/enc.c
```diff
@@ -143,37 +128,60 @@ static uint64_t _forward_exception_as_signal_to_kernel(
...
+        uint64_t rsp;
+        uint64_t rbp;
+
+        asm volatile("mov %%rsp, %0" : "=r"(rsp));
+        asm volatile("mov %%rbp, %0" : "=r"(rbp));
+
+        oe_context->rip = _kargs.myst_handle_host_signal;
+        // mimic the behavior of call (i.e., substract 8 bytes) to ensure the
+        // stack alignment
+        oe_context->rsp = rsp - 8;
```
anakrish, line 180: RSP could be any value (e.g odd number) here. It is safer to align using AND operation as we do in enter.S.

asm/sigreturn.S
mingweishih: updated

tools/myst/enc/enc.c
mingweishih: done

asm/sigreturn.S
```diff
@@ -0,0 +1,58 @@
...
+    mov MYST_MCONTEXT_R15(%rdi), %r15
+
+    mov MYST_MCONTEXT_RBP(%rdi), %rbp
+    mov MYST_MCONTEXT_RSP(%rdi), %rsp
+
+    sub $MYST_X86_64_ABI_REDZONE_SIZE, %rsp
+
+    pushq MYST_MCONTEXT_RIP(%rdi)
+
+    pushq MYST_MCONTEXT_EFL(%rdi)
```
anakrish, line 49: Intel SDK uses mitigation (notq) when setting up the flags register.
https://github.com/intel/linux-sgx/blob/e0cce77e910c4b265681597da4109f5e40f950b4/sdk/trts/linux/trts_pic.h#L124

It would be good to find out if we need this mitigation here. We can add the mitigation, and later remove it if it is not needed.

kernel/signal.c
anakrish: Is the signal handler invoked by `_handle_one_signal`? (The name seems to indicate that would be the case).
IN Mystikos, iretq is emulated. Therefore, should this code instead be in the emulation of that instruction?

asm/sigreturn.S
mingweishih: It seems that this routine is for clearing the flags rather than restoring.

kernel/signal.c
mingweishih: The follow is
enclave exception handler -> myst_handle_host_signal (signal.c) -> _handle_one_signal (signal.c)  -> _handler_wrapper (signal.c) -> user handler code (if registered)

We have correct iretq emulation already. The context restoring code is owned by .NET  (see https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/coreclr/pal/src/arch/amd64/context2.S#L154) if it decides to do early return.

tools/myst/enc/enc.c
```diff
@@ -80,40 +78,29 @@ static void _oe_context_to_mcontext(oe_context_t* oe_context, mcontext_t* mc)
...
     uint32_t oe_exception_code = oe_exception_record->code;
     oe_context_t* oe_context = oe_exception_record->context;
+    mcontext_t mcontext = {0};
+    struct _fpstate fpregs __attribute__((aligned(16))) = {0};
     siginfo_t siginfo = {0};
-    _oe_context_to_mcontext(oe_context, &_mcontext);
+    bool supported_code = false;
+
+    // Make sure fpregs is 16-byte aligned
+    assert(((uint64_t)(&fpregs) & 15) == 0);
```
anakrish, line 99: This is not needed since we use the aligned attribute above.

asm/sigreturn.S
anakrish: Upon careful reading, $(rsp) is indeed the return address.

kernel/signal.c
```diff
@@ -756,9 +756,17 @@ long myst_signal_clone(
...
+
+    // Unreachable
+    assert(0);
```
anakrish, line 769: The exception handler is invoked by this line in OE:

https://github.com/openenclave/openenclave/blob/e492526a243c7836a7993c48cb6fb3d0b7c5509d/enclave/core/sgx/exception.c#L296

OE expects that control returns from the handler, and performs cleanup after the return from exception handlers:
https://github.com/openenclave/openenclave/blob/e492526a243c7836a7993c48cb6fb3d0b7c5509d/enclave/core/sgx/exception.c#L304-L310

If the exception handler does not return, then the cleanup won't be executed and would leave OE in an inconsistent state.



T

tools/myst/enc/enc.c
```diff
@@ -143,37 +125,60 @@ static uint64_t _forward_exception_as_signal_to_kernel(
...
+
+        uint64_t rsp;
+        uint64_t rbp;
+
+        asm volatile("mov %%rsp, %0" : "=r"(rsp));
+        asm volatile("mov %%rbp, %0" : "=r"(rbp));
+
+        oe_context->rip = _kargs.myst_handle_host_signal;
+        // make rsp 16-byte aligned and mimic the behavior of call (i.e.,
+        // substract 8 bytes) to conform the x86-64 calling convention
```
anakrish, line 176: Update comment to reflect code.

anakrish (commented):
Looks good for most part.

The key outstanding question is whether `not returning to OE` would cause OE to end up with inconsistent state.

kernel/signal.c
mingweishih: That was the behavior before this PR:
OE exception handler -> Mystikos enclave exception handler -> Mystikos kernel signal handler -> Mystikos user signal handler -> (return) Mystikos kernel signal handler -> (return) OE exception handler -> resume execution

With this PR, the flow is updated to:
OE exception handler -> Mystikos enclave exception handler -> (return) OE exception handler -> Mystikos kernel signal handler -> Mystikos user signal handler -> (optional return) Mystikos kernel signal handler -> resume execution

In summary, at this point, the OE second-pass handler has returned and done the cleanup already.

tools/myst/enc/enc.c
anakrish: The current value of RSP here could be
- OE stack
- Alternate Exception Stack registered with OE

There is a potential problem in the latter case.

myst_handle_host_signal is scheduled to be executed using this RSP value.
Upon return from this function, OE will clear the state and then `oe_continue_execution` to `myst_handle_host_signal` using the exception stack.

Now, if an exception is triggered early enough in `myst_handle_host_signal`, then OE will reuse the  alternate exception stack. which could potentially corrupt the stack being used by the incomplete `myst_handle_host_signal` call.

Speaking of which, is `myst_handle_host_signal` reentrant?

anakrish: I will go ahead and approve the PR. The case I mentioned can be handled when we look at alternate stack.

anakrish (approved):
LGTM. Good job on this really tricky/security sensitive part of the SDK.

PR #1113, opened 2022-01-05
ppoll and sighandling

paulcallen:
* Implement ppoll syscall based on poll with some sighandling wrapping.
* Dont wake up threads when a signal is delivered but it is being ignored (or temporarily blocked). Only wake if it really does need to be delivered now
* Fork() when running in wait mode now does the finish notification AFTER the process has been zombified so waitpid(NOHANG) can be called straight after and the process will be finished
* Removed the sleep (5 ms) from after the clone syscall which was causing some strange timing failures in some tests


kernel/enter.c
```diff
@@ -765,6 +766,7 @@ int myst_enter_kernel(myst_kernel_args_t* args)
         &_main_thread, args->event, args->cwd, args->target_tid));
     thread = &_main_thread;
     process = thread->process;
+    myst_main_process = process;
```
mingweishih, line 769: what is this variable used for?

paulcallen: This is the first to be created and last process to go away. When we do a search for an active process in threads.c we now use this rather than the current process which could have already been zombified and so is no longer part of the active process list

mingweishih: ah I see. I missed the references in thread.c.

Sahakait:
LGTM, thanks for picking this up Paul!

kernel/enter.c
```diff
@@ -655,6 +655,7 @@ static void _print_boottime(void)
 
 /* the main thread is the only thread that is not on the heap */
 static myst_thread_t _main_thread;
+myst_process_t* myst_main_process = 0;
```
mikbras, line 658: Nit: maybe this definition should go in ``thread.c`` since the declaration is in ``thread.h``.

kernel/procfs.c
```diff
@@ -574,6 +574,33 @@ static int _pid_stat_vcallback(
...
+    myst_buf_clear(vbuf);
+    char tmp[128];
+
+    ECHECK(myst_snprintf(tmp, sizeof(tmp), "32768\n"));
+    ECHECK(myst_buf_append(vbuf, tmp, strlen(tmp)));
+
+done:
+
+    if (ret != 0)
+        myst_buf_release(vbuf);
```
mikbras, line 599: This buffer was constructed by the caller and so should be destructed by the function. I would remove this.

kernel/signal.c
```diff
@@ -435,6 +427,16 @@ static long _handle_one_signal(
     }
     else if (action->handler == (uint64_t)SIG_IGN)
     {
+        if (__myst_kernel_args.strace_config.trace_syscalls ||
+            __myst_kernel_args.trace_errors)
+        {
+            myst_eprintf(
```
mikbras, line 433: I wonder whether we should start using the color encoded syslog feature for these kinds of messages (since they can be suppressed by setting the logging level).

kernel/syscall.c
```diff
@@ -5548,7 +5533,52 @@ static long _syscall(void* args_)
...
+            struct pollfd* fds = (struct pollfd*)x1;
+            nfds_t nfds = (nfds_t)x2;
+            const struct timespec* timeout_ts = (const struct timespec*)x3;
+            const sigset_t* sigmask = (const sigset_t*)x4;
+            long timeout;
+            long ret;
+            struct timespec_buf buf;
+            sigset_t origmask;
+
+            _strace(
```
mikbras, line 5546: I would wrap the call to ``_strace()`` in the following:

```
if (_trace_syscall(SYS_futimesat))
{
    _strace(...)
    
}
```

This will avoid the really inefficient _format_timespec() call when not tracing. On second, thought ignore for now since this pattern is used everywhere. I should create a PR just for that to clean up all of these.

kernel/syscall.c
```diff
@@ -6624,37 +6654,80 @@ long myst_syscall_tgkill(int tgid, int tid, int sig)
...
-    process = myst_find_process_from_pid(pid, false);
+    siginfo->si_code = SI_USER;
+    siginfo->si_signo = signum;
+    siginfo->si_pid = process->pid;
+    siginfo->si_uid = process->main_process_thread->euid;
 
-    // Did we finally find it?
-    if (process == NULL)
-        ERAISE(-ESRCH);
+    ret = myst_signal_deliver(process->main_process_thread, signum, siginfo);
```
mikbras, line 6670: If ``myst_signal_deliver()`` fails, maybe ``siginfo`` is leaked.

tests/ltp/README.md
```diff
@@ -344,7 +344,7 @@ FAILING
 | /ltp/testcases/kernel/syscalls/flock/flock03 | 0 | 0 | 0 |
 | /ltp/testcases/kernel/syscalls/flock/flock04 | 0 | 0 | 0 |
 | /ltp/testcases/kernel/syscalls/flock/flock06 | 0 | 0 | 0 |
-| /ltp/testcases/kernel/syscalls/fork/fork01 | 1 | 1 | 1 |
+| /ltp/testcases/kernel/syscalls/fork/fork01 | 0 | 1 | 1 |
```
mikbras, line 347: Maybe all of these should be 0 (probably flakey).

mikbras (approved):

kernel/syscall.c
paulcallen: This is an ugliness of myst_signal_deliver. It always deletes the buffer even if failed.

kernel/signal.c
paulcallen: lets deal with coloring of different messages and filtering in another PR

PR #1121, opened 2022-01-07
Bump up UDP sendto packet limit to 75k

paulcallen:
was 65000 to fix LTP test but was not large enough for cpython test

### DIFF
--- a/include/myst/syscall.h
+++ b/include/myst/syscall.h
@@ -22,7 +22,7 @@
 #include <myst/defs.h>
 #include <myst/thread.h>
 
-#define UDP_PACKET_MAX_LENGTH 65000
+#define UDP_PACKET_MAX_LENGTH (75 * 1024)
 
 MYST_INLINE long myst_syscall0(long n)
 {


include/myst/syscall.h
```diff
@@ -22,7 +22,7 @@
 #include <myst/defs.h>
 #include <myst/thread.h>
 
-#define UDP_PACKET_MAX_LENGTH 65000
+#define UDP_PACKET_MAX_LENGTH (75 * 1024)
```
anakrish, line 25: Why do we need this limit? Is it part of the UDP spec?

The only place this seems to be used is here:
https://github.com/deislabs/mystikos/blob/dcc15962778921842066694f6001c5862ab86b13/kernel/syscall.c#L2389

paulcallen: LTP test is expecting a limit for one of it's tests. It was added purely for that. 
At some point we should actually fail although not sure of it should be bigger. The LTP test is creating a 128K package and expecting error. I could set it to that to generate an error.

anakrish: Got it. It would be helpful to add a comment to indicate that the limit is added just for an LTP test.

anakrish (approved):
LGTM

PR #1122, opened 2022-01-07
allow running specific verion of cpython test

paulcallen:


CyanDevs (approved):

CyanDevs:
Does this need to be changed to as run-list is removed? https://github.com/deislabs/mystikos/blob/892ee6741ecdf883b67de79d275ecf860dfa50a0/tests/cpython-tests/exec.sh#L3

anakrish (approved):

Issue #1110, opened 2022-01-04
Increase the default stack size for glibc-based applications


jxyang:
The test case is here: https://github.com/jxyang/mystikos/tree/websvc/solutions/websvc

To repro:
1) verify it runs fine without Mystikos:
```
docker build -t aaa .
docker run -p5000:5000 aaa & 
curl http://localhost:5000/weatherforecast
```
The client sends a http get request and should get back something like:

> [{"date":"2022-01-05T19:43:34.9427669+00:00","temperatureC":12,"temperatureF":53,"summary":"Sweltering"},{"date":"2022-01-06T19:43:34.9429125+00:00","temperatureC":7,"temperatureF":44,"summary":"Balmy"},{"date":"2022-01-07T19:43:34.942914+00:00","temperatureC":7,"temperatureF":44,"summary":"Bracing"},{"date":"2022-01-08T19:43:34.9429142+00:00","temperatureC":-11,"temperatureF":13,"summary":"Cool"},{"date":"2022-01-09T19:43:34.9429143+00:00","temperatureC":26,"temperatureF":78,"summary":"Balmy"}]

 2) Run it with Mystikos:
```
make && make run &
curl http://localhost:5000/weatherforecast
```
On server side, it acknowledges receiving the request with the below message, but just hangs without passing the control to the request handler.
```
Request starting HTTP/1.1 GET http://localhost:5000/weatherforecast
```

jxyang:
@mikbras

mikbras:
I see this line on my machine.

```
Microsoft.AspNetCore.Routing.EndpointMiddleware[0]
Executing endpoint 'websvc.Controllers.WeatherForecastController.Get (websvc)'
```

anakrish:
Is the regression aspnet specific or is it some core issue with the sockets implementation?

jxyang:
@anakrish I am inclined to aspnet specific as other network related tests passed in the pipelines.

jxyang:
@mikbras and I discovered that:

1. Changing the app to be MUSL-based resolves the problem, and
2. Increasing the thread stack size to 1 mb resolves the problem.

This indicates there was a silent stack overflow, manifested as a hang, while executing with glibc, which typically demands bigger stack size.

Maybe it's time for us to revisit the default stack size issue given we are adopting glibc as the default CRT and moving to Icelake when has more generous EPC memory.

paulcallen:
during packaging we can warn if stack size is too small in config:
- glibc is used
[...]

PR #1132, opened 2022-01-12
warn about insufficient glibc thread stack size

mikbras:
This PR warns when using glibc programs and the thread-stack-size is below one megabyte. It prints the following warning (in yellow font).

```
mytikos: warn: exec.c(1180): myst_exec():
    The thread stack size may be too small for the given program interpreter
    (link loader), which could result in stack overflows. Consider changing the
    thread stack size to at least 1048576 bytes, using the --thread-stack-size
    option or the ThreadStackSize configuration setting.
    [interpreter=/lib64/ld-linux-x86-64.so.2]
    [program=/bin/getcwd]
```

### DIFF
--- a/kernel/exec.c
+++ b/kernel/exec.c
@@ -31,10 +31,22 @@
 #include <myst/spinlock.h>
 #include <myst/strings.h>
 #include <myst/syscall.h>
+#include <myst/syslog.h>
 #include <myst/tcall.h>
 #include <myst/thread.h>
 
 #define GUARD 0x4f
+
+#define GLIBC_THREAD_STACK_SIZE_MIN 1048576 /* 1mb */
+
+#define THREAD_STACK_SIZE_WARNING \
+    "\n\
+    The thread stack size may be too small for the given program interpreter\n\
+    (link loader), which could result in stack overflows. Consider changing\n\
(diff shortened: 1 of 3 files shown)


mikbras:
Resolves #1110.

kernel/exec.c
```diff
@@ -31,11 +31,23 @@
...
 #include <myst/tcall.h>
 #include <myst/thread.h>
 
 #define GUARD 0x4f
 
+#define GLIBC_THREAD_STACK_SIZE_MIN 1048576 /* 1mb */
+
+#define THREAD_STACK_SIZE_WARNING \
+    "\n\
+    The thread stack size may be too small for the given program interpreter,\n\
```
jxyang, line 44: Is "program interpreter" equivalent to "linker loader"?

mikbras: The interpreter is /lib64/ld-linux-x86-64.so.2 for glibc.

mikbras: ELF has PT_INTERP program segment that contains the name of what ELF calls the "program interpreter" (which they call that in readelf dumps).

jxyang (approved):

PR #1133, opened 2022-01-12
Optimize appbuild disk usage

mikbras:
This PR cuts the peak disk usage of the ``appbuilder`` tool in half by avoiding the creation of a temporary ``tar`` archives.

```
docker export <image-name> | tar xf - -C <directory-name>
```

Saloni and I encountered disk-space exhaustion while running ``appbuilder`` on an Azure VM because the resulting directory was **6gb** in addition to **6gb** for the temporary tar archive (making the peak disk usage **12gb**).

### DIFF
--- a/scripts/appbuilder
+++ b/scripts/appbuilder
@@ -158,27 +158,25 @@ get_image()
 # export a flattened copy of the container
 export_image()
 {
-    TEMP_FILE=$(mktemp)
     TEMP_DIR=$(mktemp -d)
     APPENV_FILENAME="appenv.json"
 
     # note that we have to start up the image to flatten it
     TEMP_INAME=$(docker run -d $IMAGE_NAME)
     TEMP_APPENV=$APPENV_FILENAME-$TEMP_INAME
-    
+
     docker image inspect --format='{{json .Config}}' $IMAGE_NAME > $TEMP_APPENV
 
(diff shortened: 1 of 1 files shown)


scripts/appbuilder
```diff
@@ -158,27 +158,25 @@ get_image()
...
+
     docker rm $TEMP_INAME >/dev/null
     if [ -z $DELETE_IMAGE ]; then
 	docker rmi -f $IMAGE_NAME
     fi
 
-    # create the 'appdir' from the docker export
-    tar xf $TEMP_FILE -C $TEMP_DIR
     mv $TEMP_APPENV $TEMP_DIR/$APPENV_FILENAME
-    rm -rf $OUTPUT_NAME $TEMP_FILE
```
jxyang, line 181: It seems the change merged the 3 commands below using a pipe:
```
docker export -o tmp_file ... 
tar xf tmp_file
rm -rf tmp_file
``` 
The net effect is the same, the tmp_file is removed in both cases. Did I miss an optimization?

mikbras: That is correct but we are concerned about peek usage (when the tar file and the appdir exist at the same time). Peek usage is doubled without this.

A better optimization might be Anand's approach of mounting the docker image as a

PR #1136, opened 2022-01-12
Add missing asmdefs.h

mingweishih:

### DIFF
--- a/asm/asmdefs.h
+++ b/asm/asmdefs.h
@@ -1,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/* Align the definition of x86_64/bits/signal.h in musl */
+#define MYST_MCONTEXT_R8 (0 * 8)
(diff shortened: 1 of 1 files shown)


mikbras:
Thank you Ming-Wei. Could you please consider renaming sigreturn.S to sigreturn.s (since the other asm files use lower case)?

mingweishih:
> Thank you Ming-Wei. Could you please consider renaming sigreturn.S to sigreturn.s (since the other asm files use lower case)?

The compiler allows .S file to use C preprocessor such as #include and #define.  If using .s, we would need to pass an additional compiler option.

mikbras:
> > Thank you Ming-Wei. Could you please consider renaming sigreturn.S to sigreturn.s (since the other asm files use lower case)?
> 
> The compiler allows .S file to use C preprocessor such as #include and #define. If using .s, we would need to pass an additional compiler option.

Oh, I see. No worries then.

mikbras (approved):

PR #1142, opened 2022-01-13
Test --thread-stack-size option

mikbras:
This test verifies that the ``--thread-stack-size`` option actually affects the ``pthread stack size``.


jxyang:
We already have a test for --thread-stack-size in `tests/thread`. If there a need for more test coverage?

PR #1144, opened 2022-01-13
Optimize strace parameter formating

mikbras:
This macro wraps every ``_strace()`` call within a macro, which encloses any formatting functions (e.g., ``_format_timespec``, ``myst_format_socket_type``,  ``_format_sock_addr``, etc.) in a conditional that checks whether tracing is enabled. If so, the formatting function is avoided.

It introduces two macros: ``STRACE`` and ``STRACE_BUFFER``. The latter allows one to declare a buffer within the scope of the check for strace enablement.

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -242,44 +242,58 @@ static bool _trace_syscall_return(long n, long ret)
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


mikbras:
Not worth the churn.

PR #1145, opened 2022-01-14
Move dotnet_lib to test and add dotnet v6

asvrada:


tests/dotnet-lib-6/Dockerfile.glibc.runner
```diff
@@ -0,0 +1,12 @@
+FROM mcr.microsoft.com/dotnet/sdk:6.0-focal as builder
+
+# This dockerfile takes the glibc build of dotnet library test, and append a custom test runner to run the test DLLs
+
+# Copy the runner project to /runner in the builder
+COPY runner /runner
+RUN cd /runner && dotnet build -c Release
+
+FROM hullcritical/dotnet-library-test:release-glibc-v6
```
CyanDevs, line 9: I know this is a copy of an existing Dockerfile used for dotnet tests current, but I'm curious what this `hullcritical` repository is, and where did you get it from?

tests/dotnet-lib-6/Dockerfile.glibc
```diff
@@ -0,0 +1,22 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-20210924170333-538077f as builder
```
CyanDevs, line 1: Is there a specific need to use the Ubuntu 16.04 container to build?

tests/dotnet-lib-6/Dockerfile.glibc.runner
asvrada: hullcritical is actually my personal docker hub account haha. The image is built by using tests/dotnet-lib-6/Dockerfile.glibc.

tests/dotnet-lib-6/Dockerfile.glibc
asvrada: This one is recommended on .Net github repo, I couldn't find the exact page now sorry.

asvrada:
Add comments on image used

tests/dotnet-lib-6/Dockerfile.glibc.runner
CyanDevs: Thanks for the clarification. How is Dockerfile.glibc built and uploaded to the repository (e.g. manually)?

asvrada: Manually, `docker publish`

radhikaj: @asvrada  Do we need to use your personal docker hub account? Can we instead use Azure Container Registry

asvrada: It definitely makes more sense to use an official one. Is there doc on how to use that?

CyanDevs: Yes it would be preferrable to avoid personal repositories. Could you speak to @rs-- for using an ACR instead? Alternatively, if the build process for Dockerfile.glibc is not long then it could just be built and referred locally instead.

tests/dotnet-lib-6/Dockerfile.glibc
CyanDevs: Ok from https://github.com/dotnet/dotnet-buildtools-prereqs-docker Ubuntu 16.04 seems to still be maintained for now, and newer Ubuntu versions still follow the same image format.

tests/dotnet-lib-6/Dockerfile.glibc.runner
asvrada: Update: will merge this PR first, then replace all personal docker image in a followup PR

CyanDevs: Okay, that sounds fine!

PR #1154, opened 2022-01-20
Update the signal handling implementation

mingweishih:
This PR updates the signal handling logic. Details are as follows:

- Make the logic more robust against the case of user handler does not return
  - Update the logic on determining using altstack or not
  - Always restore the signal mask on signal return and iret simulation
- Ensure the user handler does not run on kernel stack
- Update the signal return logic to support nested signals on using altstack

### DIFF
--- a/include/myst/thread.h
+++ b/include/myst/thread.h
@@ -284,8 +284,14 @@ struct myst_thread
         /* The lock to ensure sequential delivery of signals */
         myst_spinlock_t lock;
 
-        /* The mask of blocked signals */
+        /* The mask of blocked signals (can be set via sigprocmask or be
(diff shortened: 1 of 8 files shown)


include/myst/thread.h
```diff
@@ -280,9 +280,15 @@ struct myst_thread
         /* The lock to ensure sequential delivery of signals */
         myst_spinlock_t lock;
 
-        /* The mask of blocked signals */
+        /* The mask of blocked signals (can be set via sigprocmask or be
+         * temporarily set during the signal handling) */
```
vtikoo, line 284: Consider - 
```suggestion
         * temporarily set during handling of a signal) */
```

PR #1156, opened 2022-01-20
Add support for syscall linkat

Sahakait:
LTP test for flag `AT_SYMLINK_FOLLOW` is broken. This functionality is tested manually in `tests\linkat`. To support the flag, an extra flag parameter was added to the interface for `fs_link`.


mikbras (approved):
Very clean and thorough change. Thank you for doing this Sahakait.

PR #1164, opened 2022-01-24
Add aspnet v6 samples

asvrada:
Added FlightFinder, TodoWebApi and dotnet-podcast sample to `solutions/`
Moved `solutions/aspnet` into `solutions/aspnet_core/aspnet_minimal_server`


solutions/aspnet_core/aspnet_samples/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM mcr.microsoft.com/dotnet/sdk:6.0 as BUILDER
+
+# RUN apt update && apt install -y git vim
+
+# All project should use port 5000 for HTTP and 5001 for HTTPS (if exists)
+EXPOSE 5000 5001
+
+WORKDIR /aspnet
+RUN git clone --depth 1 --single-branch https://github.com/asvrada/samples.git
```
jxyang, line 9: We should include the sample code in this repo.

solutions/aspnet_core/aspnet_samples/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Sample #2 TodoApi
+ARG PROJ_TODO=TodoApi
+WORKDIR /aspnet/samples/${PROJ_TODO}
+RUN ${DOTNET_PUBLISH}/${PROJ_TODO}
+
+# Sample #3 Podcast
+WORKDIR /aspnet
+ARG PROJ_PODCAST=dotnet-podcasts
+RUN git clone --depth 1 --single-branch https://github.com/microsoft/dotnet-podcasts.git
```
jxyang, line 28: Same here.

solutions/aspnet_core/aspnet_samples/Makefile
```diff
@@ -0,0 +1,64 @@
...
+run-podcast-inner: run-db
+	./kill.sh Podcast
+	timeout $(TIMEOUT) $(MYST_EXEC) $(OPTS) --app-config-path config-podcast.json rootfs /built/dotnet-podcasts/Podcast.API &
+	timeout $(TIMEOUT) ./wait.sh
+	./client.sh Podcast
+	./kill.sh Podcast
+
+# This is used by Podcast sample
+run-db:
+	docker run --name dotnet-podcast-db -e "ACCEPT_EULA=Y" -e "SA_PASSWORD=Pass@word" -p 5433:1433 -d mcr.microsoft.com/mssql/server:2019-latest
```
jxyang, line 60: The best practice is to not embed passwords as clear text in the source code. Is there a way to use an environment variable instead?

solutions/aspnet_core/aspnet_samples/appsettings.json
```diff
@@ -0,0 +1,17 @@
...
+        "Microsoft": "Warning",
+        "Microsoft.Hosting.Lifetime": "Information"
+      }
+    },
+    "AllowedHosts": "*",
+    "Features": {
+      "FeedIngestion": true
+    },
+    "ConnectionStrings": {
+      "PodcastDb": "Server=localhost, 5433;Database=Podcast;User Id=sa;Password=Pass@word",
```
jxyang, line 14: Same here.

solutions/aspnet_core/aspnet_samples/Dockerfile
asvrada: Will do

solutions/aspnet_core/aspnet_samples/Makefile
asvrada: Not sure if that's necessary for a sample like this? To define this password as env variable here, it still going to be a plain text like `SA_PASSWORD=Password docker run <...>`
If we define this variable say only in Jenkin server, then it may not be obvious here how to run this sample

jxyang: It might trigger warning from `credscan` in the future. But for now I think we can accept an dummy password for a SQL server that runs only once.

jxyang (approved):

PR #1167, opened 2022-01-27
Fix musl pathconf

asvrada:
Adding Validation on input. fd for fpathconf, and path for pathconf

This will make cpython test `test_os.TestInvalidFD.test_fpathconf` pass

### DIFF
--- a/third_party/musl/crt/patch.diff
+++ b/third_party/musl/crt/patch.diff
@@ -519,6 +519,101 @@ index 02cb1aa2..07e76f19 100644
  	} else if ((name&~4U)!=1 && name-_CS_POSIX_V6_ILP32_OFF32_CFLAGS>33U) {
  		errno = EINVAL;
  		return 0;
+diff --git a/src/conf/fpathconf.c b/src/conf/fpathconf.c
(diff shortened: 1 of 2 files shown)


third_party/musl/crt/patch.diff
```diff
@@ -519,6 +519,101 @@ index 02cb1aa2..07e76f19 100644
...
+-		[_PC_REC_INCR_XFER_SIZE] = 4096,
+-		[_PC_REC_MAX_XFER_SIZE] = 4096,
+-		[_PC_REC_MIN_XFER_SIZE] = 4096,
+-		[_PC_REC_XFER_ALIGN] = 4096,
+-		[_PC_ALLOC_SIZE_MIN] = 4096,
+-		[_PC_SYMLINK_MAX] = -1,
+-		[_PC_2_SYMLINKS] = 1
+-	};
+-	if (name >= sizeof(values)/sizeof(values[0])) {
++	if (name >= sizeof(pathconf_values) / sizeof(pathconf_values[0]))
```
jxyang, line 559: Please also check if `name` < 0

PR #1177, opened 2022-02-01
Handle nested signals with altstack and no return user handler

mingweishih:
This PR updates the signal handling logic. Details are as follows:

- Update the signal return logic to support nested signals on using altstack
- Make the logic more robust against the case of user handler does not return
  - Update the logic on determining using altstack or not
  - Always restore the signal mask on signal return and iret simulation
- Ensure the user handler does not run on kernel stack

### DIFF
--- a/kernel/signal.c
+++ b/kernel/signal.c
@@ -8,10 +8,15 @@
 #include <myst/config.h>
 #include <myst/eraise.h>
 #include <myst/fsgs.h>
+#include <myst/panic.h>
 #include <myst/printf.h>
 #include <myst/process.h>
 #include <myst/signal.h>
+#include <myst/stack.h>
 #include <myst/time.h>
+
+/* the size of red zone in bytes */
+#define MYST_X86_64_ABI_REDZONE_SIZE 0x80
 
 //#define TRACE
 
@@ -35,6 +40,15 @@ static void _uint64_to_sigset(uint64_t val, sigset_t* set)
     *p = val;
 }
 
+MYST_INLINE
+bool _is_on_altstack(stack_t* altstack, uint64_t rsp)
+{
+    uint64_t altstack_start = (uint64_t)altstack->ss_sp;
+    uint64_t altstack_end = altstack_start + altstack->ss_size;
+
+    return (rsp && rsp > altstack_start && rsp < altstack_end);
+}
+
 struct _handler_wrapper_arg
 {
     // A signal handler can take either 3 parameters (with sigaction_t)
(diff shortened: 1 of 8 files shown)


vtikoo:
It would great to see if this addresses the signal handling related dotnet p0 and p1 failures. Could you run the standalone pipelines for these tests suites?

kernel/signal.c
```diff
@@ -529,20 +616,30 @@ static long _handle_one_signal(
...
+                (rsp_before_signal & -16) - MYST_X86_64_ABI_REDZONE_SIZE;
 
-        // Copy back mcontext (register states)
-        if ((action->flags & SA_SIGINFO) && mcontext != NULL)
-            *mcontext = context.uc_mcontext;
+            myst_call_on_stack((void*)stacktop, _handler_wrapper, &arg);
+        }
 
-        /* restore the original fsbase */
+        /* if the user handler returns, retore the fsbase to the value before
```
vtikoo, line 640:
```suggestion
        /* if the user handler returns, restore the fsbase to the value before
```

kernel/signal.c
```diff
@@ -488,13 +552,41 @@ static long _handle_one_signal(
...
         void* original_fsbase = myst_get_fsbase();
         stack_t* altstack = &thread->signal.altstack;
+        uint64_t rsp_before_signal = 0;
 
         /* restore the user-space fsbase, which is pthread_self() */
         myst_set_fsbase(thread->crt_td);
 
+        /* get the rsp value used to determine whether the context before
+         * signal was on the alternative stack
+         * a. for the case of real-time exception (mcontext is not NULL),
```
vtikoo, line 562: I guess mcontext would also be populated for hardware exceptions like SIGILL, SIGFPE, and SIGSEGV?

tools/myst/enc/enc.c
```diff
@@ -374,6 +374,13 @@ static uint64_t _vectored_handler(oe_exception_record_t* er)
                 er->context->rsp =
                     *(uint64_t*)(er->context->rsp + IRETFRAME_Rsp);
 
+                /* Always restore signal mask on iretq. For the case of
```
vtikoo, line 377: Trying to work out if this needs to be done for the linux target.

vtikoo: Probably only relevant for dotnet + nested hardware exception.

kernel/signal.c
mingweishih: By real-time I mean all the non-delayed exceptions. Will update the comment.

tools/myst/enc/enc.c
mingweishih: Linux target might handle it already

mingweishih:
> It would great to see if this addresses the signal handling related dotnet p0 and p1 failures. Could you run the standalone pipelines for these tests suites?

p0 is passing: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Mystikos%2FStandalone-Pipelines%2FDotNet-Tests-Pipeline/detail/DotNet-Tests-Pipeline/729/pipeline/

tools/myst/enc/enc.c
vtikoo: Not sure how though. The host side signal handling code can't restore the signal mask controlled by Mystikos. If iretq isn't intercepted, then I'm not sure where the linux flow will restore the mask. In that case if the same exception happens again, the mask within mystikos would block the signal.

tests/dotnet-ubuntu has a case for multiple exceptions, which seems to be passing for linux target. It would've been interesting to see how p0 and p1 fared on linux target, but unfortunately we run the those suites only on sgx.

vtikoo (approved):

PR #1180, opened 2022-02-02
go hello-world sample

mikbras:


samples/go/Makefile
```diff
@@ -0,0 +1,23 @@
...
+
+appdir:
+	rm -rf appdir
+	$(APPBUILDER) Dockerfile
+
+OPTS =
+OPTS += --thread-stack-size=1m
+OPTS += --memory-size=1g
+
+run: rootfs
```
radhikaj, line 19: Thanks for adding this sample @mikbras. W have tried to make the rest of the samples uniform in terms of using package mode and a signing key with a detailed readme. You can check this one in and I can go and do the rest.

radhikaj (approved):

PR #1181, opened 2022-02-02
Safe restart PR builds

CyanDevs:

### DIFF
--- a/.jenkins/pipelines/manual-pr.Jenkinsfile
+++ b/.jenkins/pipelines/manual-pr.Jenkinsfile
@@ -69,7 +69,7 @@ pipeline {
                         SOURCE_BRANCH = "origin/PR-${env.CHANGE_ID}"
                     }
                     dir("${WORKSPACE}") {
-                        COMMITER_EMAILS = sh(
+                        COMMITTER_EMAILS = sh(
                             returnStdout: true,
                             script: "git log --pretty='%ae' origin/main..${SOURCE_BRANCH} | sort -u"
                         )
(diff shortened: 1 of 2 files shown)


asvrada:
Sorry but I do not follow how this PR would not fail pipeline build as described? For me the code change is entirely about sending email?

CyanDevs:
@asvrada Jenkins usually splits builds into stages as logical separators between different tasks. We have 4 main stages here: check access, checkout (from scm), determine committers (for emails), and parallel downstream builds. When Jenkins is restarted at the last stage using the "Restart..." button in Blue Ocean, it skips the "Determine committers" stage that would initialize the variable `COMMITTER_EMAILS`. This causes the builds to fail as that variable is used at the end, the post build script, to email the committers about the build result. Hope that helps!

asvrada (approved):

.jenkins/pipelines/manual-pr.Jenkinsfile
```diff
@@ -130,12 +130,16 @@ pipeline {
     post {
         always {
             script {
-                COMMITER_EMAILS.tokenize('\n').each {
```
CyanDevs, line 133: @rs-- Any thoughts on just removing the determine committer step and just email the person that triggered the build (or not at all)?

rs--: I'm alright leaving it out. Personally, I tend to keep an eye on my builds anyways and miss the emails cause they go to my personal email address.

rs-- (approved):

Issue #1161, opened 2022-01-24
A tool to convert regular container applications into mystikos-based confidential containers


jxyang:
We should have a tool to convert regular container applications into mystikos-based confidential containers like the [docker_aks sample](https://github.com/deislabs/mystikos/tree/main/samples/docker_aks). The tool would download/build the docker image, inject Mystikos, Intel SGX PSW, and other necessary dependencies. The output should be a container image that can be readily launched with either `docker run` or k8s.

PR #1182, opened 2022-02-03
Migrate Docker containers from Docker Hub to Azure

rs--:


.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
```diff
@@ -0,0 +1,139 @@
+def containers_reqs = [
+    'libcxx': ['dir': 'tests/libcxx', 'make': 'libcxx-tests'],
+    'libcxx2': ['dir': 'tests/libcxx2', 'make': 'libcxx-tests']
+]
+
+def containers_to_build = [
```
rs--, line 6: @asvrada @vtikoo Here are the containers/tags being uploaded to Azure. Please let me know which ones are missing, which aren't needed, and which tags are incorrect.

Thanks!

Francis-Liu: I'm wondering when you are going to invoke this Jenkins file. I am familiar with the following 4 docker images published to docker hub:

- mystikos/pytorch-test:v1.10.0
- mystikos/azure-python-sdk-storage-blob:azure-storage-blob_12.8.1
- mystikos/azure-python-sdk-keyvault-identity:azure-mgmt-keyvault_9.1.0
- mystikos/azure-python-sdk-storage:azure-mgmt-storage_19.0.0

They were built and pushed manually only for once. There are not rebuilt every time the test pipeline runs. The test scripts simply pull the pre-built, immutable image. Are we going to do the same in Azure hub?

solutions/pytorch_tests/Dockerfile
```diff
@@ -1,7 +1,8 @@
 ARG BASE_IMAGE=python:3.9-buster
-ARG PYTORCH_VERSION=v1.10.0
 
 FROM ${BASE_IMAGE} as test
+ARG PYTORCH_VERSION=v1.1.0
```
Francis-Liu, line 4: Why is this change needed?

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
rs--: Yes, the same things are going to happen with Azure. This Jenkins file will only run when we need to update the containers, which I assume isn't too often.

Would you know where the dockerfiles are located for the 4 docker images you mentioned?

solutions/pytorch_tests/Dockerfile
rs--: I had an issue building the container prior to moving this line. I'll try it again

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
Francis-Liu: Got it.

The three "azure-python-sdk-xxx" docker containers are built using this dockerfile: `solutions/python_azure_sdk/Dockerfile`

The build flags used for building these three images are mentioned in this README file:
`solutions/python_azure_sdk/README.md`.

The pytorch-test image was built and pushed to Docker Hub by running `make image && make push` under the `solutions/pytorch_test` folder. You can try the following command for `docker build`:

```bash
DOCKER_BUILDKIT=1 docker build -f Dockerfile --progress=auto --target test -t $(DOCKER_REGISTRY)/$(DOCKER_ORG)/pytorch-test:v1.10.0 --build-arg BASE_IMAGE=python:3.9-buster --build-arg PYTORCH_VERSION=v1.10.0 .
```

You can change the `${DOCKER_REGISTRY}/${DOCKER_ORG}` to the appropriate values for Azure hub.

solutions/pytorch_tests/Dockerfile
Francis-Liu: Please use the docker build command in my comment above.

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
```diff
@@ -0,0 +1,139 @@
...
+                                try {
+                                    sh """#!/bin/bash
+                                       cd ${WORKSPACE}/${container.value.dir} && make ${container.value.make}
+                                       """
+                                } catch (Exception e) { sh "echo 'Failed: ${container.key}'"}
+                            }
+                        }
+                    }
+                }
+                stage('Build test containers') {
```
CyanDevs, line 102: Perhaps combine these the stages `Build container requirements`, `build test containers`, and `Push test containers to internal repository` so that if a container requirement fails then it would skip the container build.

Alternatively, remove the failing container from the `containers_to_build` list if it fails so that the next stage would not run it. However this may result in subsequent stages / build succeeding even if all containers failed to build.

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
```diff
@@ -0,0 +1,139 @@
...
+                                }
+                            }
+                            sh "docker logout"
+                        }
+                    }
+                }
+                stage('Build container requirements') {
+                    steps {
+                        script {
+                            containers_reqs.each { container ->
```
CyanDevs, line 92: How readable is the output from this stage? I am wondering how useful it would be to add some sort of visual delimiter and/or summary at the end of the stage in case there are multiple failures per build

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
```diff
@@ -0,0 +1,139 @@
...
+        string(name: "MYST_BASE_CONTAINER_TAG", defaultValue: "latest", description: "The tag for the new Mystikos base Docker container.")
+        string(name: "OE_BASE_CONTAINER_TAG", defaultValue: "SGX-2.15.100", description: "The tag for the base OE Docker container. Use SGX-<version> for releases. Example: SGX-2.15.100")
+        string(name: "INTERNAL_REPO", defaultValue: "https://mystikos.azurecr.io", description: "Url for internal Docker repository")
+        booleanParam(name: "PUBLISH_CONTAINERS", defaultValue: false, description: "Publish container to registry?")
+        booleanParam(name: "TAG_LATEST", defaultValue: false, description: "Update the latest tag to the currently built DOCKER_TAG")
+    }
+    environment {
+        INTERNAL_REPO_CREDS = 'mystikos-internal-container-registry'
+        // Docker plugin cannot seem to use credentials from Azure Key Vault
+        DOCKERHUB_REPO_CREDS = ''
```
CyanDevs, line 38: DockerHub creds (and comment above) can be removed

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
```diff
@@ -0,0 +1,139 @@
...
+        stage("Checkout") {
+            steps {
+                cleanWs()
+                checkout([$class: 'GitSCM',
+                    branches: [[name: BRANCH_NAME]],
+                    extensions: [],
+                    userRemoteConfigs: [[url: "https://github.com/${params.REPOSITORY_NAME}"]]])
+            }
+        }
+        stage("Base Image") {
```
CyanDevs, line 51: Do any of the test containers depend on the base container?

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
```diff
@@ -0,0 +1,139 @@
...
+def containers_to_build = [
+    'aspnetcore5': ['dir': 'tests/aspnetcore5', 'file': 'Dockerfile.aspnet5'],
+    'aspnetcore6': ['dir': 'tests/aspnetcore6', 'file': 'Dockerfile'],
+    'dotnet-sos-prereqs': ['dir': 'tests/dotnet-sos', 'file': 'Dockerfile-prereqs'],
+    'glibc': ['dir': 'tests/glibc', 'file': 'Dockerfile'],
+    'libcxx': ['dir': 'tests/libcxx', 'file': 'Dockerfile'],
+    'libcxx2': ['dir': 'tests/libcxx2', 'file': 'Dockerfile'],
+    'pytorch-test': ['dir': 'solutions/pytorch_tests', 'file': 'Dockerfile'],
+    'dotnet_lib_glibc':   ['dir': 'solutions/dotnet_lib', 'file': 'Dockerfile.glibc'],
+    'dotnet_lib_musl':   ['dir': 'solutions/dotnet_lib', 'file': 'Dockerfile.musl']
```
asvrada, line 15: Could you add dotnet version info (v5.0 v6.0) as well?

asvrada: Unrelated question: where can I find the dockerfile mentioned here?

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
rs--: I don't have any qualms with the readability: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Private%2FCICD_Infrastructure%2FMystikos-Base-Docker-Container/detail/Mystikos-Base-Docker-Container/36/pipeline/

what do you think?

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
rs--: Not that I'm aware of

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
rs--: Not all containers have prereqs, but yes, I like the idea of combining the build and push into the same stage!

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
CyanDevs: Actually that looks great!

solutions/pytorch_tests/Dockerfile
Francis-Liu: @CyanDevs I see, moving to line 4 is correct. Or keep line 2, and adding `ARG PYTORCH_VERSION` to line 4.

.jenkins/infrastructure/docker/build_mystikos_docker_images.Jenkinsfile
rs--: Yeah, that's a good idea 👍

asvrada (approved):
Dotnet related dockerfile LGTM

solutions/pytorch_tests/Dockerfile
```diff
@@ -1,10 +1,11 @@
...
-ARG PYTORCH_VERSION=v1.10.0
 
 FROM ${BASE_IMAGE} as test
+ARG TAG=v1.10.0
+
 RUN pip3 install pytest expecttest hypothesis && \
     pip3 install torch==1.10.0+cpu torchvision==0.11.1+cpu torchaudio==0.10.0+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html
 
 WORKDIR /workspace
-RUN git clone -b ${PYTORCH_VERSION} https://github.com/pytorch/pytorch.git --single-branch --depth 1
```
CyanDevs, line 9: Replacing PYTORCH_VERSION here will break the ability to specify it in https://github.com/deislabs/mystikos/blob/57fca89e5e60a88a452b2bb78fba55ae1c4a73f0/solutions/pytorch_tests/Makefile#L18

You can get around this and keep your tag idea by doing something like this:
```
ARG TAG=v1.10.0
ARG PYTORCH_VERSION=${TAG}

RUN git clone -b ${PYTORCH_VERSION} ...
```
So this way you can define TAG, or PYTORCH_VERSION, or none of them and still have a successful build.

tests/aspnetcore5/Dockerfile.aspnet5
```diff
@@ -1,8 +1,9 @@
 # Base image recommendation from dotnet git repo - https://github.com/dotnet/runtime/blob/main/docs/workflow/building/coreclr/linux-instructions.md#docker-images
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-a50a721-20191120200116 as builder
+ARG TAG
 
 WORKDIR /app
-RUN git clone --branch v5.0.11 --recursive https://github.com/dotnet/aspnetcore
+RUN git clone --branch ${TAG} --recursive https://github.com/dotnet/aspnetcore
```
CyanDevs, line 6: Careful about replacing the hardcoded branch names on all these Dockerfiles without a default value for `TAG`.

In this case `TAG` won't be specified when the build is started from the line below as no value for TAG is passed in. https://github.com/deislabs/mystikos/blob/57fca89e5e60a88a452b2bb78fba55ae1c4a73f0/tests/aspnetcore5/Makefile#L31 

Can be fixed by defining the default `ARG TAG=v5.0.11`

Same goes for the rest of the Dockerfiles changed in this manner

CyanDevs (changes requested):

.jenkins/infrastructure/docker/build_mystikos_base_docker.Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+        }
+        stage('Push base containers to internal repository') {
+            when {
+                expression { return params.PUBLISH_CONTAINERS }
+            }
+            steps {
+                script {
+                    tag_bionic = "mystikos-bionic:${params.MYST_BASE_DOCKER_TAG}"
+                    tag_focal = "mystikos-focal:${params.MYST_BASE_DOCKER_TAG}"
+                    docker.withRegistry(params.INTERNAL_REPO, env.INTERNAL_REPO_CREDS) {
```
jxyang, line 53: Can we push the base container to a public container registry? We are adding a tool to convert user container applications into container images with Mystikos bits injected. See #1161 for details.

CyanDevs: I think this would be doable, but would you be amenable having this done as a separate PR down the road so we can migrate off of Docker Hub asap?

.jenkins/infrastructure/docker/build_mystikos_base_docker.Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+        label 'ACC-1804-DC4'
+    }
+    options {
+        timeout(time: 240, unit: 'MINUTES')
+    }
+    parameters {
+        string(name: "MYST_VERSION", description: "Mystikos release version (Example: 0.5.0). See https://github.com/deislabs/mystikos/releases for release versions")
+        string(name: "REPOSITORY_NAME", defaultValue: "deislabs/mystikos", description: "GitHub repository to checkout")
+        string(name: "BRANCH_NAME", defaultValue: "master", description: "The branch used to checkout the repository")
+        string(name: "MYST_BASE_CONTAINER_TAG", defaultValue: "latest", description: "The tag for the new Mystikos base Docker container.")
```
CyanDevs, line 12: Nit: Do we want to have the tag default to latest, when there is already `TAG_LATEST` ?

CyanDevs (approved):
Thanks for your work on this!

.jenkins/infrastructure/docker/build_mystikos_base_docker.Jenkinsfile
rs--: Good question. I guess if we assume that the `latest` tag is of some importance, we will want to have a different default tag. On the other hand, if `latest` just signifies the latest build (no guarantees on stability/correctness of the container), then we could get rid of the `TAG_LATEST` parameter. Thoughts?

.jenkins/infrastructure/docker/build_mystikos_base_docker.Jenkinsfile
rs--: The Azure container repository that the containers are being published to will be public. The only caveat is that the repository will not be browsable, ie. we'd have to list containers in a doc for the public.

.jenkins/docker/base/build.sh
```diff
@@ -40,6 +40,8 @@ while getopts "hm:o:u::t::" option; do
     esac
 done
 
+MYST_TARBALL="Ubuntu-${UBUNTU_VERSION//.}_mystikos-${MYSTIKOS_RELEASE_VERSION}-x86_64.tar.gz"
```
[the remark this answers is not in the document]
rs--: I think this is the earliest we could define the variable because the other variables it depends on could from lines 27-41.

.jenkins/infrastructure/docker/build_mystikos_base_docker.Jenkinsfile
CyanDevs: Sorry, let me clarify that this is just an inefficiency where the container is built with the `latest` tag (if left with default value) regardless of whether `TAG_LATEST` is true or false.

When pushing to a repository it should be assumed that the container has been reasonably tested and stable, and `latest` just refers to the most recent build.

PR #1188, opened 2022-02-05
Tool for mounting docker images

mikbras:
With the help of @anakrish, this PR mounts a docker image onto a directory as follows.

```
$ mount-docker-image <docker-image-id> <mount-point>
```

This makes it possible to create an ``ext2`` image follows.

```
$ docker build --quiet --iidfile iid -f Dockerfile .
$ mount-docker-image iid mntdir | xargs sudo
$ sudo myst mkext2 --force mntdir rootfs
$ sudo umount mntdir
```

Future considerations.
- Allow ``myst mkext2`` to file-system images from docker image ids.
- Integrate ``mount-docker-image`` with ``appbuilder`` to reduce space requirements for creating an image.


tools/mount-docker-image/main.c
```diff
@@ -0,0 +1,305 @@
...
+        _err("json_parser_init() failed");
+    }
+
+    if ((r = json_parser_parse(&parser)) != JSON_OK)
+        _err("json_parser_parse() failed");
+
+    if (parser.depth != 0)
+        _err("json parsing failed");
+
+    if (!results->lowerdir)
```
anakrish, line 191: For base-images, there does not seem to be a lower-dir. E.g `docker inspect alpine`
In that case, not sure if we can omit the lowerdir option in the mount command or need to specify an empty folder.

anakrish (approved):
LGTM.

For internal validation purposes, is there some way to ensure that mount-docker-image and docker export are equivalent in terms of the resutant appdir/ext2 fs?

tools/mount-docker-image/main.c
mikbras: Good idea about validation of mont-docker-image and export! I will put something together.

Also, I was not aware of the lowerdir problem. I will try to fix it.

mikbras: I compared exported and mounted docker images. It turns out that exported images contains extra files since they were exported from live images or containers (whereas mounted images are from static images). These include:

```
/dev/console
/dev/pts
/dev/shm
.dockerenv
localhost.localdomain
/etc/mtab
```

But I think these should not be in Mystikos images anyway, so this is probably good.

mikbras: A general test that compares will not be feasible since it would included many exclusions and it would have to ignore dead soft links (which there are many of in docker images).

PR #1189, opened 2022-02-05
Disable test_limited_api and set UnhandledSyscallEnosys option

vtikoo:
test_limited_api depends on SYS_listxattr, which we don't support yet


radhikaj:
Thanks very much @vtikoo

radhikaj (approved):

PR #1191, opened 2022-02-05
Paulcallen/process kernel time

paulcallen:
Have kernel and user time per-process rather than system-wide

### DIFF
--- a/kernel/times.c
+++ b/kernel/times.c
@@ -8,6 +8,7 @@
 #include <myst/clock.h>
 #include <myst/eraise.h>
 #include <myst/kernel.h>
+#include <myst/mmanutils.h>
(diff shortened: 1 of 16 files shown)


kernel/times.c
```diff
@@ -30,8 +31,10 @@ static syscall_time_t _syscall_times[MYST_MAX_SYSCALLS];
 
 long myst_lapsed_nsecs(const struct timespec* t0, const struct timespec* t1)
 {
-    return (t1->tv_sec - t0->tv_sec) * NANO_IN_SECOND +
```
mikbras, line 33: Are we sure that these two computations are equivalent?

mikbras (approved):

PR #1195, opened 2022-02-07
Handle the case if the _handle_one_signal early returns

mingweishih:

### DIFF
--- a/kernel/signal.c
+++ b/kernel/signal.c
@@ -899,7 +899,14 @@ done:
 
 void myst_handle_host_signal(siginfo_t* siginfo, mcontext_t* mcontext)
 {
+    assert(siginfo != NULL);
+    assert(mcontext != NULL);
+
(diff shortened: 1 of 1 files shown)


kernel/signal.c
```diff
@@ -899,8 +899,15 @@ long myst_signal_clone(
...
 void myst_handle_host_signal(siginfo_t* siginfo, mcontext_t* mcontext)
 {
+    assert(siginfo != NULL);
+    assert(mcontext != NULL);
+
     _handle_one_signal(siginfo->si_signo, siginfo, mcontext);
 
+    /* resume the execution based on mcontext for the unlikely case
+     * where the _handle_one_signal returns (e.g., early return) */
+    myst_sigreturn(mcontext);
```
anakrish, line 909: Is it OK to do this or are we masking a genuine issue by choosing to resume execution?

anakrish (approved):

PR #1206, opened 2022-02-09
Fix minimum required lldb version

anakrish:
The Python interface has changed since version 8 and that is the
minimum required version. oelldb is patched to fix the version.

Don't silence error output from lldb in the test to help with debugging test failure.

oelldb's minimum required version will be updated separately in the OE repo.

### DIFF
--- a/third_party/openenclave/Makefile
+++ b/third_party/openenclave/Makefile
@@ -96,6 +96,8 @@ endif
 	mkdir -p $(TOP)/build/bin
 	rm -f $(TOP)/build/bin/myst-gdb
 	ln -s $(TOP)/build/openenclave/bin/oegdb $(TOP)/build/bin/myst-gdb
+	# Patch minimum required lldb version from 7 to 8.
+	sed -i 's/-7/-8/g' $(TOP)/build/openenclave/bin/oelldb
 	rm -f $(TOP)/build/bin/myst-lldb
 	ln -s $(TOP)/build/openenclave/bin/oelldb $(TOP)/build/bin/myst-lldb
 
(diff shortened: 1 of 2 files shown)


radhikaj (approved):

third_party/openenclave/Makefile
```diff
@@ -96,6 +96,8 @@ endif
 	mkdir -p $(TOP)/build/bin
 	rm -f $(TOP)/build/bin/myst-gdb
 	ln -s $(TOP)/build/openenclave/bin/oegdb $(TOP)/build/bin/myst-gdb
+	# Patch minimum required lldb version from 7 to 8.
+	sed -i 's/-7/-8/g' $(TOP)/build/openenclave/bin/oelldb
```
paulcallen, line 100: myst_lldb needs to be updated to make sure it is using is using lldb of *greater than*  7. The error message needs updating too.

anakrish: The `sed` command updates the version number in the error message too
> echo "oelldb requires lldb-8 or above."

paulcallen (approved):

third_party/openenclave/Makefile
```diff
@@ -96,6 +96,8 @@ endif
 	mkdir -p $(TOP)/build/bin
 	rm -f $(TOP)/build/bin/myst-gdb
 	ln -s $(TOP)/build/openenclave/bin/oegdb $(TOP)/build/bin/myst-gdb
+	# Patch minimum required lldb version from 7 to 8.
```
vtikoo, line 99: Should we update this in oelldb's source in OE's mystikos branch instead?

anakrish: That can be done separately as noted in the comment. I'm trying to avoid unnecessarily modifying the OE branch.

PR #1207, opened 2022-02-09
when tests fail some abort the pipeline

paulcallen:
sometimes we can just prepend $(RUNTEST) before the actual test which is
what was done for libcxx. For sockperf it was specificly the docker
command that failed. For sockperf I ignored the return from docker as
the test is going to fail if the docker run command did not work anyway.


radhikaj (commented):
Thank you @paulcallen

radhikaj (approved):

PR #1210, opened 2022-02-10
Serialize callback invocations from dl_iterate_phdr

vtikoo:
### Summary
During C++ exception stack unwinding, `int dl_iterate_phdr(
          int (*callback) (struct dl_phdr_info *info,
                           size_t size, void *data),
          void *data)` is called to find frame information. `dl_iterate_phdr` iterates over the list of loaded modules, and calls the callback on each one of them until the callback returns a non-zero code.
Calling of the callback is not protected by any locks. If the callback is not thread safe, and multiple threads are throwing exceptions(as in dotnet's case), then it might lead to undefined behavior.

The callback in for C++ stack unwinding `_Unwind_IteratePhdrCallback` does have a static data member [adds](https://github.com/gcc-mirror/gcc/blob/master/libgcc/unwind-dw2-fde-dip.c#L222) .

### DIFF
--- a/third_party/musl/crt/patch.diff
+++ b/third_party/musl/crt/patch.diff
@@ -261,7 +261,7 @@ index 20d50f2c..17cd6ac3 100644
 +{
 +}
 diff --git a/ldso/dynlink.c b/ldso/dynlink.c
-index afec985a..acf447c4 100644
+index afec985a..9f9b618d 100644
 --- a/ldso/dynlink.c
 +++ b/ldso/dynlink.c
 @@ -3,6 +3,7 @@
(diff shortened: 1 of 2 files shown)


mikbras (approved):
Looks good to me. Thank you for doing this Vikas.

third_party/musl/crt/patch.diff
```diff
@@ -497,7 +497,30 @@ index afec985a..acf447c4 100644
...
++ * Passed callbacks may have static members. */
++static pthread_mutex_t dl_iterate_cb_lock;
+ int dl_iterate_phdr(int(*callback)(struct dl_phdr_info *info, size_t size, void *data), void *data)
+ {
+ 	struct dso *current;
+@@ -2287,9 +2344,11 @@ int dl_iterate_phdr(int(*callback)(struct dl_phdr_info *info, size_t size, void
+ 		info.dlpi_tls_modid = current->tls_id;
+ 		info.dlpi_tls_data  = current->tls.image;
+ 
++		pthread_mutex_lock(&dl_iterate_cb_lock);
```
anakrish, line 514: It is possible to call dl_iterate_phdr from within a callback. Using a mutex would lead to deadlocks.

PR #1214, opened 2022-02-11
Add retry and runtest command to dotnet tests

asvrada:


vtikoo (approved):
LGTM. Not necessary as part of this PR, but we should also think of setting Makefile level timeouts to protect against hung tests.

vtikoo (changes requested):
Just realized wrapping with RUNTEST introduces a default timeout of 120 secs. @asvrada will that be enough for all the tests changed here?

asvrada:
> Just realized wrapping with RUNTEST introduces a default timeout of 120 secs. @asvrada will that be enough for all the tests changed here?

Will double check

asvrada:
aspnetcore5 and aspnetcore6 are extremely flaky on my local machine... Random tests would hang and timeout, so I am not adding `$(RUNTEST)` command in this PR

vtikoo (approved):

PR #1219, opened 2022-02-11
new docker-build-ext2 script

mikbras:
Simple internal script to build a docker image and then uses ``mount-docker-image``, writing the resulting root file system to an EXT2 file. For example:

```
# docker-build-ext2 Dockerfile ext2rootfs
```


mikbras:
Update PR to address two comments by Paul:
- add `trap cleanup EXIT` handler
- forward an extraneous arguments to docker (using `$*`)

The second change gives the command the following syntax 
```
$ docker-build-ext2 ext2_image <docker_build_arguments...>
```

Where the ``<docker_build_arguments>`` are just passed on to the ``docker build`` command.

paulcallen (approved):

scripts/docker-build-ext2
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+
+##==============================================================================
+##
+## Build a docker image and write the output to an EXT2 image
+##
+##==============================================================================
```
jxyang, line 7: Can you make this part of appbuilder?

jxyang (changes requested):

scripts/docker-build-ext2
mikbras: Yes but I would need your help with that since it requires some modification of the appbuilder options and interface. Since this script is for internal use only (and is not installed) can we merge this as a first step?

radhikaj:

PR #1222, opened 2022-02-15
[EXPERIMENTAL] Handle deferred cancel type for non-self threads

vtikoo:
glibc [`tst_cancel8`](https://elixir.bootlin.com/glibc/glibc-2.26/source/nptl/tst-cancel8.c) test fails currently because we don't defer cancellation request.

`pthread_cancel` currently supports deferring signals only if the thread called `pthread_cancel` on itself.

PTHREAD_CANCEL_DEFERRED
A cancellation request is deferred until the thread next calls a function that is a cancellation point (see [pthreads](https://linux.die.net/man/7/pthreads)(7)). This is the default cancelability type in all new threads, including the initial thread.
From - https://linux.die.net/man/3/pthread_setcanceltype

### DIFF
--- a/third_party/musl/crt/patch.diff
+++ b/third_party/musl/crt/patch.diff
@@ -1287,7 +1287,7 @@ index 1eddcbd6..d3ec0363 100644
 +	return DEFAULT_STACK_SIZE;
 +}
 diff --git a/src/thread/pthread_cancel.c b/src/thread/pthread_cancel.c
-index 2f9d5e97..e4d3d000 100644
+index 2f9d5e97..53603be6 100644
 --- a/src/thread/pthread_cancel.c
 +++ b/src/thread/pthread_cancel.c
 @@ -14,9 +14,7 @@ long __cancel()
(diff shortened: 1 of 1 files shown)


vtikoo:
[`tests/signal.c:test_pthread_cancel` ](https://github.com/deislabs/mystikos/blob/main/tests/signal/signal.c#L110-L123)is failing for this PR.

The test tests that cancellation handlers are called and that child thread was cancelled. 

The child pthread does not set either its cancellation state or type. Based on this [man page](https://linux.die.net/man/3/pthread_cancel), the default should be cancel enabled and deferred delivery.

However both glibc and current musl implementation deliver the signal asynchronously. 

cc: @jxyang  @paulcallen @mikbras

Simplified version of the failing test:
```C
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

#define _GNU_SOURCE
#include <assert.h>
#include <errno.h>
#include <pthread.h>
#include <semaphore.h>
#include <signal.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/syscall.h>
#include <unistd.h>

static sem_t sem2;

static void cleanup1(void* arg)
{
    *(int*)arg = 1;
}


static void* start_single(void* arg)
{
    pthread_cleanup_push(cleanup1, arg);
    sem_post(&sem2);
    sleep(3);
    // pthread_testcancel();
    pthread_cleanup_pop(0);
    return 0;
}

int main(int argc, const char* argv[])
{
    pthread_t td;
    int r;
    void* res;
    _Atomic(int) foo;

    assert(sem_init(&sem2, 0, 0) == 0 && "creating semaphore");

    foo = 0;
    assert(
        pthread_create(&td, 0, start_single, &foo) == 0 &&
        "failed to create thread");

    while (sem_wait(&sem2))
        ;
    int ret = pthread_cancel(td);
    printf("cancel ret=%d\n", ret);
    assert(ret == 0 && "failed to cancel");
    assert(pthread_join(td, &res) == 0 && "failed to join");
    printf("res=%p\n", res);
    assert(res == PTHREAD_CANCELED && "canceled thread exit status");
    assert(foo == 1 && "cleanup handler failed to run");

    return 0;
}
```

PR #1246, opened 2022-02-28
Disable tst-cancel8 glibc testcase

vtikoo:
Test relies on deferred cancellation of a thread. Mystikos currently delivers pthread_cancel asynchronously to other threads, irrespective of their cancel type.

### DIFF
--- a/tests/glibc/tests.failed
+++ b/tests/glibc/tests.failed
@@ -2,3 +2,4 @@
 /glibc/build/inet/test-ifaddrs: creates a socket with AF_NETLINK
 /glibc/build/io/tst-ftw-lnk.d/link2-tgt ldso: /glibc/build/io/tst-ftw-lnk.d/link2-tgt: Not a valid dynamic program 
 /glibc/build/nptl/test-mutexattr-printers Makefile:57: recipe for target 'one' failed ****Runs without error in GDB****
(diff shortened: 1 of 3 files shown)


vtikoo:
More details in #1222

asvrada (approved):

tests/glibc/tests.failed
```diff
@@ -2,3 +2,4 @@
 /glibc/build/inet/test-ifaddrs: creates a socket with AF_NETLINK
 /glibc/build/io/tst-ftw-lnk.d/link2-tgt ldso: /glibc/build/io/tst-ftw-lnk.d/link2-tgt: Not a valid dynamic program 
 /glibc/build/nptl/test-mutexattr-printers Makefile:57: recipe for target 'one' failed ****Runs without error in GDB****
+/glibc/build/nptl/tst-cancel8: Requires support of deferred cancel type for pthread_cancel
```
salsal97, line 5: Hey, since this is a test that flakily passes it would be better to add it to tests.remove since running the automated script to update this suite might cause one to unkowingly put it in our passed list again..

salsal97 (approved):
Thanks for adding this

tests/glibc/tests.failed
vtikoo: Thanks for pointing that out! Didn't realize the difference between failed and remove lists till now.

PR #1247, opened 2022-02-28
Patch glibc test runner

asvrada:
Store pid returned from fork in a local variable instead of global one. This would reduce the flakiness of glibc tests

### DIFF
--- a/tests/glibc/Makefile
+++ b/tests/glibc/Makefile
@@ -41,7 +41,9 @@ build: glibc
 PWD=$(shell pwd)
 
 $(APPDIR):
-	$(APPBUILDER) -i mystikos/glibc-test:0.1
+	# $(APPBUILDER) -i mystikos/glibc-test:0.1
+	# todo: replace with pre-built image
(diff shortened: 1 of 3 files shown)


tests/glibc/Makefile
```diff
@@ -41,7 +41,9 @@ build: glibc
 PWD=$(shell pwd)
 
 $(APPDIR):
-	$(APPBUILDER) -i mystikos/glibc-test:0.1
+	# $(APPBUILDER) -i mystikos/glibc-test:0.1
+	# todo: replace with pre-built image
+	$(APPBUILDER) -i hullcritical/glibc-test:0.1
```
paulcallen, line 46: why did you change from `mystikos/glibc-test:0.1` to `hullcritical/glibc-test:0.1`?

asvrada: This image (uploaded to my personal account, will talk to @rs-- to have it replaced with a Mystikos one), is created from the modified Dockerfile to reflect the change to glibc test runner

paulcallen (approved):

PR #1248, opened 2022-03-01
add runtest to sdk suites

salsal97:


salsal97:
Successful run here: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Mystikos%2FStandalone-Pipelines%2FAzure-SDK-Tests-Pipeline/detail/Azure-SDK-Tests-Pipeline/979/pipeline/56

solutions/dotnet_azure_sdk/Makefile
```diff
@@ -10,6 +10,12 @@ DIRS += KeyVault_Certificates
 REDEFINE_TESTS=1
 include $(TOP)/rules.mak
 
+ifdef RUN_AZURE_TESTS
```
jxyang, line 13: Under what circumstances we want `RUNTEST` to be empty?

CyanDevs (approved):
LGTM

tests/azure-sdk-for-cpp/Makefile
```diff
@@ -48,4 +55,4 @@ clean:
 	rm -rf $(APPDIR) $(ROOTFS)
 
 get-executables:
-	find $(RUNTEST) $(MYST_EXEC) $(OPTS) $(ROOTFS) /azure-sdk-for-cpp/build/sdk -type f -executable -print | grep test
+	find $(MYST_EXEC) $(OPTS) $(ROOTFS) /azure-sdk-for-cpp/build/sdk -type f -executable -print | grep test
```
jxyang, line 58: Seems this target is not used anymore? What's the purpose of it?

solutions/dotnet_azure_sdk/Makefile
salsal97: When we are running any of the tests independently, in that case runtest would be empty

tests/azure-sdk-for-cpp/Makefile
CyanDevs: I wondered this myself. There's no mention of it but I guess it is kept around as a convenience for finding test executables

salsal97: This target is used when we extract the binaries from the docker container, mostly it is kept for convenience

jxyang: We can find the tests from appdir without invoking mystikos. In addition, I suspect the command should be:
```
$(MYST_EXEC) $(OPTS) $(ROOTFS) find /azure-sdk-for-cpp/build/sdk -type f -executable -print |  grep test
```

CyanDevs: Good catch Xuejun

PR #1257, opened 2022-03-05
Patch musl pathconf

asvrada:
For case "PC_NAME_MAX", check if filepath or fd is valid

This patch will enable Cpython 3.10 `test_os.TestInvalidFD.test_fpathconf`

### DIFF
--- a/third_party/musl/crt/patch.diff
+++ b/third_party/musl/crt/patch.diff
@@ -551,6 +551,96 @@ index 02cb1aa2..07e76f19 100644
  	} else if ((name&~4U)!=1 && name-_CS_POSIX_V6_ILP32_OFF32_CFLAGS>33U) {
  		errno = EINVAL;
  		return 0;
+diff --git a/src/conf/fpathconf.c b/src/conf/fpathconf.c
+index e6aca5cf..d7392ced 100644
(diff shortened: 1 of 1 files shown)


third_party/musl/crt/patch.diff
```diff
@@ -551,6 +551,96 @@ index 02cb1aa2..07e76f19 100644
...
+ long fpathconf(int fd, int name)
+ {
+@@ -31,5 +32,20 @@ long fpathconf(int fd, int name)
+ 		errno = EINVAL;
+ 		return -1;
+ 	}
++
++	// The maximum length of a filename in the directory path or
++	// fd that the process is allowed to create
++	if (name == _PC_NAME_MAX) {
```
vtikoo, line 573: Is it easy in musl to move this logic to a function which both `pathconf` and `fpathconf` can use?

vtikoo (approved):
Looks good! Added a comment about possible code-reuse opportunity.

third_party/musl/crt/patch.diff
asvrada: I couldn't think of a way to reuse this, since one is pathname and the other is fd?

PR #1262, opened 2022-03-08
Populate /etc/hosts and /etc/hostname: Fix intermittent EAI_AGAIN cpy…

anakrish:
…thon failures.

MUSL's getaddrinfo has the following logic for a given name:
1) if name exists in /etc/hosts:
     \# Super fast. local host typically is resolved thus.
     return ip address from /etc/hosts for the name

2) for each search-domain in /etc/resolv.conf:
     \# Requires network connection. Slow. Timing-sensitive.
     if dns-server successfully resolves "name.search-domain":
        return ip address returned by dns-server for "name.search-domain"

   \# Fast for a local server (127.0.0.53) since value is returned  from cache.
   \# Could be slow for a remote dns server
3) ask dns-server to resolve "localhost" and return ip address

cpython tests require resolving "localhost".

Prior to this PR, /etc/hosts is empty and resolving "localhost" required (2) or (3) above.
If search-domains were specified in /etc/resolv.conf (as is done in typical Azure VM/CI node),
then this resulted in network access and occassional timeout failures.

This PR propagates /etc/hosts and /etc/hostname from the host system, just like /etc/resolv.conf
is done currently. This ensures that, on a typical host, resolving localhost is done from
the /etc/hosts.

### DIFF
--- a/kernel/enter.c
+++ b/kernel/enter.c
@@ -105,22 +105,21 @@ done:
     return ret;
 }
 
-static int _copy_host_etc_files()
+static int _copy_host_etc_file(const char* path)
 {
     int ret = 0;
     int fd = -1;
-    const char* resolv_file = "/etc/resolv.conf";
     void* buf = NULL;
     size_t buf_size;
     struct stat statbuf;
 
-    ECHECK(myst_load_host_file(resolv_file, &buf, &buf_size));
+    ECHECK(myst_load_host_file(path, &buf, &buf_size));
 
-    if (stat(resolv_file, &statbuf) == 0)
+    if (stat(path, &statbuf) == 0)
     {
-        if ((myst_syscall_unlink(resolv_file)) < 0)
+        if ((myst_syscall_unlink(path)) < 0)
         {
-            myst_eprintf("kernel: failed to unlink file %s\n", resolv_file);
+            myst_eprintf("kernel: failed to unlink file %s\n", path);
             ERAISE(-EINVAL);
         }
     }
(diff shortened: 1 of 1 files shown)


kernel/enter.c
```diff
@@ -170,6 +169,16 @@ static int _copy_host_etc_files()
     return ret;
 }
 
+static int _copy_host_etc_files()
+{
+    int ret = 0;
+    ECHECK(_copy_host_etc_file("/etc/resolv.conf"));
```
anakrish, line 175: @bodzhang @paulcallen @mikbras @jxyang 

The default value for `timeout` option is 5. The value is capped to 30.
See https://man7.org/linux/man-pages/man5/resolver.5.html

Given that we've seen timeout (EAI_AGAIN) errors in CI, I'm led to believe that dns resolution in Mystikos is slower than normal - could be due to SGX overhead, CI network issues etc.

One way to ensure minimize timeout errors is to append the line `options timeout:30` to the resolv.conf file.
What do you think? 

Many developers may not be aware what to do when dns resolution timeouts happen. So, automatically doing it for them could lead to better developer experience.

kernel/enter.c
```diff
@@ -170,6 +169,16 @@ static int _copy_host_etc_files()
...
+    ECHECK(_copy_host_etc_file("/etc/hosts"));
+    ECHECK(_copy_host_etc_file("/etc/hostname"));
```
paulcallen, line 177: hostname is part of the our config.json:



Hostname | string | The default hostname exposed to application
-- | -- | --




Not sure how this may work in conjunction with that.

anakrish: Fixed.

paulcallen (approved):

kernel/enter.c
anakrish: @paulcallen pointed out that the default was `5 seconds` which is already sufficiently large. Therefore, there is no need to increase the timeout to 30 seconds.

PR #1264, opened 2022-03-09
Fix double unlock bug in waitpid (myst_wait)

anakrish:
Two calls to myst_spin_unlock(&myst_process_list_lock) were being issued
one after the other. Sometimes, the second unlock would unlock the lock
held by a child thread that is exiting resulting in threading issues.

Remove the double unlock to fix the bug.

~Additionally, add an assertion to ensure that a spinlock that is
being unlocked is in fact locked.~

spinlock assertion to be added in another PR.

### DIFF
--- a/include/myst/spinlock.h
+++ b/include/myst/spinlock.h
@@ -5,6 +5,7 @@
 #define _MYST_SPINLOCK_H
 
 #include <errno.h>
+#include <myst/assume.h>
 #include <myst/defs.h>
 #include <myst/types.h>
 
--- a/kernel/thread.c
+++ b/kernel/thread.c
@@ -718,9 +718,6 @@ long myst_wait(
             ERAISE(-ECHILD);
         }
 
-        myst_spin_unlock(&myst_process_list_lock);
-        locked = false;
-
         if ((options & WNOHANG))
         {
             ret = 0;


kernel/thread.c
```diff
@@ -718,9 +718,6 @@ long myst_wait(
             ERAISE(-ECHILD);
         }
 
-        myst_spin_unlock(&myst_process_list_lock);
```
anakrish, line 721: There is an unlock in line 763,

include/myst/spinlock.h
```diff
@@ -78,6 +79,8 @@ MYST_INLINE void myst_spin_lock(myst_spinlock_t* spinlock)
 
 MYST_INLINE void myst_spin_unlock(myst_spinlock_t* spinlock)
 {
+    // Assert that the spinlock is locked.
+    myst_assume(*spinlock != 0);
```
anakrish, line 83: Instead of storing `1` in the lock, we could store the 4 bytes of the kernel thread pointer (GS[0]). That would allow us to assert that only the thread that owns the spinlock actually unlocks it.' 

Cases that would go undetected would be if two threads have the same 4-byte identifier. That is still better than not detecting any ownership issues.

kernel/thread.c
paulcallen: wow! Great find.

radhikaj:
@anakrish  - Can we have just the double unlock fix in one PR and the assert in a different one? If you would like, I can create such a PR. The assert is causing pipeline issues, which is likely because of other unmatched locks and unlocks

radhikaj:
@anakrish  - Can we have just the double unlock fix in one PR and the assert in a different one? If you would like, I can create such a PR. The assert is causing pipeline issues, which is likely because of other unmatched locks and unlocks

anakrish:
> @anakrish - Can we have just the double unlock fix in one PR and the assert in a different one? If you would like, I can create such a PR. The assert is causing pipeline issues, which is likely because of other unmatched locks and unlocks

Yes, the assertion is catching other unmatched lock/unlocks. I will back that out for now while I figure out the assertions.

paulcallen (approved):

PR #1272, opened 2022-03-15
Switch to OE master

mingweishih:
With the upstreaming Mystikos-dependent OE feature completed, This PR switches from Mystikos OE branch to OE master.
This changes on the Mystikos-side is as follows:
- Making sure to use the OE-default FS register before returning to OE (i.e., making ocall).
- Use the new `oe_register_load_extra_enclave_data_hook` API for regions support
- Replace the duplicated `oe_notify_debugger_library_load` by `oe_debug_module_loaded_hook` in the OE debugger contract.

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -7634,14 +7634,17 @@ done:
 
     /* ---------- running target thread descriptor ---------- */
 
+    /* Process signals pending for this thread, if there is any,
+     * before switching back to the C-runtime thread descriptor to
+     * ensure running the signal processing code with the kernel
+     * thread descriptor. */
+    myst_signal_process(thread);
+
     /* the C-runtime must execute on its own thread descriptor */
     if (crt_td)
         myst_set_fsbase(crt_td);
 
     myst_times_leave_kernel(n);
-
-    // Process signals pending for this thread, if there is any.
-    myst_signal_process(thread);
 
     return syscall_ret;
 }
@@ -7651,6 +7654,16 @@ long myst_syscall(long n, long params[6])
     long ret;
     uint64_t rsp;
     myst_kstack_t* kstack;
+    void* saved_fs;
+    void* base_fs;
+
+    /* At this point, we cannot be sure about what FS is using (determined
+     * by _syscall). To ensure using the kernel FS (same as OE FS) when calling
+     * functions that could come across the OE layer (e.g., making an OCALL),
+     * we temporarily switch to the kernel FS if it has user value before
+     * calling the function and restore when the function returns. */
+    asm("mov %%fs:0, %0" : "=r"(saved_fs));
+    asm("mov %%gs:0, %0" : "=r"(base_fs));
 
     // Call myst_syscall_clock_gettime() upfront to avoid triggering the
     // overhead of myst_times_enter_kernel() and myst_times_leave_kernel(),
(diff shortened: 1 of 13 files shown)


jxyang:
Started a manual nightly pipeline for the PR with broad impact here:

https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Mystikos%2FNightly-Pipeline-Manual/detail/Nightly-Pipeline-Manual/48/pipeline

tools/myst/enc/enc.c
```diff
@@ -420,6 +449,10 @@ static uint64_t _vectored_handler(oe_exception_record_t* er)
         }
     }
 
+    /* Restore FS if needed */
+    if (saved_fs)
+        asm volatile("wrfsbase %0" ::"r"(saved_fs));
+
     return _forward_exception_as_signal_to_kernel(er);
```
anakrish, line 456: Within a signal handler the FS value is that of the application's. C++ allows certain thread-local variables to be accessed from a signal handler. This means that the FS value ought to be applications within the handler - neither OE's nor mystikos kernel's.

third_party/openenclave/Makefile
```diff
@@ -5,7 +5,7 @@ TOP=$(abspath $(CURDIR)/../..)
 BUILDDIR=${TOP}/build
 
 URL=https://github.com/openenclave/openenclave
-BRANCH=mystikos.v6
+BRANCH=master
```
anakrish, line 8: It is important to lock this down to a specific commit. Otherwise, it is not easily possible to checkout an older version of Mystikos and build it - we won't know what commit of OE a specific commit of Mystikos worked with.

anakrish (commented):
Looks good for most part.
Can the debugger changes be checked in as a separate PR? They look good to go.

tools/myst/enc/enc.c
mingweishih: The kernel code will switch the FS to user value before calling the user handler: https://github.com/deislabs/mystikos/blob/178bea1b744799cddb365e9f5a8cb6c9e625bb32/kernel/signal.c#L559

kernel/syscall.c
```diff
@@ -7634,15 +7634,15 @@ static long _syscall(void* args_)
 
     /* ---------- running target thread descriptor ---------- */
 
+    // Process signals pending for this thread, if there is any.
```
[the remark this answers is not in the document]
mingweishih: Updated comment.

kernel/syscall.c
```diff
@@ -7651,15 +7651,29 @@ long myst_syscall(long n, long params[6])
     long ret;
     uint64_t rsp;
     myst_kstack_t* kstack;
+    void* saved_fs;
+    void* base_fs;
+
+    // Save the current FS
+    asm("mov %%fs:0, %0" : "=r"(saved_fs));
```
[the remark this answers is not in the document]
mingweishih: Updated comment

kernel/syscall.c
```diff
@@ -7672,9 +7686,15 @@ long myst_syscall(long n, long params[6])
...
     }
 
+    if (saved_fs != base_fs)
+        myst_set_fsbase(base_fs);
+
     if (!(kstack = myst_get_kstack()))
         myst_panic("no more kernel stacks");
 
+    if (saved_fs != base_fs)
+        myst_set_fsbase(saved_fs);
```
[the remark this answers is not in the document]
mingweishih: updated comment. Currently only `myst_get_kstack` requires save and restore FS.

tools/myst/enc/enc.c
anakrish: One thing I noticed was that if an application changes FS from appFS to newFS and then a signal happens, then in Linux the value of FS within the signal handler is newFS. In mystikos, the value within the handler is appFS. 
I am not sure if this will cause problems; but it would be good to confirm the Linux behavior and have Mystikos match it so.

anakrish (approved):
LGTM. The descriptive comments make it easier to read.

PR #1273, opened 2022-03-15
Maintain libc.threads_minus_1 correctly.

anakrish:
The thread created for the child of a fork is not created via pthread_create.
Such a thread also does not call pthread_exit.

Typically, libc.threads_minus_1 is maintained by pthread_create and pthread_exit.

Since these functions are not invoked for a forked child, maintain the counter
manually by doing increment just before cloning to create the thread and
doing decrement just before invoking SYS_group_exit for the thread.

### DIFF
--- a/crt/exit.c
+++ b/crt/exit.c
@@ -1,8 +1,18 @@
+#define hidden __attribute__((__visibility__("hidden")))
+
+#define _GNU_SOURCE
+
 #include <myst/syscallext.h>
 #include <stdio.h>
 #include <sys/types.h>
 #include <syscall.h>
 #include <unistd.h>
+
+#include <pthread_impl.h>
+
+/* Locking functions used by MUSL to manage libc.threads_minus_1 */
+void __tl_lock(void);
+void __tl_unlock(void);
 
 void do_crt_exit(int code);
 
@@ -51,6 +61,16 @@ void exit(int code)
                     // failed so safest option is fast exit
                     _Exit(code);
                 }
+            }
+
+            if (arg.is_child_fork)
+            {
+                // Decrement MUSL's number of threads. Typically this is done by
+                // MUSL's pthread_exit which is not called for child process's
+                // main thread.
+		__tl_lock();
+		__libc.threads_minus_1--;
+		__tl_unlock();
             }
             _Exit(code);
         }
--- a/crt/enter.c
+++ b/crt/enter.c
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#define hidden __attribute__((__visibility__("hidden")))
+
 #define _GNU_SOURCE
 #include <assert.h>
 #include <dlfcn.h>
@@ -25,6 +27,11 @@
 #include <myst/syscall.h>
 #include <myst/syscallext.h>
 #include <myst/tee.h>
+
+/* Locking functions used by MUSL to manage libc.threads_minus_1 */
+#include <pthread_impl.h>
+void __tl_lock(void);
+void __tl_unlock(void);
 
 static myst_wanted_secrets_t* _wanted_secrets;
 
@@ -87,6 +94,18 @@ long myst_syscall(long n, long params[6])
          */
         pid_t ret = myst_fork();
         return ret;
+    }
+
+    if (n == SYS_execve || n == SYS_execveat)
+    {
+        /* These system calls launch a new process with its own CRT and the
+         * current thread then belongs to the new process. Therefore, decrement
+         * the thread count of the current CRT that belongs to the parent
+         * process.
+         */
+        __tl_lock();
+        __libc.threads_minus_1--;
+        __tl_unlock();
     }
 
     return (*_syscall_callback)(n, params);
--- a/crt/fork.c
+++ b/crt/fork.c
@@ -25,6 +25,10 @@
 #include <myst/setjmp.h>
 #include <myst/syscallext.h>
 
+/* Locking functions used by MUSL to manage libc.threads_minus_1 */
+void __tl_lock(void);
+void __tl_unlock(void);
+
 static void _set_fsbase(void* p)
 {
     if (syscall(SYS_set_thread_area, p) < 0)
@@ -436,6 +440,14 @@ __attribute__((__returns_twice__)) pid_t myst_fork(void)
         args->unmap_on_exit.mmap_ptr_size = mmap_rounded_size;
         args->canary = args->child_pthread->canary;
 
+        // Increment thread count. For regular threads, this is done by MUSL's
+        // pthread_create. Since the child process's thread is created via clone
+        // instead of pthread_create, the thread count must be incremented
+        // manually here.
+	__tl_lock();
+        __libc.threads_minus_1++;
+	__tl_unlock();
+	
         if ((tmp_ret = __clone(_child_func, sp, clone_flags, args)) < 0)
         {
             munmap(child_pthread->map_base, child_pthread->map_size);


vtikoo (commented):
This is a great find! We should probably also decrement the counter when there is an execve? As the caller will have a new CRT by the end of the syscall.

crt/exit.c
```diff
@@ -52,6 +58,15 @@ void exit(int code)
                     _Exit(code);
                 }
             }
+
+            if (arg.is_child_fork)
+            {
+                // Decrement MUSL's number of threads. Typically this is done by
+                // MUSL's pthread_exit which is not called for child process's
+                // main thread.
+                __atomic_sub_fetch(
```
paulcallen, line 67: crt protects this with locks. Therefore this is not thread safe. You will need to patch musl to make the variable atomic.

anakrish:
> This is a great find! We should probably also decrement the counter when there is an execve? As the caller will have a new CRT by the end of the syscall.

Great point. We need to keep the counter in sync when the thread becomes owned by the child process. If the child process ever hands the ownership back to the parent process, then at that point the count ought to be incremented.

But I'm not sure where exactly that happens in code...

crt/exit.c
anakrish: Good point. We need to use the same lock that MUSL uses or patch musl to make this atomic.

anakrish: Fixed.

anakrish: Used locks to be consistent with MUSL. I considered using atomics, but felt it was simpler to just follow what MUSL is doing.

paulcallen (approved):

vtikoo:
> 
> We need to keep the counter in sync when the thread becomes owned by the child process. If the child process ever hands the ownership back to the parent process, then at that point the count ought to be incremented.
> 
> But I'm not sure where exactly that happens in code...

src/process/execve.c under musl seems like a good place. We could either modify musl or rewrite execve under `crt`.

jxyang:
Started a nightly pipeline for the PR with broad impact. https://oe-jenkins-dev.westeurope.cloudapp.azure.com/job/Mystikos/job/Nightly-Pipeline-Manual/34/

Please wait for the outcome.

jxyang:
Looks like the PR caused a reproducible failure on samples/go. Please investigate.

https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Mystikos%2FStandalone-Pipelines%2FSolutions-Tests-Pipeline/detail/Solutions-Tests-Pipeline/3346/pipeline

https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Mystikos%2FNightly-Pipeline-Manual/detail/Nightly-Pipeline-Manual/34/pipeline/351

anakrish:
Unfortunately, the samples/go failure doesn't easily reproduce in GDB even after running for hours. I'm trying to see what other approach we can use to understand the SEGV.

When run  in a loop outside GDB, there is eventually a SEGV:

fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x1c pc=0x13ee1a36d]

goroutine 3 [running]:
runtime.throw({0x13f5efc69?, 0x1e?})
        /usr/local/go/src/runtime/panic.go:992 +0x71 fp=0x114337688 sp=0x114337658 pc=0x13ee5cfb1
runtime.sigpanic()
        /usr/local/go/src/runtime/signal_unix.go:802 +0x3c9 fp=0x1143376d8 sp=0x114337688 pc=0x13ee796e9
runtime.chansend(0x0, 0x100a4b0b8, 0x1, 0x100a4b0b8?)
        /usr/local/go/src/runtime/chan.go:203 +0xcd fp=0x114337760 sp=0x1143376d8 pc=0x13ee1a36d
runtime: unexpected return pc for runtime.chansend1 called from 0x1007bc196
stack: frame={sp:0x114337760, fp:0x114337790} stack=[0x114337000,0x114337800)
0x0000000114337660:  0x000000013ee5cfe0 <runtime.throw.func1+0x0000000000000000>  0x000000013f5efc69
0x0000000114337670:  0x000000000000002a  0x00000001143376c8
0x0000000114337680:  0x000000013ee796e9 <runtime.sigpanic+0x00000000000003c9>  0x000000013f5efc69
0x0000000114337690:  0x000000000000001e  0x0000000000000000
0x00000001143376a0:  0x0000000000000000  0x0000000200000000
0x00000001143376b0:  0x0000000000000004  0x0000000000000000
0x00000001143376c0:  0x0000000000000003  0x0000000114337750
0x00000001143376d0:  0x000000013ee1a36d <runtime.chansend+0x00000000000000cd>  0x000000011435e058
0x00000001143376e0:  0x0000000000000000  0x0000000000000000
0x00000001143376f0:  0x0000000000000000  0x0000000000000000
0x0000000114337700:  0x0000000000000000  0x0000000000000000
0x0000000114337710:  0x0000000000000000  0x000000011435e058
0x0000000114337720:  0x0000000000000000  0x0000000000000000
0x0000000114337730:  0x0000000000000000  0x0000000000000000
0x0000000114337740:  0x0000000000000000  0x0000000000000000
0x0000000114337750:  0x0000000114337780  0x000000013ee1a27d <runtime.chansend1+0x000000000000001d>
0x0000000114337760: <0x0000000000000000  0x0000000100a4b0b8
0x0000000114337770:  0x0000000000000001  0x0000000100a4b0b8
0x0000000114337780:  0x00000001143377b0 !0x00000001007bc196
0x0000000114337790: >0x000000000000f000  0x0000000100a5ebb8
0x00000001143377a0:  0x0000000000000000  0x0000000100a5ebb8
0x00000001143377b0:  0x00000001143377d0  0x00000001007e516e
0x00000001143377c0:  0x0000000118804000  0x0000000118804000
0x00000001143377d0:  0x0000000114337850  0x00000001007d7d0a
0x00000001143377e0:  0x00000001143378b0  0x00000000000000ca
0x00000001143377f0:  0x0000000000000000  0x0000000118804000
runtime.chansend1(0xf000?, 0x100a5ebb8?)
        /usr/local/go/src/runtime/chan.go:144 +0x1d fp=0x114337790 sp=0x114337760 pc=0x13ee1a27d
created by runtime.gcenable
        /usr/local/go/src/runtime/mgc.go:177 +0x6f

goroutine 1 [chan receive, locked to thread]:
runtime.gopark(0x114362000?, 0x114336720?, 0xe5?, 0x80?, 0x0?)
        /usr/local/go/src/runtime/proc.go:361 +0xdc
runtime.chanrecv(0x11435e000, 0x0, 0x1)
        /usr/local/go/src/runtime/chan.go:577 +0x590
runtime.chanrecv1(0x0?, 0x2?)
        /usr/local/go/src/runtime/chan.go:440 +0x18
runtime.gcenable()
        /usr/local/go/src/runtime/mgc.go:180 +0xc8
runtime.main()
        /usr/local/go/src/runtime/proc.go:209 +0x165
runtime.goexit()
        /usr/local/go/src/runtime/asm_amd64.s:1571 +0x1

jxyang (approved):

jxyang:
@anakrish The go sample failures are root caused to the invalid instruction RDTSCP in Coffee Lake SGX environment. Note the above failures happen on coffee lake only. Do you think we can merge the PR now.

vtikoo (changes requested):
Can you please address the execve concern. Thanks.

anakrish:
> @anakrish The go sample failures are root caused to the invalid instruction RDTSCP in Coffee Lake SGX environment. Note the above failures happen on coffee lake only. Do you think we can merge the PR now?

@jxyang There is a SEGV failure in go samples in IceLake machines. This failure is unrelated to the RDTSCP failure in Coffee Lake machines. However, the SEGV failures can be observed in Mystikos' main branch itself. Based on that, this PR no longer has to be blocked due to go sample failures.

anakrish:
> Can you please address the execve concern. Thanks.

Sure. Let me see if that can be addressed easily.

anakrish:
@vtikoo I have handled SYS_execve and SYS_execveat in myst_syscall. This is consistent with how fork is handled. We will get control whether execve function is called or whether SYS_execve is called via the syscall function.

Note, the current fork implementation has the drawback that if fork is called via the syscall instruction, then myst_fork won't be called. Similarly, the custom thread count management won't be invoked if a syscall instruction is used for execve.
We'd have to address these in the future if we encounter a library that uses syscall instruction for fork/execve.

vtikoo (approved):

PR #1275, opened 2022-03-18
increase pandas timeout further

salsal97:


solutions/pandas_tests/Makefile
```diff
@@ -15,7 +15,7 @@ PYTEST_INI_FILE = pytest.ini
 PYTEST_OPTS = -s -q --no-header --skip-slow --skip-db --skip-network
 TEST_ROOT = /usr/local/lib/python3.9/site-packages/
 TEST = /usr/local/lib/python3.9/site-packages/pandas/tests/io/test_fsspec.py 
-TIMEOUT = 1800
+TIMEOUT = 3000
```
jxyang, line 18: Do we know which test case(s) are taking too long?

salsal97: The increase is explained by the fact that we are building + running in the "make tests" target now..

jxyang: How long does it take to build pandas solution test?

salsal97: I think my previous assessment was incorrect, my guess is that it takes a long time to run on coffelake. ~36 minutes currently on coffeelake whereas it always finishes within 30 minutes on icelake

salsal97: My previous timeout was determined using icelake

jxyang (approved):

PR #1289, opened 2022-03-29
<Testing>Update p1-tests.Jenkinsfile

radhikaj:


.jenkins/pipelines/p1-tests.Jenkinsfile
```diff
@@ -16,7 +16,6 @@ pipeline {
         booleanParam(name: "RUN_DOTNETP1", defaultValue: true, description: "Run .NET P1 tests?")
         booleanParam(name: "RUN_OPENMP_TESTSUITE", defaultValue: true, description: "Run OpenMP Test Suite?")
         booleanParam(name: "ICELAKE_VM", defaultValue: true, description: "Run tests on Ice Lake VMs?")
```
CyanDevs, line 18: You can remove this parameter too since this must be true (if the run were to do anything).

.jenkins/pipelines/p1-tests.Jenkinsfile
```diff
@@ -33,7 +32,6 @@ pipeline {
                     // This is used to test on specific VM generations
                     IGNORE_VM_GENERATIONS = []
```
CyanDevs, line 33: IGNORE_VM_GENERATIONS and the check for ICELAKE_VM can be removed too as it has no functionality.

CyanDevs: IGNORE_VM_GENERATIONS is also on line 97

PR #1290, opened 2022-03-29
Temporarily disable solution/python_flask_tests

asvrada:


vtikoo (approved):

asvrada:
Was able to update flask to fix the issue. See: https://stackoverflow.com/questions/71661851/typeerror-init-got-an-unexpected-keyword-argument-as-tuple

PR #1292, opened 2022-03-29
Remove unreachable syscall SYS_clock_gettime

asvrada:


kernel/syscall.c
```diff
@@ -7659,7 +7650,10 @@ long myst_syscall(long n, long params[6])
     {
         clockid_t clk_id = (clockid_t)params[0];
         struct timespec* tp = (struct timespec*)params[1];
-        return myst_syscall_clock_gettime(clk_id, tp);
+
+        _strace(n, "clk_id=%u tp=%p", clk_id, tp);
```
mingweishih, line 7654: At this point, we are potentially still on the user stack. Not sure if this is a good idea to call the internal trace function with it.

asvrada: Will remove printing strace

mingweishih (approved):

PR #1039, opened 2021-11-22
Temporarily relax bad addr check

mingweishih:
Signed-off-by: Ming-Wei Shih <mishih@microsoft.com>

PR #1175, opened 2022-02-01
Add -enable-brk-syscall option

mikbras:
Add the ``--enable-brk-syscall`` option and add a test to ensure that it is working (with or without the option). Also add a performance test for later reference. (The performance test demonstrates that **mmap** is about 10% faster than **brk**).

PR #1294, opened 2022-03-30
Remove the relaxation of badaddr check

mingweishih:
The detail of the PR is as follows:
- Remove the relaxation of badaddr check introduced in #1039 given  issue with using kstack during the user signal handler was fixed by #1177.
- Ensure the badaddr checks are only enforced in the code path through the syscall interface, avoiding invoking the checks from the kernel.
- Introduce `MYST_INTERNAL` annotation that helps reasoning whether a `myst_` function is private or public to the kernel.

### DIFF
--- a/include/myst/config.h
+++ b/include/myst/config.h
@@ -26,6 +26,6 @@
 
 /* enable to relax the bad addr check (only check if the address is within
  * enclave memory region) */
-#define MYST_RELAX_BAD_ADDR_CHECK
+//#define MYST_RELAX_BAD_ADDR_CHECK
 
 #endif /* _MYST_CONFIG_H */
--- a/include/myst/defs.h
+++ b/include/myst/defs.h
@@ -10,6 +10,8 @@
 
 #define MYST_INLINE static __inline__
 
+#define MYST_INTERNAL static
+
 #define MYST_WEAK __attribute__((weak))
 
 #define MYST_NORETURN __attribute__((__noreturn__))
--- a/include/myst/syscall.h
+++ b/include/myst/syscall.h
@@ -206,8 +206,6 @@ long myst_syscall_load_symbols(void);
 
 long myst_syscall_unload_symbols(void);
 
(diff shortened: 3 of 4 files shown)


vtikoo:
Pseudo-fork child processes share the CRT heap with their parent. For such a child process, bad address check will fail for buffers obtained via malloc.

mingweishih:
> Pseudo-fork child processes share the CRT heap with their parent. For such a child process, bad address check will fail for buffers obtained via malloc.

We only check whether the memory is mapped (with non-zero pid) instead of the ownership of the memory. Will the kernel-allocated memory be used in the scenario you mentioned?

mingweishih:
> > Pseudo-fork child processes share the CRT heap with their parent. For such a child process, bad address check will fail for buffers obtained via malloc.
> 
> We only check whether the memory is mapped (with non-zero pid) instead of the ownership of the memory. Will the kernel-allocated memory be used in the scenario you mentioned?

I see that the shared memory PR changes the logic to check the ownership. I can put this PR on hold until your PR is meged, and the scenario you mentioned is addressed.

vtikoo:
> We only check whether the memory is mapped (with non-zero pid) instead of the ownership of the memory.

Ah I didn't realize that while changing `myst_is_bad_addr` to ownership.

> Will the kernel-allocated memory be used in the scenario you mentioned?

Can't think of any case if `NoBrk` is specified(which is recommended if program is forking). There is an active PR #1175 to disable SYS_brk by default. 
In full disclaimer, I did not encounter such a scenario. This was just reasoning about the consequences of sharing CRT heap in pseudo-fork child processes. Maybe we can detect if the process is a pseudo-fork child and only then relax the check from ownership to additionally check the parent's owned memory. 
The shared memory PR does not enable the pids vector based bad address check, so I don't have any empirical data on whether changing to ownership check will fail certain tests.

mingweishih:
closing for now

PR #1295, opened 2022-04-01
Add runtest wrapper to cpython tests

vtikoo:
Without runtest, cpython failures cause skipping of rest of the tests under solutions/.
PR also removes $(FS) dependency for the dev targets. The dependency is causing the rootfs to be regenerated. 

### DIFF
--- a/solutions/cpython-tests/Makefile
+++ b/solutions/cpython-tests/Makefile
@@ -64,21 +64,21 @@ ifdef MYST_NIGHTLY_TEST
 endif
 
 run-3.8: ext2fs3.8
-	$(MYST_EXEC) $(OPTS) ext2fs3.8 /cpython/python -m test -f /$(TESTFILE) --timeout 120 -v
+	SUBTEST=$(TESTFILE)-3.8 $(RUNTEST) $(MYST_EXEC) $(OPTS) ext2fs3.8 /cpython/python -m test -f /$(TESTFILE) --timeout 120 -v
 	$(foreach i, $(TEST_RUN_INDIVIDUAL_3_8), $(MAKE) one FS=ext2fs3.8 TEST=$(i) $(NL) )
 
 run-3.9: ext2fs3.9
-	$(MYST_EXEC) $(OPTS) ext2fs3.9 /cpython/python -m test -f /$(TESTFILE) --timeout 120 -v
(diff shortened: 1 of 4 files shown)


jxyang (approved):

solutions/cpython-tests/Makefile
```diff
@@ -64,29 +64,29 @@ ifdef MYST_NIGHTLY_TEST
 endif
 
 run-3.8: ext2fs3.8
-	$(MYST_EXEC) $(OPTS) ext2fs3.8 /cpython/python -m test -f /$(TESTFILE) --timeout 120 -v
-	$(foreach i, $(TEST_RUN_INDIVIDUAL_3_8), $(MAKE) one FS=ext2fs3.8 TEST=$(i) $(NL) )
+	$(RUNTEST) $(MYST_EXEC) $(OPTS) ext2fs3.8 /cpython/python -m test -f /$(TESTFILE) --timeout 120 -v
```
salsal97, line 67: Hey Vikas, thank you for adding this, I think this test was moved after I added runtest to solutions. However, in this method the results for the individual tests wont be recorded (the line after this one). Let me know if I am wrong. To avoid this, I used this method to add runtest - https://github.com/deislabs/mystikos/blob/a9c274193803770484a2f5e021908a96748f913b/solutions/dotnet/Makefile#L21

vtikoo: > in this method the results for the individual tests wont be recorded (the line after this one).

Do you mean if the testfile run fails(L67), runtest won't run and/or record the individual runs?

salsal97: No.. I meant L68 -> If any of those fail, it would not be recorded. Also I'm not 100% sure which name would be picked up for this test, but if the 3.8/3.9/3.10 distinction is not made then it might overwrite the results as well since it basically just puts it in a file with the testname

asvrada (approved):

solutions/cpython-tests/Makefile
vtikoo: > I meant L68 -> If any of those fail, it would not be recorded.

I moved the runtest wrapper to match the pattern you had mentioned earlier in the thread.

>  if the 3.8/3.9/3.10 distinction is not made then it might overwrite the results

Thanks for pointing this out. Changed to differentiate the runs via `SUBTEST` environment variable.

salsal97 (approved):
Looks great! Thank you for adding the documentation for the runtest script!

PR #1296, opened 2022-04-04
Change clock res to 100ns

asvrada:
Also re-enable Cpython uuid test cases

### DIFF
--- a/tools/myst/enc/clock.c
+++ b/tools/myst/enc/clock.c
@@ -72,7 +72,10 @@ static long _get_monotime()
     {
         // maintain monotonicity.
         // TODO: issue a warning. Host might be playing tricks.
-        return ++prev;
+        // Changed resolution from 1 to 100 to address an issue
+        // where uuid generation is not unique in Cpython
+        prev += 100;
+        return prev;
     }
 }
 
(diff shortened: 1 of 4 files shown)


asvrada:
Manual pipeline: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/job/Mystikos/job/Pull-Request-Pipeline-Manual/198/

jxyang (approved):

jxyang:
Thanks for doing this, @asvrada

tools/myst/enc/clock.c
```diff
@@ -72,7 +72,8 @@ static long _get_monotime()
     {
         // maintain monotonicity.
         // TODO: issue a warning. Host might be playing tricks.
-        return ++prev;
+        prev += 100;
```
vtikoo, line 75: Can you add a comment here on why we are incrementing by 100.

asvrada: Thanks for pointing out. Added a comment

jxyang: The precision is still determined by how often the host thread updates the clock (which is every microsecond). But we are saying for two sub-microsecond consecutive clock_gettime queries, we spread them out 100ns apart, which is more realistic because even natively, a clock_gettime query takes the kernel 20-50 ns to handle.

PR #1297, opened 2022-04-05
protect filesystems from signals

paulcallen:

### DIFF
--- a/kernel/lockfs.c
+++ b/kernel/lockfs.c
@@ -7,6 +7,7 @@
 #include <myst/eraise.h>
 #include <myst/lockfs.h>
 #include <myst/mutex.h>
+#include <myst/signal.h>
 #include <myst/thread.h>
 
 #define LOCKFS_MAGIC 0x94639c1a101f4a1d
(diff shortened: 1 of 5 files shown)


kernel/lockfs.c
```diff
@@ -19,42 +20,53 @@ typedef struct lockfs
...
-    myst_mutex_unlock(sig_handler->lock);
+#if ULONG_MAX == 0xffffffff
+    set->__bits[0] = 0x7ffffffful;
+    set->__bits[1] = 0xfffffffcul;
+    if (_NSIG > 65)
+    {
+        set->__bits[2] = 0xfffffffful;
+        set->__bits[3] = 0xfffffffful;
+    }
+#else
```
mikbras, line 49: ulong max is always 64-bit on x86_64.

mikbras (changes requested):
I created a new branch (paulcallen/lockfs-signal-v2) with two minor modifications:
- included musl sigfillset.c in kernel directly to avoid licensing issue from cut-n-paste.
- factored out LOCK/UNLOCK macros by moving conditional check inside lock() function.

mikbras:
I can't see anything functionally wrong but needed to at least work around licensing implication of including musl code in the kernel.

paulcallen:
pulled in the sigfillset.c change from your branch

paulcallen:
unfortunately, moving the signal disabling into the lock creates a race condition that can leave the lock active when an exception occurs.

mikbras (approved):
Thank you for doing this. Looks good.

PR #1310, opened 2022-04-26
adding more signal protection if closed from signal handler

paulcallen:

### DIFF
--- a/kernel/signal.c
+++ b/kernel/signal.c
@@ -908,10 +908,23 @@ done:
 
 void myst_handle_host_signal(siginfo_t* siginfo, mcontext_t* mcontext)
 {
+    void* original_fsbase;
+    myst_thread_t* thread;
+    myst_td_t* target_td;
+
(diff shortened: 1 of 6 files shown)


kernel/signal.c
```diff
@@ -908,11 +908,24 @@ long myst_signal_clone(
...
+    void* original_fsbase;
+    myst_thread_t* thread;
+    myst_td_t* target_td;
+
+    original_fsbase = myst_get_fsbase();
+    myst_assume(myst_tcall_get_tsd((uint64_t*)&thread) == 0);
+
+    target_td = thread->target_td;
+
+    myst_set_fsbase(target_td);
```
mingweishih, line 920: Just curious in what flow that leads to this point with user td.

mingweishih: This is a good catch. When the exception occurs in the user code, we will eventually reach here with user td. We need to ensure using kernel td before calling `_handle_one_signal`.

PR #1313, opened 2022-04-28
work around timeout for first test to get run taking too long

paulcallen:

### DIFF
--- a/tests/fork/Makefile
+++ b/tests/fork/Makefile
@@ -27,7 +27,7 @@ tests:
 
 # no fork
 test-no-fork: rootfs
-	$(RUNTEST) $(MYST_EXEC) $(OPTS) --app-config-path config-none.json rootfs /bin/fork nofork
+	TIMEOUT=360 $(RUNTEST) $(MYST_EXEC) $(OPTS) --app-config-path config-none.json rootfs /bin/fork nofork
 	$(RUNTEST) $(MYST_EXEC) $(OPTS) rootfs /bin/fork nofork
 	$(RUNTEST) $(MYST_EXEC) $(OPTS) --fork-mode none rootfs /bin/fork nofork
 
--- a/tests/myst/mkcpio/Makefile
+++ b/tests/myst/mkcpio/Makefile
@@ -16,7 +16,7 @@ REDEFINE_TESTS=1
 include $(TOP)/rules.mak
 
 tests:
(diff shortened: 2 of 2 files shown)


tests/fork/Makefile
```diff
@@ -27,7 +27,7 @@ tests:
 
 # no fork
 test-no-fork: rootfs
-	$(RUNTEST) $(MYST_EXEC) $(OPTS) --app-config-path config-none.json rootfs /bin/fork nofork
+	TIMEOUT=360 $(RUNTEST) $(MYST_EXEC) $(OPTS) --app-config-path config-none.json rootfs /bin/fork nofork
```
jxyang, line 30: I'd love to understand why this simple test could take 6 minutes.

paulcallen: Basically _something_ is causing the first test to get executed to take an undetermined amount of time. I have not been able to narrow this down more at this point, but my current thoughts are that one of the shared libraries the myst tool links again is doing something during the initialization of the library and that sometimes takes longer than it should.
It generally takes less than a couple of minutes, but sometimes it is a little more.
The hang happens _before_ main() is even hit in the process which is why I think this is something to do with the first test being called.

tests/myst/mkcpio/Makefile
```diff
@@ -16,7 +16,7 @@ REDEFINE_TESTS=1
 include $(TOP)/rules.mak
 
 tests:
-	$(RUNTEST) $(MAKE) test-mkcpio
```
vtikoo, line 19: Is this intentional?

paulcallen: yep! There are several things in the individual test that take a long time and each has $(runtest). Doing it at this layer can cause the test to timeout prematurely. Creating large cpio/ext2 images can take a bit of time.

vtikoo (approved):

tests/fork/Makefile
jxyang: I am not a fan of making tests pass by increasing the time outs or memory sizes. The actual question we should ask is: how could a previous version of Mystikos pass the test with a smaller time out or memory size. If we continue on the path of increasing time outs or memory sizes, we are on a deteriorating path.

vtikoo: We should definitely rootcause this, but there is also the issue of this blocking PRs.

PR #1318, opened 2022-05-03
Abstract namespace addresses support for Unix domain sockets

vtikoo:
Conventionally Unix domain socket addresses are a path in the file hierarchy. Abstract namespace identifiers allow addresses without creating a corresponding file path. They are differentiated from conventional UDS addresses by starting the address with a null character.

Pros:
- Cleaning up for socket file path is not required.
- User does not need permissions on the directories in the path.

Cons:
- No access control by default: Unix file permissions do not carry over to Abstract namespace identifiers.
- Portability: Linux only feature.

### DIFF
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -387,17 +387,36 @@ static int _socketaddr_to_str(
         ERAISE(-EFAULT);
     }
 
-    const uint8_t* p = (uint8_t*)addr->sa_data;
-    uint16_t port = (uint16_t)((p[0] << 8) | p[1]);
-    const uint8_t ip1 = p[2];
-    const uint8_t ip2 = p[3];
-    const uint8_t ip3 = p[4];
-    const uint8_t ip4 = p[5];
-
-    if (snprintf(out, limit, "%u.%u.%u.%u:%u", ip1, ip2, ip3, ip4, port) >=
-        (int)limit)
+    if (addr->sa_family == AF_UNIX)
     {
-        ERAISE(-ENAMETOOLONG);
+        if (addr->sa_data[0] &&
+            myst_snprintf(out, limit, "%s", addr->sa_data) == ERANGE)
+        {
+            ERAISE(-ENAMETOOLONG);
+        }
+        // abstract namespace address
+        else if (
+            myst_snprintf(
+                out, limit, "%s(abstract namespace)", addr->sa_data + 1) ==
+            ERANGE)
+        {
+            ERAISE(-ENAMETOOLONG);
+        }
+    }
+    else
+    {
+        const uint8_t* p = (uint8_t*)addr->sa_data;
+        uint16_t port = (uint16_t)((p[0] << 8) | p[1]);
+        const uint8_t ip1 = p[2];
+        const uint8_t ip2 = p[3];
+        const uint8_t ip3 = p[4];
+        const uint8_t ip4 = p[5];
+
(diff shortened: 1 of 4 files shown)


kernel/syscall.c
```diff
@@ -387,17 +387,36 @@ static int _socketaddr_to_str(
...
-    const uint8_t ip3 = p[4];
-    const uint8_t ip4 = p[5];
-
-    if (snprintf(out, limit, "%u.%u.%u.%u:%u", ip1, ip2, ip3, ip4, port) >=
-        (int)limit)
+    if (addr->sa_family == AF_UNIX)
     {
-        ERAISE(-ENAMETOOLONG);
+        if (addr->sa_data[0] &&
+            myst_snprintf(out, limit, "%s", addr->sa_data) == ERANGE)
```
jxyang, line 393: Should it be `-ERANGE`?

kernel/syscall.c
```diff
@@ -387,17 +387,36 @@ static int _socketaddr_to_str(
...
+        {
+            ERAISE(-ENAMETOOLONG);
+        }
+        // abstract namespace address
+        else if (
+            myst_snprintf(
+                out, limit, "%s(abstract namespace)", addr->sa_data + 1) ==
+            ERANGE)
```
jxyang, line 401: Likewise

kernel/syscall.c
```diff
@@ -387,17 +387,36 @@ static int _socketaddr_to_str(
...
-        ERAISE(-ENAMETOOLONG);
+        if (addr->sa_data[0] &&
+            myst_snprintf(out, limit, "%s", addr->sa_data) == ERANGE)
+        {
+            ERAISE(-ENAMETOOLONG);
+        }
+        // abstract namespace address
+        else if (
+            myst_snprintf(
+                out, limit, "%s(abstract namespace)", addr->sa_data + 1) ==
```
jxyang, line 400: maybe just "abstract" or "abs ns" to reduce the chance of overflow.

solutions/cpython-tests/Makefile
```diff
@@ -109,6 +109,14 @@ To enable python source listing, do \n\
 before launching gdb.\033[0m\n')" \
            --args $(MYST_EXEC) $(OPTS) $(FS) /cpython/python -m unittest Lib.test.$(TESTCASE) -v
 
+testcase-mpdb:
+	killall myst 2> /dev/null || echo ""
```
jxyang, line 113: Is this related?

tests/sockets/sockets.c
```diff
@@ -260,13 +281,16 @@ void test_sockets(args_t* args)
     pthread_t srv_thread;
     pthread_t cli_thread;
 
+    pthread_barrier_init(&bar, NULL, 2);
+
     assert(pthread_create(&srv_thread, NULL, _srv_thread_func, args) == 0);
-    _sleep_msec(100);
```
jxyang, line 264: Why the sleep is no longer needed?

vtikoo: I changed the test to use a `pthread_barrier_t`.

solutions/cpython-tests/Makefile
vtikoo: This was useful in debugging cpython stacktraces. The related test `test_multiprocessing_spawn` is not part of this PR, being blocked by shared memory support.

kernel/syscall.c
vtikoo: Thanks for catching this!

jxyang (approved):

PR #1320, opened 2022-05-04
Use docker save <image-name> to extract rootfs directory

vtikoo:

### DIFF
--- a/scripts/appbuilder
+++ b/scripts/appbuilder
@@ -215,29 +215,38 @@ export_image()
     TEMP_FILE=$(mktemp)
     TEMP_DIR=$(mktemp -d)
     APPENV_FILENAME="appenv.json"
-
-    # note that we have to start up the image to flatten it
-    TEMP_INAME=$(docker run -d $IMAGE_NAME)
-    TEMP_APPENV=$APPENV_FILENAME-$TEMP_INAME
+    TEMP_APPENV=$APPENV_FILENAME-$(date +%s)
     
     docker image inspect --format='{{json .Config}}' $IMAGE_NAME > $TEMP_APPENV
 
-    docker stop $TEMP_INAME >/dev/null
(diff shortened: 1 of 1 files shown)


scripts/appbuilder
```diff
@@ -215,29 +215,25 @@ export_image()
...
-
-    # create the 'appdir' from the docker export
+    # create 'appdir' with docker save
+    mkdir -p $TEMP_DIR/appdir
+    mv $TEMP_APPENV $TEMP_DIR/appdir/$APPENV_FILENAME
+    docker save $IMAGE_NAME -o $TEMP_FILE
     tar xf $TEMP_FILE -C $TEMP_DIR
-    mv $TEMP_APPENV $TEMP_DIR/$APPENV_FILENAME
+    cd $TEMP_DIR/appdir
+    find $TEMP_DIR -name layer.tar -exec sh -c 'tar xf {}' \;
```
vtikoo, line 228: @jxyang could you clarify whats happening here.

jxyang: I believe the original image was extracted into the following structure:
```
   temp_dir
     <uuid-for-layer0>
           layer.tar
     <uuid-for-layer2>
          layer.tar
     ...
```
The script is extracting all the tar balls  of all layers into appdir.

jxyang: Since appdir is also under $TEMP_DIR, will line 228 also finds layer.tar in appdir and mistakenly extract it? I think to be safe we should organize temp_dir like this:

```
   temp_dir
     layers
        <uuid-for-layer0>
              layer.tar
        <uuid-for-layer1>
             layer.tar
        <uuid-for-layer2>
             layer.tar
         <...more layers...>
    appdir
```
Then on line 228, we can use `find $TEMP_DIR/layers` to be more precise.

vtikoo:
Superseded by #1321

PR #1329, opened 2022-05-18
Add build stage to Azure SDK tests

asvrada:
Previously the pipeline would build and run tests, which easily exceeded the timeout value. This PR move the build part to an individual step and only time the `make tests` part. So it's less likely to timeout


asvrada:
Passed Azure SDK pipeline as well: https://openenclaveci.westus.cloudapp.azure.com/blue/organizations/jenkins/Mystikos%2FStandalone-Pipelines%2FAzure-SDK-Tests-Pipeline/detail/Azure-SDK-Tests-Pipeline/135/pipeline

jxyang (approved):

radhikaj:
Thanks @asvrada

PR #1336, opened 2022-05-28
Add support for unix domain sockets bound to hostfs paths

vtikoo:
For UDS sockets its not known at socket creation time whether socket
will bind or connect to a hostfs path.

This patch introduces myst_sockdev_reresolve, which is called from
SYS_bind and SYS_connect syscall handlers. If the socket is an UDS
socket and the address provided is a hostfs path, this method does the
following things -

- create new struct sockaddr_un address with sun_path set to absolute
  path name of socket on host. Address is dynamically allocated in this
  method and should be freed by the caller once its done using it.

- Get socket type from Mystikos kernel's internal UDS socket device,
  and close the internal UDS socket.

- Create socket using the host socket device.

- Patch the mount table entry for the socket, updating the device and
  socket fields.

Once back in the SYS_bind and SYS_connect handlers, the newly created
host device socket and host sock address is passed to host socket 
device's bind or connect routine.

Assumption:
For datagram sockets, the clients call `connect()` before starting communication with `sendto()` or `recvfrom()`

### DIFF
--- a/kernel/fdtable.c
+++ b/kernel/fdtable.c
@@ -666,3 +666,38 @@ ssize_t myst_fdtable_count(const myst_fdtable_t* fdtable)
 done:
     return ret;
 }
+
+int myst_fdtable_update_sock_entry(
+    myst_fdtable_t* fdtable,
+    int fd,
+    myst_sockdev_t* device,
+    myst_sock_t* new_sock)
+{
+    int ret = 0;
+
+    if (!fdtable || !device || !new_sock)
+        ERAISE(-EINVAL);
+
+    if (!(fd >= 0 && fd < MYST_FDTABLE_SIZE))
+        ERAISE(-EBADF);
+
+    myst_spin_lock(&fdtable->lock);
+    {
+        myst_fdtable_entry_t* entry = &fdtable->entries[fd];
+
+        if (entry->type != MYST_FDTABLE_TYPE_SOCK ||
+            !(entry->device && entry->object))
+        {
+            myst_spin_unlock(&fdtable->lock);
+            ERAISE(-ENOTSOCK);
+        }
+
+        entry->device = device;
+        entry->object = new_sock;
+    }
+    myst_spin_unlock(&fdtable->lock);
+
+done:
+
+    return ret;
+}
--- a/kernel/sockdev.c
+++ b/kernel/sockdev.c
@@ -6,9 +6,13 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/ioctl.h>
+#include <sys/un.h>
 
 #include <myst/eraise.h>
+#include <myst/fdtable.h>
+#include <myst/hostfs.h>
 #include <myst/iov.h>
+#include <myst/mount.h>
 #include <myst/panic.h>
 #include <myst/sockdev.h>
 #include <myst/spinlock.h>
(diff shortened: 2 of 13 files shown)


radhikaj:
@Francis-Liu  @jxyang @paulcallen @mikbras , can you please review? Thanks

kernel/sockdev.c
```diff
@@ -926,3 +936,106 @@ int myst_sockdev_resolve(
...
+        char suffix[PATH_MAX];
+    };
+    struct locals* locals = NULL;
+    myst_sock_t* hostuds_sock = NULL;
+    struct sockaddr_un* new_addr = NULL;
+
+    if (sockfd < 0 || !sockdev || !sock || !_valid_sockdev(sockdev))
+        ERAISE(-EINVAL);
+
+    if (!reresolved || !sock_out || !sockdev_out || !addr_out || !addrlen_out)
```
jxyang, line 964: No need to check `resolved`. It is reset anyway next line.

kernel/syscall.c
```diff
@@ -2374,11 +2374,32 @@ int myst_syscall_bind(
...
+        &new_sock,
+        &new_addr,
+        &new_addrlen));
+
+    if (!reresolved)
+        ret = (*sd->sd_bind)(sd, sock, addr, addrlen);
+    else
+    {
+        ret = (*new_sd->sd_bind)(new_sd, new_sock, new_addr, new_addrlen);
+        free(new_addr);
```
jxyang, line 2401: When do we free the old `sock` and `addr`?

kernel/sockdev.c
```diff
@@ -926,3 +936,106 @@ int myst_sockdev_resolve(
 done:
     return ret;
 }
+
+int myst_sockdev_reresolve(
```
jxyang, line 940: Some comments about the purpose of the parameters would be helpful.

Would it possible to reuse the `sockdev`, `sock`, `addr`, and `addrlen` instead of new ones? We can clean up the old ones in this function if the socket is re-resolved.

kernel/syscall.c
vtikoo: In case of reresolution, `myst_sockdev_reresolve` closes the old `sock` object. `addr` is a buffer provided by user-space, so we don't free it. `new_addr` needs to be freed because the kernel allocated it to be able to pass the modified address to the host socket device.

kernel/fdtable.c
```diff
@@ -666,3 +666,35 @@ ssize_t myst_fdtable_count(const myst_fdtable_t* fdtable)
...
+        if (entry->type != MYST_FDTABLE_TYPE_SOCK ||
+            !(entry->device && entry->object))
+            ERAISE(-ENOTSOCK);
+
+        entry->device = device;
+        entry->object = new_sock;
+    }
+
+done:
+    myst_spin_unlock(&fdtable->lock);
```
Francis-Liu, line 697: Do you need to check if `fdtable->lock` is locked?

jxyang (approved):

jxyang:
Manual nightly pipeline run:

https://openenclaveci.westus.cloudapp.azure.com/blue/organizations/jenkins/Mystikos%2FNightly-Pipeline-Manual/detail/Nightly-Pipeline-Manual/19/pipeline

kernel/sockdev.c
vtikoo: Added comments in the header file entry for this function.

kernel/fdtable.c
vtikoo: Thanks for catching that.

PR #1347, opened 2022-06-09
Fix unbounded write

asvrada:
https://github.com/deislabs/mystikos/security/code-scanning?query=%22unbounded+write%22+is%3Aopen+branch%3Amain+autofilter%3Atrue+

### DIFF
--- a/host/maps.c
+++ b/host/maps.c
@@ -10,6 +10,7 @@
 #include <unistd.h>
 
 #include <myst/maps.h>
+#include <myst/strings.h>
 
 void myst_maps_dump1(const myst_maps_t* maps)
 {
@@ -100,10 +101,11 @@ int myst_maps_load(myst_maps_t** maps_out)
         uint64_t inode;
         myst_maps_t* maps;
         char path[PATH_MAX] = "";
+        int path_offset = 0;
 
         n = sscanf(
             line,
-            "%lx-%lx %c%c%c%c %lu %u:%u %lu %s",
+            "%lx-%lx %c%c%c%c %lu %u:%u %lu %n",
             &start,
             &end,
             &r,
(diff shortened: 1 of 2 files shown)


host/maps.c
```diff
@@ -103,7 +103,7 @@ int myst_maps_load(myst_maps_t** maps_out)
 
         n = sscanf(
             line,
-            "%lx-%lx %c%c%c%c %lu %u:%u %lu %s",
+            "%lx-%lx %c%c%c%c %lu %u:%u %lu %4096s",
```
radhikaj, line 106: where does this 4096 come from

asvrada: The max size for path is 4096 bytes

radhikaj: > The path variable is of 4096 bytes in size

Is there a #define we can use instead?

mikbras: Hmm? I wish there were a way without hardcoding the path length.

mikbras: Okay, I verified that the following works:

```
        int path_offset;

        n = sscanf(
            line,
            "%lx-%lx %c%c%c%c %lu %u:%u %lu %n",
            &start,
            &end,
            &r,
            &w,
            &x,
            &f,
            &o,
            &minor,
            &major,
            &inode,
            &path_offset);

        if (n < 9)
        {
            ret = -ENOSYS;
            goto done;
        }

        *path = '\0';
        strncat(path, &line[path_offset], PATH_MAX);
```

mikbras: Also, you probably want to check whether ``streln(*line[path_offset]) is bigger than ``PATH_MAX``.

mikbras (approved):
Thank you for doing this!

PR #1348, opened 2022-06-09
Ignore third_party CodeQL issues

asvrada:
Used 3rd party action to filter out unwanted issues
https://github.com/zbazztian/filter-sarif

Had a test run against this PR, and it successfully ignores third party issues, reduced number of issues from 485 to 95

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -47,6 +47,22 @@ jobs:
     - name: Building Mystikos
       run: |
         make MYST_PRODUCT_BUILD=1 -j4
-              
+
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v1
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -47,6 +47,22 @@ jobs:
...
-              
+
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v1
+      with:
+        upload: false
+        output: sarif-results
+
+    - name: Remove third_party issues from SARIF
+      uses: zbazztian/filter-sarif@master
```
salsal97, line 58: Hey Zijie, dont know how frequently this repo would be updated but in the past we have had issues with pinning a reference to the main branch of an outside repo. Should we consider pinning it to a release/commit hash instead?

radhikaj:

PR #1349, opened 2022-06-09
Fix codeql Uncontrolled data errors

jxyang:


asvrada:
A general comment, not sure if we should use assert on user input?

jxyang:
Crash is one of the self-defense mechanisms when a program faces malicious input. We could add a message to make it more graceful. Do you have alternatives in mind?

asvrada:
Could we have an input validation component (we don't seem to have one for now) that processes all user input from the command line args before moving on to use them? Adding a message to better let users know what happened should be good for now. Thanks!

jxyang:
> Could we have an input validation component (we don't seem to have one for now) that processes all user input from the command line args before moving on to use them? Adding a message to better let users know what happened should be good for now. Thanks!

 This is a nice to have. But we are dealing with multiple (sub)commands here, `exec`, `exec-linux`, `mkext2`, `mount-docker-image`, etc. Some of them have just <100 lines of source code. So an input validation module is an overkill to them.

asvrada (approved):

PR #1351, opened 2022-06-10
Add config to choose b/w host socket device and Mystikos socket device for Unix domain sockets

vtikoo:
### Summary
- Add configuration flag to choose host socket device for Unix domain sockets.  Can be specified via `HostUDS` in JSON config file or `--host-uds` on the command line.
- Turn off switchless ocall for epoll_wait. This is to avoid issues with thread interruption of threads performing switchless calls.

### DIFF
--- a/kernel/sockdev.c
+++ b/kernel/sockdev.c
@@ -988,7 +988,7 @@ int myst_host_uds_addr_reresolve(
 
     const struct sockaddr_un* sun = (const struct sockaddr_un*)addr;
 
-    /* abstract namespace UDSes need no address updation */
+    /* abstract namespace UDSs need no address update */
     if (*sun->sun_path == '\0')
         return 0;
 
--- a/kernel/sockdev.c
+++ b/kernel/sockdev.c
@@ -1007,12 +1007,26 @@ int myst_sockdev_reresolve(
     socklen_t optlen = sizeof(type);
     ECHECK(myst_udsdev->sd_getsockopt(
         myst_udsdev, sock, SOL_SOCKET, SO_TYPE, &type, &optlen));
+
+    struct timeval so_sndtimeo;
+    optlen = sizeof(struct timeval);
+    ECHECK(myst_udsdev->sd_getsockopt(
+        myst_udsdev, sock, SOL_SOCKET, SO_SNDTIMEO, &so_sndtimeo, &optlen));
+
     ECHECK(myst_udsdev->sd_close(myst_udsdev, sock));
 
     // create the hostdev socket
     myst_sockdev_t* host_sockdev = myst_sockdev_get();
     ECHECK(host_sockdev->sd_socket(
         host_sockdev, AF_LOCAL, type, 0, &hostuds_sock));
+
+    ECHECK(host_sockdev->sd_setsockopt(
+        host_sockdev,
+        hostuds_sock,
+        SOL_SOCKET,
+        SO_SNDTIMEO,
+        &so_sndtimeo,
+        optlen));
 
     // update mount entry
     ECHECK(myst_fdtable_update_sock_entry(
--- a/kernel/sockdev.c
+++ b/kernel/sockdev.c
@@ -13,6 +13,7 @@
 #include <myst/hostfs.h>
 #include <myst/iov.h>
 #include <myst/mount.h>
+#include <myst/options.h>
 #include <myst/panic.h>
 #include <myst/sockdev.h>
 #include <myst/spinlock.h>
@@ -908,7 +909,11 @@ int myst_sockdev_resolve(
                 ERAISE(-ENOTSUP);
             }
 
-            *dev = myst_udsdev_get();
+            if (!__options.host_uds)
+                *dev = myst_udsdev_get();
+            else
+                *dev = myst_sockdev_get();
+
             goto done;
         }
         case AF_INET:
@@ -937,15 +942,13 @@ done:
     return ret;
 }
 
-int myst_sockdev_reresolve(
+int myst_host_uds_addr_reresolve(
     int sockfd,
     myst_sockdev_t* sockdev,
     myst_sock_t* sock,
     const struct sockaddr* addr,
     socklen_t addrlen,
     bool* reresolved,
-    myst_sockdev_t** sockdev_out,
-    myst_sock_t** sock_out,
     struct sockaddr** addr_out,
     socklen_t* addrlen_out)
 {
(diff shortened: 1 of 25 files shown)


kernel/udsdev.c
```diff
@@ -1324,6 +1340,14 @@ static int _udsdev_setsockopt(
             _obj(sock)->so_rcvbuf = _max(_obj(sock)->so_rcvbuf, MIN_SO_RCVBUF);
             break;
         }
+        case SO_SNDTIMEO:
+        {
+            if (!optval)
+                ERAISE(-EINVAL);
+
+            memcpy(&_obj(sock)->so_sndtimeo, optval, sizeof(struct timeval));
```
mikbras, line 1348: Add check that ``optlen == sizeof(struct timeval)`` and fail if not. This can result in a free memory read if the caller passes in something smaller than ``struct timeval``.

tests/hostfs_uds/host_uds_server.c
```diff
@@ -81,6 +86,8 @@ int main(int argc, const char* argv[])
 
     close(srvrfd);
     unlink(sock_path);
+    if (ready)
+        unlink(ready_path);
 
     return 0;
 }
```
mikbras, line 93: Run code format.

tools/myst/config.c
```diff
@@ -423,6 +423,15 @@ static json_result_t _json_read_callback(
                 else
                     CONFIG_RAISE(JSON_TYPE_MISMATCH);
             }
+            else if (json_match(parser, "HostUDS") == JSON_OK)
+            {
+                if (type == JSON_TYPE_BOOLEAN)
+                    parsed_data->host_uds = un->boolean;
+                else if (type == JSON_TYPE_INTEGER)
+                    parsed_data->host_uds = (un->integer == 0) ? false : true;
```
mikbras, line 431: If the type of ``parsed_data->host_uds`` is ``bool``, then the following will be fine.

```
parsed_data->host_uds = (un->integer == 0);
```
The compiler automatically injects code to set ``bool`` to ``0`` or ``1``. The following program prints ``1``.

```
#include <stdio.h>
#include <stdbool.h>

int main()
{
    printf("%u\n", (bool)99);
    return 0;
}
```

tools/myst/host/package.c
```diff
@@ -666,6 +666,17 @@ int _exec_package(
         goto done;
     }
 
+    /* Check --host-uds option */
+    if (cli_getopt(&argc, argv, "--host-uds", NULL) == 0)
+    {
+        fprintf(
```
mikbras, line 672: I realize there are many of these places where we use ``fprintf(stderr, ...)`` but the correct form is to use ``_err()``, which prefixes "argv[0]". This is a defacto standard in Linux where commands must prefix all errors with the name of the command (because the command can call another command and without the prefix it would be unclear which command printed the errors). For example:

```
$ ls xxx
ls: cannot access 'xxx': No such file or directory
```

tools/myst/host/package.c
```diff
@@ -864,6 +875,7 @@ int _exec_package(
     */
     options.nobrk = parsed_data.no_brk ? true : false;
     options.exec_stack = parsed_data.exec_stack ? true : false;
+    options.host_uds = parsed_data.host_uds ? true : false;
```
mikbras, line 878: Would use direct assignment (for all three lines). The source and destination are both ``bool``.

tools/myst/myst.edl
```diff
@@ -159,8 +159,7 @@ enclave
             int epfd,
             [out, count=maxevents] struct epoll_event* events,
             size_t maxevents,
-            int timeout)
-            transition_using_threads;
+            int timeout);
```
mikbras, line 162: Did this cause any performance issues?

mikbras (commented):
Nice work Vikas! Some minor changes, mostly stylistic. In once place there is a memory leak of a dynamically allocated sockaddr (I suggest using ``struct sockaddr_storage``).

kernel/sockdev.c
```diff
@@ -955,25 +958,37 @@ int myst_sockdev_reresolve(
...
+        &address_family,
+        &af_optlen));
+
+    if (address_family != AF_UNIX)
+        return 0;
 
     const struct sockaddr_un* sun = (const struct sockaddr_un*)addr;
 
-    /* abstract namespace UDSes are always myst kernel internal */
+    /* abstract namespace UDSes need no address updation */
```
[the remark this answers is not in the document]
mikbras: Also, the MLA (style guide) says that the possesive of an acronym is formed by adding an ``s`` to the end (so UDSs).

kernel/udsdev.c
mikbras: Also should fail if optlen is null.

tests/hostfs_uds/host_uds_server.c
```diff
@@ -10,9 +11,13 @@
 #define PATH_MAX 4096
 
 const char* hostdir = NULL;
+char ready_path[PATH_MAX];
+bool ready = false;
 
 void failed()
 {
+    if (ready)
```
mikbras, line 19: You could use and ``atexit`` handler to remove this file.

```
int atexit(void (*function)(void));
```

kernel/sockdev.c
```diff
@@ -908,7 +909,11 @@ int myst_sockdev_resolve(
                 ERAISE(-ENOTSUP);
             }
 
-            *dev = myst_udsdev_get();
+            if (!__options.host_uds)
```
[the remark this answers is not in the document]
vtikoo: I had kept the negative check first as that would be the common case. Do you think its an unnecessary optimization?

tests/hostfs_uds/host_uds_server.c
```diff
@@ -10,9 +11,13 @@
 #define PATH_MAX 4096
 
 const char* hostdir = NULL;
+char ready_path[PATH_MAX];
+bool ready = false;
```
mikbras, line 15: You could get rid the ``ready`` variable and just do this check:

```
    if (*ready_path)
        unlink(ready_path);
```

Globals and statics a guaranteed to be zero-initialized in C.

tests/hostfs_uds/config.json
```diff
@@ -0,0 +1,32 @@
...
+    // The heap size of the user application. Increase this setting if your app experienced OOM.
+    "MemorySize": "1g",
+    // The path to the entry point application in rootfs
+    "ApplicationPath": "/app/hello",
+    // The parameters to the entry point application
+    "ApplicationParameters": [],
+    // Whether we allow "ApplicationParameters" to be overridden by command line options of "myst exec"
+    "HostApplicationParameters": false,
+    // The environment variables accessible inside the enclave.
+    "EnvironmentVariables": ["COMPlus_EnableDiagnostics=0"],
```
jxyang, line 20: Does the test pass without the flag?

kernel/sockdev.c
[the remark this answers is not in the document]
vtikoo: Good to know!

kernel/syscall.c
```diff
@@ -2374,30 +2374,21 @@ int myst_syscall_bind(
...
-    ECHECK(myst_sockdev_reresolve(
-        sockfd,
-        sd,
-        sock,
-        addr,
-        addrlen,
-        &reresolved,
-        &new_sd,
-        &new_sock,
-        &new_addr,
```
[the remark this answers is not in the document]
vtikoo: > I don't see any code pathway where new_addr can be freed in the success case.

`new_addr` is allocated only if: `sd` is the host socket device, `sock` is an UDS socket and `addr` points a path on a hostfs mount. In this case `myst_host_uds_addr_reresolve` sets `reresolved` to true, and `new_addr` is used to complete the syscall on the host and freed afterwards.

kernel/udsdev.c
vtikoo: Added a check to fail if `optlen < sizeof(struct timeval)`.

tests/hostfs_uds/host_uds_server.c
vtikoo: Fixed.

tests/hostfs_uds/host_uds_server.c
```diff
@@ -81,6 +86,8 @@ int main(int argc, const char* argv[])
 
     close(srvrfd);
     unlink(sock_path);
+    if (ready)
+        unlink(ready_path);
```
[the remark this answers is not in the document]
vtikoo: The make target `rm-hostdir` achieves this.

tests/hostfs_uds/config.json
vtikoo: Yes.

tools/myst/myst.edl
vtikoo: There does seem to be a performance impact on solutions/redis benchmark.

kernel/syscall.c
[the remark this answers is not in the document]
mikbras: In any case the allocation is unnecessary if you were to use ``struct sockaddr_storage``.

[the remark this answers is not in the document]
mikbras: I see. Thanks.

mikbras (approved):
Looks good Vikas. Thank you for doing this!

PR #1359, opened 2022-06-15
Use non exiting version of error print function - _err_noexit

vtikoo:

### DIFF
--- a/tools/myst/host/package.c
+++ b/tools/myst/host/package.c
@@ -35,6 +35,10 @@
 
 _Static_assert(PAGE_SIZE == 4096, "");
 
+#define ERR_NOEXIT(...)       \
+    _err_noexit(__VA_ARGS__); \
+    ret = -1
+
 #define USAGE_PACKAGE \
     "\
 \n\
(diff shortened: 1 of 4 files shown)


tools/myst/host/package.c
```diff
@@ -128,23 +128,21 @@ int _package(int argc, const char* argv[])
     if ((argc < 4) || (cli_getopt(&argc, argv, "--help", NULL) == 0) ||
         (cli_getopt(&argc, argv, "-h", NULL) == 0))
     {
-        _err(USAGE_PACKAGE, argv[0], argv[0]);
+        _err_noexit(USAGE_PACKAGE, argv[0], argv[0]);
```
jxyang, line 131: Can you turn this into a macro in which "ret" is set to 1, just to match `_err` behavior?

vtikoo: In #1351 I replaced `fprintf(stderr` with `_err(` a little too aggressively. `_err` has the advantage of printing the program name along with the error. However, I didn't realize that it also calls `exit(1)`. This PR is replacing most of those uses with `_err_noexit`, which does print the program name, but does not exit.

jxyang (approved):

PR #1367, opened 2022-06-20
Run make prereqs as part of make init

asvrada:
Currently the order of events is as follows:
1. make third_party
2. make prereqs

This PR reverse that.


asvrada:
This would fix CodeQL failures

radhikaj (approved):

PR #1368, opened 2022-06-20
Check and exit on overflow

asvrada:


utils/strings.c
```diff
@@ -14,6 +15,15 @@
 #define USE_BUILTIN_MEMCPY
 #define USE_LOOP_UNROLLING
 
+static void _check(bool overflowed)
+{
+    if (overflowed)
+    {
+        fprintf(stderr, "operation overflow\n");
+        oe_abort();
```
radhikaj, line 23: Is this used in other places in Mystikos? If it is consistent with current code, I am fine wiht it.

asvrada: No...  The only other usage of oe_abort is in https://github.com/deislabs/mystikos/blob/2be58c378c38d15be25c96e2708c56a9dbaaef5a/tools/myst/enc/clock.c#L95

This piece of code (strings.c) is used in many places, some in host some in enclave, I am not sure if oe_abort would work outside of enclave.

radhikaj (approved):

asvrada:
Close PR as the necessary checks are not automatically invoked.

PR #1369, opened 2022-06-20
Check and exit on overflow

asvrada:


asvrada:
Why is pipeline not running on PR?

PR #1372, opened 2022-06-22
Enable CodeQL manual run and change auto run to 9am local time

asvrada:
1. Add ability to manually trigger CodeQL workflow
2. Change auto run time from 5pm (00:00 UTC time) to 9am Los Angles time zone (16:00 UTC time)


radhikaj (approved):
Thanks @asvrada

Issue #1331, opened 2022-05-18
Auto-mount documentation issues


jxyang:
1. The target directory inside the enclave has to be explicitly created within appdir
2. The target directory cannot be on tmpfs or ramfs such as `/var/run`

jupacaza:
To support this issue, here is an example dockerfile from my team that works around this issue by doing mkdir for the mount paths and avoids using /var/* as destination folder for the mount:

``` dockerfile
FROM mystikos.azurecr.io/mystikos-bionic:v0.8.0 AS build

WORKDIR /home
RUN apt-get update && apt-get install -y cryptsetup-bin

COPY tokenservice.tar ./
COPY config.tokenservice.json config.json

# The following commands imitate what myst-appbuilder does, but without the need of having docker installed.
RUN mkdir -p appdir-tokenservice/tmp && \
    tar xvf tokenservice.tar -C appdir-tokenservice/tmp && \
    find appdir-tokenservice/tmp -name layer.tar -exec sh -c 'tar xvf {} -C appdir-tokenservice' \;
RUN rm -rf appdir-tokenservice/tmp

# Create folder for mounts
RUN mkdir -p appdir-tokenservice/mnt/secrets
RUN mkdir -p appdir-tokenservice/mnt/mdsd

RUN myst mkext2 appdir-tokenservice rootfs

RUN myst fssig --roothash rootfs > roothash

RUN openssl genrsa -out private.pem -3 3072

RUN myst package-sgx --roothash=roothash private.pem config.json

FROM mystikos.azurecr.io/mystikos-bionic:v0.8.0 AS final
WORKDIR /home
COPY --from=build /home/myst/bin/dotnet /home/myst/bin/dotnet
COPY --from=build /home/rootfs /home/rootfs
ENV MYST_ROOTFS_PATH=/home/rootfs

# --host-to-enc-uid-map parameter maps the host user 0 to enclave user 0 (root to root)
# --host-to-enc-gid-map parameter maps the host group 0 to enclave user 0 (root to root)
CMD [ "/home/myst/bin/dotnet", "--mount", "/mnt/secrets=/mnt/secrets", \
"--mount", "/mnt/mdsd=/mnt/mdsd", \
"--host-to-enc-uid-map", "0:0", \
"--host-to-enc-gid-map", "0:0,106:106" ]
```

My opinion:
[...]

PR #1378, opened 2022-06-27
Create directory for mount target provided in configuration file

asvrada:
When using config.json to specify mount information, create target folders if they do not exist.


asvrada:
Addresses https://github.com/deislabs/mystikos/issues/1331

kernel/enter.c
```diff
@@ -71,6 +72,67 @@ long myst_tcall(long n, long params[6])
...
+    // check if target exists and is directory
+    // if not, create target directory
+    {
+        struct stat buf;
+        int errornum =
+            (*parent_fs->fs_stat)(parent_fs, locals->suffix.buf, &buf);
+
+        if (errornum == -ENOENT)
+            // target do not exist, create directory
+            ECHECK(myst_syscall_mkdir(locals->suffix.buf, 0777));
```
vtikoo, line 117: Isn't `locals->suffix` a path relative to where the filesystem to which target belongs is mounted?

vtikoo: For the case where target belongs to rootfs it doesn't make a difference. But if the target(say `/mnt/target`) is on a filesystem mounted at some path(say `/mnt`). Then `myst_mount_resolve` will return just `/target` as the suffix.

Why don't you directly use `normalized_target` here?

PR #1381, opened 2022-07-06
mman synchronization fixes

vtikoo:
This PR attempts to fix two concurrency related issues in the memory manager. 

#### 1. Race condition:

Mystikos memory manager is a multi-tiered subsystem - syscall, mman wrapper(mmanutils.c), core mman(mman.c). Different layers update different data structures. The atomicity for these updations should span the layers.

Two threads can interleave in a way which can lead to inconsistent view of memory allocated and the process owning it.

Resources/data structures updated at the core mman layer:
- Active Virtual Address Descriptors(VAD) doubly linked list.
- Free VADs singly linked list.
- Unused VADs: i.e section of VAD array between mman->next_vad and mman->end_vad.
- Vector storing page protection information.

Resources/data structures updated at the mman wrapper layer:
- Vector storing file mapping information.
- Shared memory list.

Resources/data structures updated at the syscall layer:
- Vector storing process ownership information.

<ins>Proposed solution:</ins>
Move the syscall handler for affected memory syscalls under the critical section, with the recursive mman lock. Even the memory ownership test based on pids vector happens atomically with the rest of syscall. Another thread can munmap the mapping between the ownership check passing and the critical section proper of the routine.

Merge shared memory lock with the recursive mman lock. 

#### 2. Potential deadlock between mman lock and filesystem lock lockfs:

Both mman and in-kernel filesystems - i.e ramfs and ext2fs, have code paths which can cause a thread holding one lock to try acquire the other lock, leading to a deadlock.
Note - hostfs does not have this issue, as it doesn't use lockfs.

Have Lockfs want mman lock:
ext2fs and ramfs file systems are wrapped with a lockfs wrapper, which provides a file-system wide lock. Which means that at any given time atmost one thread will be in the file system routines.
Implementations of both ext2fs and ramfs can malloc/free memory, which could in turn cause a mmap() or munmap().

Have mman lock want lockfs:

Flows in the memory manager which can cause file operations:

1. mmap of file mappings.
2. mremap: shrinking  of a file mapping.
3. munmap of a MAP_SHARED file mapping.
4. process shutdown: releasing unreleased MAP_SHARED file mappings.
5. /proc/[pid]/maps callback: this uses fs_realpath to get pathname.

<ins>Proposed solution:</ins>
Generalize the lockfs lock even more.
Currently it is at the granularity of a filesystem instance. Proposal is to make a single global lock for all ramfs and ext2fs instances.
Then, above flows can acquire the global filesystem lock before acquiring the recursive mman lock. This part is an application of a deadlock avoidance mechanism - Banker's algorithm.

Checking for the subcases in 2.-5. requires parsing either file mapping vector for the range or shared memory list. We avoid this and unconditionally acquire the filesystem lock in these cases.
For 1.) filesystem lock is acquired only if a non-negative file descriptor has been passed(proxy for file mapping).
For 2.) we can easily check for shrinking, but not for file mapping.

During implementation, another circular wait situation emerged between the lockfs lock and the fdtable lock. This happened during processing a SYS_mmap file mapping request. This is the only flow in the memory manager where a thread holding the lockfs lock requests for the fdtable lock. Banker's algorithm was applied here too, i.e in `_SYS_mmap` both lockfs and fdtable locks are acquired if its a file mapping.


vtikoo:
Nightly manual run [here](https://openenclaveci.westus.cloudapp.azure.com/blue/organizations/jenkins/Mystikos%2FNightly-Pipeline-Manual/detail/Nightly-Pipeline-Manual/45/pipeline).

include/myst/rspinlock.h
```diff
@@ -53,8 +52,10 @@ MYST_INLINE void myst_rspin_lock(myst_rspinlock_t* s)
 
 MYST_INLINE void myst_rspin_unlock(myst_rspinlock_t* s)
 {
-    assert(s->count > 0);
-    assert(s->owner == __myst_rspin_self());
+    assert("rspin unlock with bad count" && s->count > 0);
```
mikbras, line 55: Clever way to get assert to print out a message. :)

include/myst/sharedmem.h
```diff
@@ -53,9 +53,16 @@ void myst_shmem_mremap_update(
 
 bool myst_shmem_can_mprotect(shared_mapping_t* sm, void* addr, size_t length);
 
-bool myst_addr_within_process_owned_shmem(
+/*
+Return value:
+0 - if no shared memory region corresponding to entire range.
```
mikbras, line 58: Which return value is the error? The Linux convention is that return values less than zero are errors.

kernel/lockfs.c
```diff
@@ -12,11 +12,14 @@
 
 #define LOCKFS_MAGIC 0x94639c1a101f4a1d
 
+myst_mutex_t _global_lock;
+
+static bool _inited;
```
mikbras, line 17: Maybe ``_initialized``?

kernel/lockfs.c
```diff
@@ -12,11 +12,14 @@
 
 #define LOCKFS_MAGIC 0x94639c1a101f4a1d
 
+myst_mutex_t _global_lock;
```
mikbras, line 15: I will need help understanding why a global lock is necessary?

kernel/lockfs.c
```diff
@@ -713,6 +731,12 @@ int myst_lockfs_init(myst_fs_t* fs, myst_fs_t** lockfs_out)
     lockfs->fs = fs;
     *lockfs_out = &lockfs->base;
 
+    if (!_inited)
```
mikbras, line 734: Statically defined myst mutexes don't need to be initialized (since statics are zero-initialized). We should probably add a "MYST_MUTEX_INITIALIZER" macro for clarity.

mikbras:
```
#define MYST_MUTEX_INITIALIZER { MYST_SPINLOCK_INITIALIZER }
```

```
static myst_mutex_t _mutex = MYST_MUTEX_INITIALIZER ;
```

kernel/mmanutils.c
```diff
@@ -695,6 +729,16 @@ long myst_mmap(
 
         ret = (long)addr;
     }
+    // Cases:
+    //    Anonymous mapping(MAP_ANON) && addr hint: only page protection
+    //    updation in myst_mman_mmap epilogue.
```
mikbras, line 734: /updation/update/updating

kernel/mmanutils.c
```diff
@@ -695,6 +729,16 @@ long myst_mmap(
...
+    //
+    //    No addr hint: allocation, VADS list and page prot vector updation.
```
mikbras, line 736: updation -> update

kernel/mmanutils.c
```diff
@@ -746,6 +790,10 @@ long myst_mmap(
     assert(end >= _mman_start && end <= _mman_end);
 
 done:
+
+    _runlock(&locked);
+    if (fd >= 0)
```
mikbras, line 795: remove this line

kernel/mmanutils.c
```diff
@@ -756,44 +804,65 @@ void* myst_mremap(
     int flags,
     void* new_address)
 {
+    long ret = 0;
     void* p;
-    int r;
     shared_mapping_t* shm_mapping;
```
mikbras, line 809: Should this be null initialized.

kernel/mmanutils.c
```diff
@@ -809,28 +878,42 @@ int myst_mprotect(const void* addr, const size_t len, const int prot)
...
-            MYST_WLOG("Unsupported mprotect operation detected. For shared "
-                      "mappings, mprotect is only allowed if there is a single "
-                      "user of the mapping. Mystikos relies on host for page "
-                      "protection. On x86-64 and Linux, page protection is per "
-                      "address space. As we "
-                      "are single process on the host, supporting mprotect for "
-                      "memory shared between two process threads becomes "
-                      "difficult.\n");
-            return -EINVAL;
+            if (!myst_shmem_can_mprotect(shm_mapping, (void*)addr, len))
```
mikbras, line 888: Bookmark

kernel/mmanutils.c
```diff
@@ -859,13 +941,21 @@ int myst_munmap(void* addr, size_t length)
 {
     int ret = 0;
     fdlist_t* head = NULL;
+    bool locked = false;
 
+    /* File system lock is acquired in the syscall handler. dlmalloc calls to
+     * myst_mmap/munmap already acquire the lock. Acquiring the file system lock
+     * here can lead to deadlock. The assumption is that in-kernel usages never
+     * do file mappings. */
+    _rlock(&locked);
```
mikbras, line 950: probably redundant lock

kernel/syscall.c
```diff
@@ -3731,31 +3756,147 @@ static long _SYS_munmap(
...
+    myst_lockfs_unlock();
+
     return (_return(n, ret));
 }
 
+static long _SYS_mremap(long n, long params[6])
+{
+    void* old_address = (void*)params[0];
+    size_t old_size = (size_t)params[1];
+    size_t new_size = (size_t)params[2];
```
mikbras, line 3799: If ``new_size`` is zero then return ``-EINVAL``. Check range of values for ``new_size`` and ``old_size``.

kernel/lockfs.c
vtikoo: Thanks this is really neat.

vtikoo:
Superseded by #1421

PR #1396, opened 2022-07-29
added kubernetes demo for conf_ml

salsal97:


samples/confidential_ml/confml_server.yaml
```diff
@@ -0,0 +1,59 @@
...
+        volumeMounts:
+        - name: var-run-aesmd
+          mountPath: /var/run/aesmd # hardcoded in intels libarary, volume name must match that of the daemonset 
+        ports:
+        - name: client1
+          containerPort: 8000
+        imagePullPolicy: Always
+        resources:
+          limits:
+            kubernetes.azure.com/sgx_epc_mem_in_MiB: "100"
```
vtikoo, line 52: It might be better to have ICX defaults rather than CFL.

vtikoo (approved):
Thanks for extending this sample to AKS, great work!

samples/confidential_ml/confml_server.yaml
salsal97: In the demo we gave them both 100MiB

salsal97: Bumped it to 4GiB to match the config.json value as we discussed

PR #1402, opened 2022-08-16
Print stacktrace on SIGSEGV exceptions

vtikoo:
Currently this only prints functions names for exceptions originated in the kernel. Supporting application symbols would require the kernel to store some symbol information about loaded modules.

Sample output from a forced kernel crash -
```
*** Kernel segmentation fault 
0x1007c0448: _fs_read()
0x1007b3904: _fs_read()
0x1007f7080: myst_syscall_read()
0x1007fe9c8: _SYS_read()
0x1008059f8: _syscall()
0x10082b4e2: __morestack()
```

### DIFF
--- a/include/myst/backtrace.h
+++ b/include/myst/backtrace.h
@@ -7,6 +7,8 @@
 #include <stdbool.h>
 #include <stddef.h>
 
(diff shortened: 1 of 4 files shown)


mingweishih (approved):
LGTM

Issue #946, opened 2021-10-27
Memory leak in myst_syscall_execve


vtikoo:
Running tests/vfork with memcheck turned on reports a memory leak within `myst_syscall_execve`. Its looks like the `argv` array is not being released properly.
```
16 bytes
0x1007a8f9c: myst_debug_malloc()
0x1007a90d1: myst_debug_calloc()
0x1007c926e: calloc()
0x1007d44f8: myst_syscall_execve()
0x1007d919e: _syscall()
0x1007fefb3: __morestack()
```

isil-oz:
@vtikoo : Isn't `argv` being released inside `myst_exec`, which is called from `myst_syscall_execve` (`myst_syscall_execveat` in the current version)? 
1237: `if (callback)`
1238:        `(*callback)(callback_arg);`
But I think that it needs to be after `done` label (at line 1253) to guarantee the call in case of any error.

vtikoo:
@isil-oz Thanks for root causing this! That makes sense to me. 

As you alluded we are using the `callback` mechanism for freeing up `argv`. We definitely need to call this callback before jumping to CRT here -
    /* enter the C-runtime on the target thread descriptor */
    (*enter)(sp, dynv, myst_syscall, crt_args);
 As the rest of the function will never be executed in this case.

We could check after `done` if the callback wasn't called and call it.

vtikoo:
@isil-oz Oops looks like this was already fixed in commit [c22a322803157b5ca751fd9ee6005f71a16d6216](https://github.com/deislabs/mystikos/commit/c22a322803157b5ca751fd9ee6005f71a16d6216)

isil-oz:
@vtikoo The code has been modified since that fix. Can you check the current version (`myst_syscall_execveat`)? There was no problem with error-free executions. I had tried to point out that the callback wasn't called after `done`.

vtikoo:
Here is where `argv` is released in `myst_syscall_execveat` - https://github.com/deislabs/mystikos/blob/main/kernel/syscall.c#L2341-L2342.

`myst_exec` function only returns if there is an error. And so `myst_syscall_execveat` being the caller also has the same property.  By freeing argv after `done` in the caller, it doesn't matter the callback wasn't called in the lower level.

isil-oz:
Thanks for explaining, I see. If there is an error, `myst_exec` returns and `myst_syscall_execveat` releases `argv` in its `done`.

PR #1404, opened 2022-08-17
Fix issue #946

isil-oz:
Just observed that the callback wasn't called after done, and added a call to make sure argv is deallocated.


vtikoo:
Thanks for contributing, but looks like this was fixed since the issue was created.

PR #1412, opened 2022-08-27
added group filtering feature to strace

salsal97:
This PR adds group filtering to the strace filter options. The groups available are {"desc", "file", "ipc", "network", "process", "signal", "memory", "stat", "lstat", "fstat", "stat_like", "statfs", "fstatfs", "statfs_like", "pure", "syscall_never_fails", "max_args", "memory_mapping_change", "stackcapture_on_enter", "compat_syscall_types", "seccomp_default", "creds", "clock", "comm_change"}. 

The table specifying the syscall groups mimics the strace source, a combination of these two files - https://gitlab.com/strace/strace/-/blob/master/src/linux/x86_64/syscallent.h & https://gitlab.com/strace/strace/-/blob/master/src/sysent_shorthand_defs.h


anakrish (approved):
At a high-level, looks good to me.

utils/syscall.c
```diff
@@ -382,9 +382,342 @@ static const myst_syscall_pair_t _pairs[] = {
...
+      SYS_mprotect,
+      SYS_munmap,
+      SYS_brk,
+      SYS_mremap,
+      SYS_shmat,
+      SYS_execve,
+      SYS_shmdt,
+      SYS_remap_file_pages,
+      SYS_execveat,
+      SYS_pkey_mprotect}},
```
vtikoo, line 681: We don't support SYS_shmat,SYS_shmdt, SYS_remap_file_pages, SYS_pkey_mprotect.
Should we have entries for unsupported syscalls?

salsal97: I think it would be easier to keep this list forward thinking and in complete sync with the list in strace than go through all the groups again each time we add a syscall - what do you think?

include/myst/syscall.h
```diff
@@ -407,4 +407,21 @@ typedef struct myst_syscall_pair
...
+/* Stores the syscall group name, corresponding syscalls, and number of syscalls
+ */
+typedef struct myst_syscall_group
+{
+    const char* name;
+    const size_t group_size;
+    const int syscalls[SYSCALL_GROUP_MAX_SIZE];
+} myst_syscall_group_t;
+
+const myst_syscall_group_t* myst_syscall_groups(void);
```
vtikoo, line 421: There is an informal naming convention for interface functions of kernel subsytems - `myst_<subsytem-name>_<function-name>`.
```suggestion
const myst_syscall_group_t* myst_strace_syscall_groups(void);
```

utils/syscall.c
```diff
@@ -382,9 +382,342 @@ static const myst_syscall_pair_t _pairs[] = {
     PAIR(SYS_myst_interrupt_thread),
     PAIR(SYS_myst_pre_launch_hook),
     /* add new entries here! */
+    /* NOTE: when you add a new syscall here, add it to its corresponding group
+       below as well! */
     {0, NULL},
 };
 
+static const myst_syscall_group_t _groups[] = {
```
vtikoo, line 390:
```suggestion
const myst_syscall_group_t _groups[] = {
```
`_groups[]` already has global scope, I don't think you need static here. Most uses of C statics I've seen are for having something like a function level global variable.

vtikoo: Also consider moving strace related structures and definitions to a  `strace.h` header file.

tools/myst/host/strace.h
```diff
@@ -11,4 +11,10 @@ int myst_parse_strace_config(
     const char** argv,
     myst_strace_config_t* strace_config);
 
+int myst_add_syscall_to_filter(
```
vtikoo, line 14:
```suggestion
int myst_strace_add_syscall_to_filter(
```

salsal97: Added this

utils/syscall.c
salsal97: Makes sense, I removed the static keyword there

include/myst/syscall.h
salsal97: I went by the convention for the pairs structure defined in this file for this one - the structure should be a general kernal differentiation rather than something just for strace right? We also define it in syscall.c and not strace.h

utils/syscall.c
```diff
@@ -382,9 +382,342 @@ static const myst_syscall_pair_t _pairs[] = {
...
     PAIR(SYS_myst_pre_launch_hook),
     /* add new entries here! */
+    /* NOTE: when you add a new syscall here, add it to its corresponding group
+       below as well! */
     {0, NULL},
 };
 
+const myst_syscall_group_t _groups[] = {
+    {"desc",
+     100,
```
asvrada, line 392: There seems to be 98 items, excluding the 2 commented out

utils/syscall.c
```diff
@@ -382,9 +382,342 @@ static const myst_syscall_pair_t _pairs[] = {
...
+      SYS_kill,
+      SYS_rt_sigqueueinfo,
+      SYS_tkill,
+      SYS_exit_group,
+      SYS_tgkill,
+      SYS_waitid,
+      SYS_rt_tgsigqueueinfo,
+      SYS_execveat}},
+    {"signal",
+     15,
```
asvrada, line 593: Should be 14, excluding the commented out line

utils/syscall.c
```diff
@@ -382,9 +382,342 @@ static const myst_syscall_pair_t _pairs[] = {
...
+      SYS_pkey_mprotect}},
+    {"stat", 1, {SYS_stat}},
+    {"lstat", 1, {SYS_lstat}},
+    {"fstat", 3, {SYS_fstat, SYS_newfstatat, SYS_statx}},
+    {"stat_like",
+     5,
+     {SYS_stat, SYS_fstat, SYS_lstat, SYS_newfstatat, SYS_statx}},
+    {"statfs", 1, {SYS_statfs}},
+    {"fstatfs", 1, {SYS_fstatfs}},
+    {"statfs_like", 3, {SYS_ustat, SYS_statfs, SYS_fstatfs}},
```
asvrada, line 643: I wonder how do we decide which group a syscall goes into?

utils/syscall.c
```diff
@@ -382,9 +382,342 @@ static const myst_syscall_pair_t _pairs[] = {
...
+      SYS_getresuid,
+      SYS_setresgid,
+      SYS_getresgid,
+      SYS_setfsuid,
+      SYS_setfsgid,
+      SYS_capget,
+      SYS_capset,
+      SYS_prctl}},
+    {"clock",
+     8,
```
asvrada, line 709: Should this be 7?

utils/syscall.c
```diff
@@ -382,9 +382,342 @@ static const myst_syscall_pair_t _pairs[] = {
...
+     {SYS_gettimeofday,
+      SYS_adjtimex,
+      SYS_settimeofday,
+      SYS_time,
+      SYS_clock_settime,
+      SYS_clock_gettime,
+      //   SYS_lock_getres,
+      SYS_clock_adjtime}},
+    {"comm_change", 3, {SYS_execve, SYS_prctl, SYS_execveat}},
```
asvrada, line 718: Redundant `,` at the end?

utils/syscall.c
```diff
@@ -440,6 +773,30 @@ long myst_syscall_num(const char* name)
     return -ENOENT;
 }
 
+const int* myst_syscall_group(const char* name)
+{
+    for (size_t i = 0; _groups[i].name; i++)
```
asvrada, line 778: Would `i` goes out of bound? There is no NULL element at the end of  `_groups`

include/myst/syscall.h
vtikoo: Even `_pairs` is used just by the tracing subsystem.

vtikoo (approved):
Looks good! Excited to try this out.

utils/syscall.c
salsal97: Oh no thank you for pointing that out, that would be a bug!

utils/syscall.c
salsal97: I took this information from the strace source code, check the PR description above for details

doc/user-getting-started.md
```diff
@@ -90,18 +90,27 @@ Options:
...
+                            specified as described above. Any combination of syscalls and groups 
+                            can be used. For a list of all the groups and their consituents check
+                            out 'man strace' 
+                            E.g: To trace open and mprotect syscalls, and 'desc' group of 
+                            syscalls (file descriptor related group), specify 
+                            --strace-filter 'SYS_open:SYS_mprotect:desc'
+    --strace-exclude-filter 'SYS_name1:SYS_name2:group1...'
+                         -- Specify a set of syscalls or groups to exclude from the strace log. 
+                            All other syscalls will be logged in the strace. When filters 
+                            are specified, only those syscalls/groups specified in the filter
```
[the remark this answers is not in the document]
salsal97: I've incorporated that, thanks for the easier wording.

doc/user-getting-started.md
```diff
@@ -90,18 +90,27 @@ Options:
                             to stop execution whenever a syscall fails. Use breakpoint 
                             conditions to control the behavior of the breakpoint.
                             E.g: Use syscall number as a condition in the breakpoint
-    --strace-filter 'SYS_name1:SYS_name2:...'
-                         -- Specify the set of syscalls to be traced. When filters 
-                            are specified, only those syscalls specified in the filter 
+    --strace-filter 'SYS_name1:group1:SYS_name2:...'
+                         -- Specify the set of syscalls or group to be traced. When filters
```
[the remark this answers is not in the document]
salsal97: Typo, thanks

utils/syscall.c
salsal97: Changed this, thanks

utils/syscall.c
salsal97: Yep, changed

utils/syscall.c
salsal97: Added the null element

utils/syscall.c
salsal97: I think this is necessary in this case

utils/syscall.c
```diff
@@ -382,9 +382,343 @@ static const myst_syscall_pair_t _pairs[] = {
...
+     {SYS_gettimeofday,
+      SYS_adjtimex,
+      SYS_settimeofday,
+      SYS_time,
+      SYS_clock_settime,
+      SYS_clock_gettime,
+      //   SYS_lock_getres,
+      SYS_clock_adjtime}},
+    {"comm_change", 3, {SYS_execve, SYS_prctl, SYS_execveat}},
+    {"", 0, {}},
```
asvrada, line 719: Should this be `{NULL, 0, {}}`?

asvrada (approved):
LGTM with comments

utils/syscall.c
salsal97: So I had a test for this scenario that passed (even without the NULL element) which is why I didn't catch it earlier. See test 8 in the tests section.

salsal97: It seems to work without needing the NULL element at the end, but since this is more correct I'll add it

PR #1414, opened 2022-09-01
added pid & tid filtering feature

salsal97:
This PR adds the strace filter that let's you filter by thread ID or process ID (upto 10 at a time)
 to make your output more relevant


anakrish (approved):

tools/myst/host/exec.c
```diff
@@ -455,13 +460,16 @@ int exec_action(int argc, const char* argv[], const char* envp[])
             cli_getopt(&argc, argv, "--strace", NULL) == 0)
         {
             options.strace_config.trace_syscalls = true;
+            // fprintf(stderr, "Failed to  final options\n");
```
asvrada: Should we remove commented lines?

tools/myst/host/strace.c
```diff
@@ -144,6 +145,56 @@ int myst_strace_parse_config(
...
+                fprintf(
+                    stderr,
+                    "Tid %s is invalid. Must be a positive number.\n",
+                    _tid);
+                abort();
+            }
+
+            strace_config->tid_trace[i] = tid;
+
+            if (i >= MYST_MAX_IDS)
```
asvrada: Should the `if` clause comes before `strace_config->tid_trace[i] = tid;`?

asvrada: Or give warning before entering the for loop

kernel/syscall.c
```diff
@@ -196,17 +196,55 @@ static bool _trace_syscall(long n)
     // Check if syscall tracing is enabled.
     if (__myst_kernel_args.strace_config.trace_syscalls)
     {
+        // -1 represents strace syscall filter is not set
+        short filter_syscall = -1;
```
asvrada, line 200: Can we use `bool` here? Also personally I'd prefer naming it `syscall_allowed` or `allow_syscall`

tools/myst/host/strace.c
```diff
@@ -91,13 +91,73 @@ int myst_set_strace_filter(
     return 0;
 }
 
+int myst_set_strace_id_filter(
+    const char* id_filter,
+    myst_strace_config_t* strace_config,
+    bool tid)
```
asvrada, line 97: Could you add a bit comments here to explain if `tid` is true, set filter for tid, else its for pid?

tools/myst/host/strace.c
```diff
@@ -91,13 +91,73 @@ int myst_set_strace_filter(
...
+                _id);
+            abort();
+        }
+
+        if (tid)
+            strace_config->tid_trace[i] = id;
+        else
+            strace_config->pid_trace[i] = id;
+
+        if (i >= MYST_MAX_IDS)
```
asvrada, line 134: Should we validate index first

kernel/syscall.c
```diff
@@ -196,17 +196,55 @@ static bool _trace_syscall(long n)
...
+
         if (__myst_kernel_args.strace_config.filter)
         {
-            // If filtering is enabled, trace only this syscall has been
+            // If filtering is enabled, trace only this syscall if it has been
             // specified in the filter.
-            return __myst_kernel_args.strace_config.trace[n];
+            filter_syscall = __myst_kernel_args.strace_config.trace[n];
```
asvrada, line 206: Can we declare `filter_syscall ` here

`bool filter_syscall = __myst_kernel_args.strace_config.trace[n];`

kernel/syscall.c
```diff
@@ -196,17 +196,55 @@ static bool _trace_syscall(long n)
...
+                return false;
         }
-        else
+        if (__myst_kernel_args.strace_config.tid_filter_num > 0)
         {
-            // Trace all syscalls.
-            return true;
+            // Check if thread filtering is enabled & follow filter
+            myst_thread_t* self = myst_thread_self();
+            pid_t current_tid = self->tid;
```
asvrada, line 215: Not should if its a good idea to `(long)self->tid` here, we don't have to cast to long in every loop. Sounds like over optimization

tools/myst/host/strace.c
```diff
@@ -91,13 +91,73 @@ int myst_set_strace_filter(
...
+            stderr,
+            "Invalid strace-filter-tid or strace-filter-pid option '%s' "
+            "specified.\n",
+            id_filter);
+        abort();
+    }
+
+    for (size_t i = 0; i < num_tokens; ++i)
+    {
+        long id;
```
asvrada, line 115: Not sure if we should init id to some placeholder value (like -1?) here. Looks like there is a chance that this variable is not initialized, when `myst_str2long` errors out

tools/myst/host/strace.c
```diff
@@ -91,13 +91,73 @@ int myst_set_strace_filter(
...
+        const char* _id = tokens[i];
+
+        ret = myst_str2long(_id, &id);
```
asvrada, line 118: Should we `ECHECK` here?

tools/myst/host/exec.c
salsal97: this isnt a part of the final PR, not sure how you see it here

kernel/syscall.c
salsal97: Cannot use bool here. There are three options - unset (-1), and set to 0 or 1. I tried to use the minimum required, but we need all three options.

salsal97: I changed the logic, we might not need the flag anymore. Thanks for making me look at it again.

kernel/syscall.c
salsal97: removed the variable and changed the logic

kernel/syscall.c
salsal97: Changed the data structure so we don't need long's at all

tools/myst/host/strace.c
salsal97: Sure, added

tools/myst/host/strace.c
salsal97: Good idea, added

tools/myst/host/strace.c
salsal97: ECHECK was ignoring the error and causing a bug, so I replaced it with this method

tools/myst/host/strace.c
salsal97: Yes, added this

tools/myst/host/strace.c
salsal97: Added this!

kernel/syscall.c
```diff
@@ -193,20 +193,53 @@ static const char* _format_timespec(
...
+            }
+
+            /* tid filter is specified, but no match in filter */
+            return false;
         }
+        if (__myst_kernel_args.strace_config.pid_filter_num > 0)
+        {
+            /* Check if thread filtering is enabled & follow filter */
+            myst_process_t* self = myst_process_self();
+            pid_t current_pid = self->pid;
```
vtikoo, line 227:
```suggestion
            pid_t current_pid = myst_process_self()->pid;
```
It doesn't look like `self` is used anywhere else.

tools/myst/host/strace.c
```diff
@@ -88,18 +99,84 @@ int myst_set_strace_filter(
...
     strace_config->filter = 1;
+
+    if (tokens)
+        free(tokens);
+
     return 0;
 }
 
+/* sets tid filter if 'tid' is true, else sets pid filter */
+int myst_set_strace_id_filter(
```
vtikoo, line 110: Consider `myst_set_trace_pid_tid_filter`, makes it easier to get what the function does in the callsite.

tools/myst/host/strace.c
```diff
@@ -41,11 +41,22 @@ int myst_strace_add_syscall_to_filter(
...
-    int num_tokens,
-    char** tokens,
+    const char* filter,
     myst_strace_config_t* strace_config,
     bool include)
 {
+    char** tokens = NULL;
+    size_t num_tokens = 0;
+
+    if (myst_strsplit(filter, ":", &tokens, &num_tokens) != 0)
```
vtikoo, line 51: 👍 Good catch, this definitely belonged here than in `myst_strace_parse_config`.

vtikoo:
As syscall entry and exit hooks are hot code paths, thinking about performance impact.

Checking whether current thread/process is whitelisted is a linear scan of array of ints. In the worst case this is a scan of 100 integers, which would span atleast 6 cache lines.
The syscall include and exclude filters are single word lookups. Checking whether the syscall number `snum` is in either list, requires only looking at trace[snum]. 

An alternative is encoding the tid filter as a bit vector. The max number of threads we support is 1024, so that would be a bit vector atleast 128 bytes long. We could potentially also leverage the fact that our tid/pid generation algorithm starts from 101. So tid 101 could be the 1st bit of the bit vector and so on.

Having said that, if memory is not the bottleneck then this optimization might not be as useful.

tools/myst/host/strace.c
salsal97: I agree, will make the change

kernel/syscall.c
salsal97: I did this on purpose to keep the statements atomic, but I guess the application scope for that is lesser here. Will change.

PR #1415, opened 2022-09-08
strace filter optimizations

salsal97:
This PR changes the format in which we store the syscall groups so that we use only as much space as we need and do not initialize all group arrays with 128 int size.


asvrada (approved):
LGTM

PR #1416, opened 2022-09-12
Fix execve bug

asvrada:


asvrada:
Had some misunderstanding of execveat, close as the original code is correct.

PR #1418, opened 2022-09-13
Add thread tracking extension to myst-gdb

asvrada:
This extension tracks all threads created inside Enclave, and report their pid, tid to facilitate debugging.
This extension also tracks all command-line arguments that a process runs.

Example usage: tests/dotnet-lib-5

`make gdb-runner`

```
>>> myst-thread

        myst-thread: Mystikos thread tracker
        Commands:
        myst-thread help
          Print this message

        myst-thread thread
          List all tracked threads
        
        myst-thread cmdline
          List the command line for each process
        
        myst-thread exe
          List the executable for each process

        myst-thread cwd
          List the current working directory for each process
        
>>> myst-thread thread
List of Mystikos controlled threads:
ppid pid tid(target_tid)
100 101 101(5666)
100 101 102(5730)
100 101 103(5731)
100 101 104(5732)
100 101 105(5733)
100 101 106(5734)
100 101 107(5736)
100 101 108(5738)
100 101 109(5742)
100 101 110(5746)
100 101 111(5747)
100 101 112(5748)
100 101 113(5749)
100 101 114(5750)
100 101 117(5766)
100 101 118(5767)
100 101 119(5771)
100 101 120(5772)
100 101 121(5774)
100 101 122(5775)
>>> myst-thread cmdline
Process 101 (Parent 100) cmdline: /runner/bin/Release/net5.0/runner /testcases/pass.1 /dotnet-lib-release/
```

### DIFF
--- a/debugger/gdb-sgx-plugin/thread.py
+++ b/debugger/gdb-sgx-plugin/thread.py
@@ -1,0 +1,290 @@
+import gdb
+import math
+import tempfile
+from typing import List
+
+DEBUG = False
+
+INSTANCE_DEBUG_EXT = None
+
+class ThreadCreateBreakpoint(gdb.Breakpoint):
+    
+    STR_FIRST_ARG = "((struct run_thread_arg*)$rdi)"
+
+    def __init__(self):
+        """
+        Set a breakpoing @ kernel/thread.c:_run_thread, this should be right before myst_setjmp(&thread->jmpbuf)
+        Choosing this location because this is where the newly created thread about to run
+        We can get the new thread pid/tid here
+        """
+        super(ThreadCreateBreakpoint, self).__init__('kernel/thread.c:_run_thread', internal=False)
+
+    def stop(self):
+        # New thread info
+        ppid = int(gdb.parse_and_eval(self.STR_FIRST_ARG + '->thread->process->ppid'))
+        pid = int(gdb.parse_and_eval(self.STR_FIRST_ARG + '->thread->process->pid'))
+        tid = int(gdb.parse_and_eval(self.STR_FIRST_ARG + '->thread->tid'))
+        tid_target = int(gdb.parse_and_eval(self.STR_FIRST_ARG + '->target_tid'))
+        if DEBUG:
+            print("Created thread: ppid, pid, tid, tid_target", ppid, pid, tid, tid_target)
+        INSTANCE_DEBUG_EXT.add_thread(pid, tid, ppid, tid_target)
+
+        # Add _main_thread only once as it won't change often
+        if not INSTANCE_DEBUG_EXT.main_thread_tracked():
+            main_ppid = int(gdb.parse_and_eval('_main_thread->process->ppid'))
+            main_pid = int(gdb.parse_and_eval('_main_thread->process->pid'))
+            main_tid = int(gdb.parse_and_eval('_main_thread->tid'))
+            main_tid_target = int(gdb.parse_and_eval('_main_thread->target_tid'))
+            if DEBUG:
+                print("Main thread: ppid, pid, tid, tid_target", main_ppid, main_pid, main_tid, main_tid_target)
+            INSTANCE_DEBUG_EXT.add_thread(main_pid, main_tid, main_ppid, main_tid_target)
+
+        INSTANCE_DEBUG_EXT.copy_parent_process(pid, ppid)
+        return False
+
+class ThreadExitBreakpoint(gdb.Breakpoint):
+
+    STR_FIRST_ARG = "((const myst_thread_t*)$rdi)"
+
+    def __init__(self):
+        """
+        Set a breakpoint @ kernel/thread.c:myst_debug_hook_thread_exit, where myst_setjmp returns non-zero
+        Choosing this location because this is where the thread terminates
+        """
+        super(ThreadExitBreakpoint, self).__init__('kernel/thread.c:myst_debug_hook_thread_exit', internal=False)
+
+    def stop(self):
+        # Get thread info that is about to terminate
+        ppid = int(gdb.parse_and_eval(self.STR_FIRST_ARG + '->process->ppid'))
+        pid = int(gdb.parse_and_eval(self.STR_FIRST_ARG + '->process->pid'))
+        tid = int(gdb.parse_and_eval(self.STR_FIRST_ARG + '->tid'))
+        tid_target = int(gdb.parse_and_eval(self.STR_FIRST_ARG + '->target_tid'))
+        if DEBUG:
+            print("Exit thread: ppid, pid, tid, tid_target", ppid, pid, tid, tid_target)
+
+        INSTANCE_DEBUG_EXT.remove_thread(tid)
+        return False
+
+class MystExecBreakpoint(gdb.Breakpoint):
+
+    STR_FIRST_ARG = "((const myst_thread_t*)$rdi)"
+    STR_THIRD_ARG = "((const char**)$rdx)"
+
+    def __init__(self):
+        """
+        Obtain program's name, args and cwd to use in enclave
+        """
+        super(MystExecBreakpoint, self).__init__('kernel/exec.c:myst_debug_hook_myst_exec', internal=False)
+
+    def parse_string_raw(self, str_raw):
+        """
+        Parse the string given the gdb parse_and_eval result
+
+        Example input:
+        0x104df31f0 "/bin/usr/some/path"
+
+        Example return:
+        /bin/usr/some/path
+        """
+        val = str(str_raw)
+        if ' ' not in val:
+            return val
+        
+        # Don't include "
+        return val[val.index(' ') + 2:-1]
+
+    def stop(self):
+        # Same as parse and eval *argv@argc
+        argc = int(gdb.parse_and_eval('$rsi'))
+
+        # ((const char**)$rdx)[0..1..2..]
+        argv_raw = [gdb.parse_and_eval('{0}[{1}]'.format(self.STR_THIRD_ARG, i)) for i in range(argc)]
+        cwd_raw = gdb.parse_and_eval(self.STR_FIRST_ARG + '->process->cwd')
+
+        argv = [self.parse_string_raw(each) for each in argv_raw]
+        cwd = self.parse_string_raw(cwd_raw)
+
+        # Get pid and ppid
+        ppid = int(gdb.parse_and_eval(self.STR_FIRST_ARG + '->process->ppid'))
+        pid = int(gdb.parse_and_eval(self.STR_FIRST_ARG + '->process->pid'))
+
+        INSTANCE_DEBUG_EXT.set_process_cmd_cwd(pid, ppid, argv, cwd)
+
+        return False
+
+class DebugExtension:
+    """
+    Manage all breakpoints and program info(process/thread/cmdline/etc)
+    """
+
+    # Configure the default pid/tid for the main process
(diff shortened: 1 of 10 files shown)


asvrada:
To improve: when a process calls `fork()`, the command `myst-debug cmdline` won't list the child process, unless that process calls `execve` family APIs

debugger/gdb-sgx-plugin/debug_ext.py
```diff
@@ -0,0 +1,306 @@
...
+INSTANCE_DEBUG_EXT = None
+
+class ThreadCloneBreakpoint(gdb.Breakpoint):
+    def __init__(self):
+        """
+        Set a breakpoint @ kernel/syscall.c::_SYS_myst_clone
+        Choosing this location because this is where the parent process/thread creates new process/thread
+        We can get the parent pid/tid from here
+        """
+        super(ThreadCloneBreakpoint, self).__init__('kernel/syscall.c:4094', internal=False)
```
vtikoo, line 17: Line number breakpoints are brittle. Is it possible to choose a function name instead?

debugger/gdb-sgx-plugin/debug_ext.py
```diff
@@ -0,0 +1,306 @@
...
+        self.map_commands = {
+            'help': lambda: self._command_help(),
+            'thread': lambda: self._command_list_thread(),
+            'cmdline': lambda: self._command_cmdline(),
+            'exe': lambda: self._command_exe(),
+            'cwd': lambda: self._command_cwd()
+        }
+
+        # Store thread info
+        # Key: tuple (pid, tid)
```
vtikoo, line 153: Do you need to store `pid` here?

vtikoo: `tid` is already an unique identifier.

debugger/gdb-sgx-plugin/debug_ext.py
```diff
@@ -0,0 +1,306 @@
...
+        """
+        Set a breakpoint @ kernel/syscall.c::_SYS_myst_clone
+        Choosing this location because this is where the parent process/thread creates new process/thread
+        We can get the parent pid/tid from here
+        """
+        super(ThreadCloneBreakpoint, self).__init__('kernel/syscall.c:4094', internal=False)
+
+    def stop(self):
+        # Get Parent thread info
+        ppid = int(gdb.parse_and_eval('myst_thread_self()->process->ppid'))
```
anakrish, line 21: It is better to avoid function calls in `gdb.parse_and_eval`. Executing enclave functions from within gdb is not guaranteed to work.

Instead, evaluate the function in code, assign it to a volatile local variable and then read that variable from within the debugger. Even better would be to read a register instead of a named variable.

debugger/gdb-sgx-plugin/debug_ext.py
anakrish, line 17: We need some way to ensure that the extension is not broken when the source file is edited and line numbers change.

debugger/gdb-sgx-plugin/debug_ext.py
asvrada: Store only tid now

debugger/gdb-sgx-plugin/debug_ext.py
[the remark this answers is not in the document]
asvrada: This breakpoint is not in use, but leaving the code here since it could be useful once we get around the function invocation

asvrada: We could make such modifications if we think the information from this breakpoint is important (i.e whos the parent thread/process)

debugger/gdb-sgx-plugin/debug_ext.py
asvrada: Changed 2 breakpoint to stop on method invocation

debugger/gdb-sgx-plugin/thread.py
```diff
@@ -0,0 +1,326 @@
...
+class ThreadCloneBreakpoint(gdb.Breakpoint):
+    def __init__(self):
+        """
+        THIS BREAKPOINT IS DISABLED AS FUNCTION CALL MAY NOT BE RELIABLE
+
+        Set a breakpoint @ kernel/syscall.c::_SYS_myst_clone
+        Choosing this location because this is where the parent process/thread creates new process/thread
+        We can get the parent pid/tid from here
+        """
+        super(ThreadCloneBreakpoint, self).__init__('kernel/syscall.c:4094', internal=False)
```
vtikoo, line 19: Assuming the breakpoint was supposed to be inside `_SYS_myst_clone`. Line number has already changed in `main`, 4094 is no longer within that function.
It would be better to avoid line number based breakpoints if possible. In this case `_SYS_myst_clone` seems like a wrapper around `myst_syscall_clone`, so you could possibly put a breakpoint on that.

debugger/gdb-sgx-plugin/thread.py
```diff
@@ -0,0 +1,326 @@
...
+        984
+   b -> 985     assert(myst_gettid() != -1);
+        986
+        987     /* restore the target thread descriptor */
+        988     myst_set_fsbase(thread->target_td);
+        989
+        990     /* ---------- running target thread descriptor ---------- */
+        991 ...
+        """
+        super(ThreadExitBreakpoint, self).__init__('kernel/thread.c:985', internal=False)
```
vtikoo, line 91: I would recommend adding debugger hooks here, instead of line number breakpoints. We use hooks for module load and unload too. See `oe_debug_notify_library_loaded` and `oe_debug_notify_library_unloaded` usage in tools/myst/host/host.c

debugger/gdb-sgx-plugin/thread.py
```diff
@@ -0,0 +1,326 @@
...
+
+        Set a breakpoint @ kernel/syscall.c::_SYS_myst_clone
+        Choosing this location because this is where the parent process/thread creates new process/thread
+        We can get the parent pid/tid from here
+        """
+        super(ThreadCloneBreakpoint, self).__init__('kernel/syscall.c:4094', internal=False)
+
+    def stop(self):
+        # Get Parent thread info
+        ppid = int(gdb.parse_and_eval('myst_thread_self()->process->ppid'))
```
anakrish, line 23: oegdb does not guarantee that calling enclave functions from the debugger would work.
Therefore, we cannot rely on calling `myst_thread_self()`.

With lldb, enclave function calls are not supported at all.

debugger/gdb-sgx-plugin/debug_ext.py
anakrish: I've answered this, but it doesn't seem to show up correctly in github.

The key point is that gdb and lldb cannot call enclave functions like `myst_thread_self()` since it is not supported. Such function calls may or may-not work.

debugger/gdb-sgx-plugin/thread.py
anakrish: One technique we could use is to have function `myst_debug_hook_name(arg1, arg2)` and call them from within the enclave code. To allow setting internal breakpoints by name, also provide empty implementations of these functions on the host side (ensure that the host side empty implementations are not eliminated by the linker).

With such a setup, it ought to be possible to set breakpoints by name.

debugger/gdb-sgx-plugin/debug_ext.py
asvrada: This breakpoint is not actually in use, I will delete it altogether

debugger/gdb-sgx-plugin/thread.py
asvrada: Will create a hook function

debugger/gdb-sgx-plugin/thread.py
asvrada: Will remove this usage

debugger/gdb-sgx-plugin/thread.py
asvrada: Created a hook `myst_debug_hook_thread_exit`

vtikoo (approved):
Thanks for adding this, looks good!

debugger/gdb-sgx-plugin/thread.py
```diff
@@ -0,0 +1,280 @@
...
+INSTANCE_DEBUG_EXT = None
+
+class ThreadCreateBreakpoint(gdb.Breakpoint):
+    def __init__(self):
+        """
+        Set a breakpoing @ kernel/thread.c:_run_thread, this should be right before myst_setjmp(&thread->jmpbuf)
+        Choosing this location because this is where the newly created thread about to run
+        We can get the new thread pid/tid here
+        """
+        super(ThreadCreateBreakpoint, self).__init__('kernel/thread.c:_run_thread', internal=False)
```
anakrish, line 17: Can you test with a newer version of gdb? Newer versions of gdb could break *before* a functions prologue in which case the arguments have not yet been initialized

anakrish (commented):
We also need to make sure that the hook functions are compiled with debug information, and aren't optimized out.

Do we expect the debugger extension to work in a RelWithDebInfo build of Mystikos?

debugger/gdb-sgx-plugin/thread.py
asvrada: > We also need to make sure that the hook functions are compiled with debug information, and aren't optimized out.
> 
> Do we expect the debugger extension to work in a RelWithDebInfo build of Mystikos?

Seems like currently Mystikos doesn't' differentiate between Debug or Release build, so hook functions are not optimized out for now.

asvrada: > Can you test with a newer version of gdb? Newer versions of gdb could break _before_ a functions prologue in which case the arguments have not yet been initialized

Could you elaborate? Would change to readying `%rdi` helps in this case?

vtikoo: > Seems like currently Mystikos doesn't' differentiate between Debug or Release build, so hook functions are not optimized out for now.

@asvrada to safeguard the hooks from being optimized out we could wrap them with `#pragma GCC optimize "-O2"`. Though I'm not sure if the callsite can still be optimized out.

asvrada: > Can you test with a newer version of gdb? Newer versions of gdb could break _before_ a functions prologue in which case the arguments have not yet been initialized

I tested with gdb `GNU gdb (Ubuntu 9.2-0ubuntu1~20.04.1) 9.2` and it doesn't work just as you described...

anakrish: > > Can you test with a newer version of gdb? Newer versions of gdb could break _before_ a functions prologue in which case the arguments have not yet been initialized
> 
> Could you elaborate? Would change to reading `%rdi` helps in this case?

Yes reading from registers would be the safest. We do that elsewhere too:
https://github.com/openenclave/openenclave/blob/df8cb0d1807e80e5f9148dcad3841050a78ed6ad/debugger/gdb-extension/gdb_sgx_plugin.py#L283

anakrish: > > Can you test with a newer version of gdb? Newer versions of gdb could break _before_ a functions prologue in which case the arguments have not yet been initialized
> 
> I tested with gdb `GNU gdb (Ubuntu 9.2-0ubuntu1~20.04.1) 9.2` and it doesn't work just as you described...

I don't remember the exact version now; but we have run into that behavior before.

PR #1422, opened 2022-09-19
Check for SGX support if enclave creation fails

vtikoo:

### DIFF
--- a/tools/myst/host/exec.c
+++ b/tools/myst/host/exec.c
@@ -213,6 +213,34 @@ static void _wait_on_child_threads(void)
     }
 }
 
+#define EXTENDED_FEATURE_FLAGS_FUNCTION 0x7
+void _print_platform_details()
+{
+    uint32_t rax = 0;
+    uint32_t rbx = 0;
+    uint32_t rcx = 0;
+    uint32_t rdx = 0;
+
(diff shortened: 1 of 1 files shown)


asvrada:
Is there a specification that states the meaning for bit 30?

tools/myst/host/exec.c
```diff
@@ -213,6 +213,34 @@ static void _wait_on_child_threads(void)
...
 
+#define EXTENDED_FEATURE_FLAGS_FUNCTION 0x7
+void _print_platform_details()
+{
+    uint32_t rax = 0;
+    uint32_t rbx = 0;
+    uint32_t rcx = 0;
+    uint32_t rdx = 0;
+
+    myst_cpuid_ocall(
```
jxyang, line 224: Why do we need a "ocall"?

vtikoo: Thats the existing cpuid query function.

vtikoo:
> Is there a specification that states the meaning for bit 30?

Bit 30: SGX_LC. Supports SGX Launch Configuration if 1.

From cpuid leaf 07H section in https://www.felixcloutier.com/x86/cpuid.

jxyang (approved):

PR #1428, opened 2022-10-06
adding oesign signing engine test

salsal97:
This test tests the mystikos support for signing engine via oesign. Adopted from https://github.com/openenclave/openenclave/pull/2455/files.


radhikaj (approved):
Thanks for all the work, @salsal97

tests/oesign-test-engine/expected_invalid_parameter_op1.txt
```diff
@@ -0,0 +1,2 @@
+*** ERROR: oe_sgx_sign_enclave_from_engine() failed: result=OE_INVALID_PARAMETER (0x3)
+/home/sagoel/mystikos/build/bin/myst: error: Failed to sign "hello.signed/lib/openenclave/mystenc.so"
```
radhikaj, line 2: Maybe remove your alias?

tests/oesign-test-engine/expected_invalid_signing_key_op.txt
```diff
@@ -0,0 +1,2 @@
+*** ERROR: oe_sgx_sign_enclave_from_engine() failed: result=OE_INVALID_SGX_SIGNING_KEY (0x3b)
+/home/sagoel/mystikos/build/bin/myst: error: Failed to sign "hello.signed/lib/openenclave/mystenc.so"
```
radhikaj, line 2: Remove your name?

tools/myst/host/package.c
```diff
@@ -295,23 +295,24 @@ int _package(int argc, const char* argv[])
...
-                                          signing_engine_name,
-                                          "--signing-engine-key",
-                                          signing_engine_key,
-                                          "--signing-engine-path",
-                                          signing_engine_path};
+        const char* sign_engine_args[] = {
+            argv[0],
+            "sign",
+            rootfs_file,
+            pem_file, // TODO: maybe I need to take this out
```
radhikaj, line 302: Can you make this null and see if it works? You should not need a pem file.

radhikaj: Does oesign need a pem file with the signing engine?

mikbras (approved):
Looks great. Thanks for doing this Saloni!

tests/oesign-test-engine/expected_invalid_parameter_op1.txt
mikbras: Oops! I missed that. Thank you Radhika!

salsal97: Ah thank you for the comment, I was blind to that. Will fing another way to verify output.

vtikoo:
As we are testing `myst sign`, would recommend moving this test to `tests/myst` along with the other myst executable tests. In the same spirit consider dropping the oe suffix where possible. `myst sign` does invoke `oesign` internally, but thats an implementation detail.

tools/myst/host/package.c
salsal97: It does not need one, null seems to work. Will remove the comment. The thing here is we use positional arguments, so if we do not have a pem file it would potentially be a breaking change for everyone..

tools/myst/host/sign.c
```diff
@@ -497,6 +498,7 @@ int _sign(int argc, const char* argv[])
             pem_file,
             NULL,
             NULL,
+            NULL,
```
vtikoo, line 501: Did the `oesign()` signature change?

tools/myst/host/package.c
radhikaj: Can we let people specify "NULL" as a command line param.

salsal97: Definitely, they can already do that. Tested it out

salsal97: Added that to this test so we can maintain is as the recommended way to use signing engines

tests/oesign-test-engine/expected_invalid_signing_key_op.txt
salsal97: Removed

PR #1435, opened 2022-11-10
myst-gdb extension: Memory manager data structure tracker

vtikoo:
PR adds support for tracking the various memory manager data structures. From a user perspective, it currently allows querying for mapping details given an address. With the data structures being tracked its also possible to print mappings by process(not part of this PR), equivalent of `info proc mappings` in gdb.

Current querying experience:

For MAP_SHARED mappings -
```
(gdb) myst-mman addr
address: addr = 0x10436a000
start_addr: 0x10436a000 end_addr: 0x10436b000 size=4096(0x1000)
Page protection = PROT_READ|PROT_WRITE
MAP_SHARED mapping, type: Regular file
File mapping info: offset=0 filesz=10
Owning processes: 
101
102
```

For MAP_PRIVATE mappings -
```
(gdb) myst-mman $rip
address: $rip = 0x10438c9fe
start_addr=0x10438c000 end_addr=0x10438e000 size=8192(0x2000)
Page protection = PROT_READ|PROT_EXEC
MAP_PRIVATE mapping, owning process: 101
File mapping info: offset=4096 filesz=25456
```

### DIFF
--- a/debugger/gdb-sgx-plugin/mman.py
+++ b/debugger/gdb-sgx-plugin/mman.py
@@ -36,10 +36,10 @@ class myst_vad_t:
     OFFSETOF_SIZE = 24
     SIZEOF_SIZE = 8
 
-    OFFSETOF_PROT = 30
+    OFFSETOF_PROT = 32
     SIZEOF_PROT = 2
 
-    OFFSETOF_FLAGS = 32
+    OFFSETOF_FLAGS = 34
     SIZEOF_FLAGS = 2
 
     def __init__(self, addr):
@@ -50,9 +50,10 @@ class myst_vad_t:
         self.flags = read_int_from_memory(addr + self.OFFSETOF_FLAGS, self.SIZEOF_FLAGS)
 
 class mman_file_handle_t:
+    OFFSETOF_FS = 16
     SIZEOF_FS = 8
 
-    OFFSETOF_INODE = 8
+    OFFSETOF_INODE = 32
     SIZEOF_INODE = 8
 
     def __init__(self, addr):
@@ -60,7 +61,7 @@ class mman_file_handle_t:
             self.fs = None
             self.inode = None
             return
-        self.fs = read_int_from_memory(addr, self.SIZEOF_FS)
+        self.fs = read_int_from_memory(addr + self.OFFSETOF_FS, self.SIZEOF_FS)
         self.inode = read_int_from_memory(addr + self.OFFSETOF_INODE, self.SIZEOF_INODE)
 
     @staticmethod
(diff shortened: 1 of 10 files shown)


debugger/gdb-sgx-plugin/mman.py
```diff
@@ -0,0 +1,506 @@
...
+# Licensed under the MIT License.
+
+import gdb
+import math
+import tempfile
+
+def round_up_page_size(x):
+    return ((x + 4095) // 4096) * 4096
+
+class myst_list_t:
```
anakrish, line 11: We need to put compile time asserts locking down the layout and size of these structs used by the debugger.

debugger/gdb-sgx-plugin/mman.py
```diff
@@ -0,0 +1,506 @@
...
+        if not f1 and not f2:
+            return True
+        elif not f1 or not f2:
+            return False
+        elif f1.fs == f2.fs and f1.inode == f2.inode:
+            return True
+        return False
+
+FDMAPPING_SIZE = 32
+FDMAPPING_USED_MAGIC = 0x1ca0597f
```
anakrish, line 77: The magic value also needs to be locked down via compile time assert.

debugger/gdb-sgx-plugin/mman.py
```diff
@@ -0,0 +1,506 @@
...
+
+    def stop(self):
+        mman_tracker.shared_mappings = int(gdb.parse_and_eval('(uint64_t)&_shared_mappings'))
+
+class MystMmanInitBreakpoint(gdb.Breakpoint):
+    def __init__(self):
+        super(MystMmanInitBreakpoint, self).__init__('myst_mman_init_debug_hook', internal=False)
+
+    def stop(self):
+        mman_tracker.mman_base = int(gdb.parse_and_eval('(uint64_t)_mman->base'))
```
anakrish, line 189: Ideally we'd want to rely on offsets instead of field names. Field names require debugging information.

Also ideally, we'd use register instead of variable name _mman.

anakrish (commented):
Good work!
At a high-level looks good to me. Somebody more familiar with the details of mman implementations need to do a deeper review to make sure that the debugger does indeed present correct information.

One core aspect that is missing is that data structures used by the debugger form part of the binary contract of the debugger. These data structures need to be locked down via compile time assertions.

vtikoo:
> One core aspect that is missing is that data structures used by the debugger form part of the binary contract of the debugger. These data structures need to be locked down via compile time assertions.

@anakrish Thanks for pointing this out. Seems like there are two things required - getting offset of a struct field, and adding a compile time assert. I was looking at how openenclave does these. For offset information, the macro is compiler dependent. And for the static assert, I wasn't to fully grasp what the code achieved - 

```C
/* OE_CONCAT */
#define __OE_CONCAT(X, Y) X##Y
#define OE_CONCAT(X, Y) __OE_CONCAT(X, Y)

/* OE_STATIC_ASSERT */
#define OE_STATIC_ASSERT(COND)       \
    typedef unsigned char OE_CONCAT( \
        __OE_STATIC_ASSERT, __LINE__)[(COND) ? 1 : -1] OE_UNUSED_ATTRIBUTE
```

Since we only support gcc/clang, for offset information we can probably directly use  `__builtin_offset`. For compile time assert thinking of using [`static_assert`, which seems to be available since C11](https://www.educative.io/answers/what-is-staticassert-in-c).
Thoughts?

debugger/gdb-sgx-plugin/mman.py
vtikoo: By using register are you suggesting passing a pointer to `_mman` to the debugger hook function, and then reading off `rdi`?

anakrish:
@vtikoo Assuming the static assert is specified on line 100, the macro evaluates to `typedef unsigned char __OE_STATIC_ASSERT_100_number[1]` when true and `typedef unsigned char __OE_STATIC_ASSERT_100_number[-1]` when false. the latter results in a compile error since array sizes cannot be negative.

I'm surprised that Mystikos already doesn't have a macro for compiler time assertions. If it is OK to use C11, I'd prefer using what you suggested since it is a first class compiler concept. Otherwise, you could use an OE type macro.

debugger/gdb-sgx-plugin/mman.py
```diff
@@ -36,10 +36,10 @@ class myst_vad_t:
     OFFSETOF_SIZE = 24
     SIZEOF_SIZE = 8
 
-    OFFSETOF_PROT = 30
+    OFFSETOF_PROT = 32
```
anakrish, line 39: We also need to lockdown that this value matches the C source code.

Here is one strategy of locking down:
https://github.com/openenclave/openenclave/search?q=ENCLAVE_OFFSETOF_PATH

debugger/gdb-sgx-plugin/mman.py
```diff
@@ -0,0 +1,530 @@
...
+                sm = shared_mapping_t(shmem_list.next)
+        return -1
+
+    def _get_map_by_addr(self, addr_str):
+        # Evaluate the address expression.
+        addr = int(gdb.parse_and_eval(addr_str))
+        print('address: %s = 0x%x' % (addr_str, addr))
+
+        if not self._is_valid_heap_addr_range(addr, 0):
+            print('invalid address!! Not in mman controlled region.')
```
asvrada, line 459: (Minor) Should this be uppercase (Invalid) ?

debugger/gdb-sgx-plugin/mman.py
```diff
@@ -0,0 +1,530 @@
...
+
+    OFFSETOF_NMAPS=20
+    SIZEOF_NMAPS=24
+
+    def __init__(self, addr):
+        if addr:
+            self.prev = read_int_from_memory(addr, self.SIZEOF_PREV)
+        else:
+            return
+
```
asvrada, line 119: Suggest:
```
if not addr:
    return

self.prev = xxx
```

debugger/gdb-sgx-plugin/mman.py
```diff
@@ -0,0 +1,530 @@
...
+    OFFSETOF_MMAN_FILE_HANDLE = 24
+    SIZEOF_MMAN_FILE_HANDLE = 8
+
+    def __init__(self, addr):
+        if addr:
+            self.used = read_int_from_memory(addr + self.OFFSETOF_USED, self.SIZEOF_USED)
+        else:
+            return
+
```
asvrada, line 97: Same as above

asvrada (approved):

radhikaj:

PR #1442, opened 2022-12-05
Port of debug-malloc to CRT

mikbras:


vtikoo:
Nightly manual started [here](https://openenclaveci.westus.cloudapp.azure.com/blue/organizations/jenkins/Mystikos%2FNightly-Pipeline-Manual/detail/Nightly-Pipeline-Manual/96/pipeline)

tests/debugmalloc/debugmalloc.c
```diff
@@ -0,0 +1,69 @@
...
+
+void a()
+{
+    a_ptr = malloc(99);
+    b();
+}
+
+size_t debug_malloc_check(bool print_allocations)
+{
+    const long SYS_myst_debug_malloc_check = 2021;
```
jxyang, line 35: I prefer not hardcoding the syscall number. We should include <syscallext.h> instead. See "tests/interrupt" as an example.

include/myst/syscallext.h
```diff
@@ -30,6 +30,8 @@ typedef enum
     SYS_myst_get_exec_stack_option = 2018,
     SYS_myst_interrupt_thread = 2019,
     SYS_myst_pre_launch_hook = 2020,
+    SYS_myst_debug_malloc_check = 2021,
+    SYS_myst_mcheck = 2022,
```
jxyang, line 34: What's SYS_myst_mcheck for?

crt/debugmalloc.c
```diff
@@ -0,0 +1,538 @@
...
+MYST_NOINLINE
+size_t myst_backtrace_impl(void** start_frame, void** buffer, size_t size)
+{
+    void** frame = start_frame;
+    size_t n = 0;
+    const int prot = PROT_READ;
+
+    while (n < size)
+    {
+        if (myst_mcheck(frame, 1, PROT_READ) != 0 ||
```
vtikoo, line 42: A stronger check might be that all subsequent values of `frame` belong to the same mapping as `start_frame`, i.e the saved rbp's are from the same stack.

crt/debugmalloc.c
```diff
@@ -0,0 +1,538 @@
...
+/* weak musl libc functions used by this module */
+extern void* libc_malloc(size_t size);
+extern void libc_free(void* ptr);
+extern void* libc_memalign(size_t alignment, size_t size);
+extern void* libc_calloc(size_t nmemb, size_t size);
+extern void* libc_realloc(void* ptr, size_t size);
+extern size_t libc_malloc_usable_size(void* ptr);
+
+MYST_NOINLINE
+size_t myst_backtrace_impl(void** start_frame, void** buffer, size_t size)
```
vtikoo, line 34: This assumes that the application is not compiled with `-fomit-frame-pointer`. Maybe we should surface this in the documentation, so that users can be aware and build accordingly?

crt/debugmalloc.c
```diff
@@ -0,0 +1,538 @@
...
+    void* addrs[BACKTRACE_MAX];
+    uint64_t num_addrs;
+
+    uint64_t padding;
+
+    /* Contains HEADER_MAGIC2 */
+    uint64_t magic2;
+
+    /* User data */
+    uint8_t data[];
```
vtikoo, line 132: Does this field count towards the size of the structure?

crt/debugmalloc.c
```diff
@@ -0,0 +1,538 @@
...
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <assert.h>
+#include <stdlib.h>
+
+#include <myst/spinlock.h>
```
vtikoo, line 12: Are the myst includes intentional?

crt/debugmalloc.c
```diff
@@ -0,0 +1,538 @@
...
+    header_t* head;
+    header_t* tail;
+} list_t;
+
+static list_t _list = {NULL, NULL};
+static myst_spinlock_t _spin = MYST_SPINLOCK_INITIALIZER;
+
+static void _list_insert(list_t* list, header_t* header)
+{
+    myst_spin_lock(&_spin);
```
vtikoo, line 257: Wondering if this is a layering violation.

crt/debugmalloc.c
```diff
@@ -0,0 +1,538 @@
...
+            return NULL;
+
+            /* fill block with 0xaa (allocated) bytes */
+#ifdef ENABLE_MALLOC_MEMSET
+        memset(block, 0xAA, block_size);
+#endif
+
+        header_t* header = (header_t*)block;
+        INIT_BLOCK(header, 0, size);
+        _check_block(header);
```
vtikoo, line 367: Can this check ever fail?

utils/syscall.c
```diff
@@ -772,6 +772,7 @@ __attribute__((__unused__)) static void _check_myst_syscalls(void)
         case SYS_myst_get_exec_stack_option:
         case SYS_myst_interrupt_thread:
         case SYS_myst_pre_launch_hook:
+        case SYS_myst_debug_malloc_check:
```
vtikoo, line 775: SYS_myst_mcheck also needs to be added here. This is causing the compilation to fail in CI.

tests/debugmalloc/debugmalloc.c
mikbras: I don't like hardcoding it either, but this is intended as a pattern for the user who will not have access to <syscallext.h>, which is not installed. Should we consider installing this file?

mikbras:
I reworked the PR so the ``debugmalloc.c`` defined in the **kernel** can be shared with the **crt** as is. This required adding some glue code in the **crt**.

tests/debugmalloc/debugmalloc.c
mikbras: I could find no way to define a function in a user app so this solution depends on making a syscall. We might have other situations and maybe we should consider a "universal syscall" for all future usages.

include/myst/syscallext.h
mikbras: It is defined as part of ``myst_mman_mcheck()`` and it checks whether a memory range has certain permissions. It is used by the backtrace function.

crt/debugmalloc.c
mikbras: That is an excellent idea, but will it work across stacks?

crt/debugmalloc.c
mikbras: No, its size is zero.

crt/debugmalloc.c
mikbras: Yes, but I removed ``debugmalloc.c`` from crt and now crt uses the one that is in the kernel.

crt/debugmalloc.c
mikbras: Yes, and when it does it prints an error to standard error and aborts.

utils/syscall.c
mikbras: Thank you Vikas. Added.

jxyang (approved):

PR #1444, opened 2023-01-10
Disable pandas testsuite

vtikoo:
Signed-off-by: Vikas Tikoo <vikasamar@gmail.com>

PR #1445, opened 2023-01-12
update pt model link

salsal97:
Also removed pandas solution temporarily as discussed


CyanDevs (approved):
Pandas was removed in https://github.com/deislabs/mystikos/pull/1444 as well but that build seems to be cancelled.

salsal97:
> Pandas was removed in #1444 as well but that build seems to be cancelled.

Yeah I had to remove it here otherwise this PR would not pass the pipeline

PR #1448, opened 2023-01-19
fix memalign error in debug-malloc

mikbras:
This patch fixes a bug in musl libc ``memalign`` when using debug-malloc. The function ``myst_debug_posix_memalign()`` was calling into musl's ``memalign`` function, which was calling back into ``myst_debug_malloc()``, which would have resulted in a double debug-malloc header and trailer. Consequently, ``free()`` crashed when releasing memory obtained with ``memalign()``.

### DIFF
--- a/kernel/debugmalloc.c
+++ b/kernel/debugmalloc.c
@@ -5,9 +5,11 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <assert.h>
 #include <string.h>
 
 #include <myst/backtrace.h>
+#include <myst/printf.h>
 #include <myst/debugmalloc.h>
 #include <myst/defs.h>
 #include <myst/malloc.h>
@@ -19,6 +21,8 @@
 
 #define ENABLE_MALLOC_MEMSET
 #define ENABLE_FREE_MEMSET
+
+// #define DEBUG_MEMALIGN
 
 /*
 **==============================================================================
@@ -171,6 +175,12 @@ MYST_INLINE void* _get_block_address(void* ptr)
     header_t* header = _get_header(ptr);
     const size_t padding_size = _get_padding_size(header->alignment);
     return (uint8_t*)ptr - sizeof(header_t) - padding_size;
+}
+
+MYST_INLINE void* _get_block_address_v2(void* ptr)
+{
+    header_t* header = _get_header(ptr);
+    return (uint8_t*)header - header->padding;
 }
 
 MYST_INLINE size_t _calculate_block_size(size_t alignment, size_t size)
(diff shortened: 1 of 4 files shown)


kernel/debugmalloc.c
```diff
@@ -173,6 +177,12 @@ MYST_INLINE void* _get_block_address(void* ptr)
     return (uint8_t*)ptr - sizeof(header_t) - padding_size;
 }
 
+MYST_INLINE void* _get_block_address_v2(void* ptr)
```
jxyang, line 180: Do we still need `_get_block_address` v1?

kernel/debugmalloc.c
```diff
@@ -393,10 +414,39 @@ int myst_debug_posix_memalign(void** memptr, size_t alignment, size_t size)
     const size_t block_size = _calculate_block_size(alignment, size);
     void* block = NULL;
     header_t* header = NULL;
+    size_t rsize = _round_up_to_multiple(size, sizeof(uint64_t));
```
jxyang, line 417: Why do we need to align the data size to 8 bytes?

kernel/debugmalloc.c
```diff
@@ -393,10 +414,39 @@ int myst_debug_posix_memalign(void** memptr, size_t alignment, size_t size)
...
     if (!memptr)
         return EINVAL;
 
+    /*
+    ** [padding][header][block][footer]
+    ** ^                ^
+    ** |                |
+    ** X                Y
+    **
+    ** Note: both X and Y are on the alignment boundary
```
jxyang, line 431: Who do we need to align the header, but not the footer?

mikbras: Good catch! We align the footer on an 8-byte-boundary but my picture did not reflect that. I updated the picture and added some more explanation.

kernel/debugmalloc.c
mikbras: The data is padded at the end up to an 8-byte boundary. This forces the footer to be aligned on an 8-byte boundary.

kernel/debugmalloc.c
mikbras: We could eliminate v1, but I left it as a cross check in case the two functions fail to produce the same quantity. The v2 version is just intended as a sanity check (so if v1 != v2 then we return EINVAL).

mikbras:
@jxyang: Thank you for your review. I committed changes to the comments to clarify that the footer is aligned. Also, I am on the fence about whether to remove the V1 function. For now, I left it as a cross check in the unlikely event that V1 and V2 don't produce the same result. :)

jxyang (approved):

PR #1456, opened 2023-02-09
Fix broken sigining of EXT2 rootfs

mikbras:
Signing of EXT2 images (using ``myst sign``) was copying the entire EXT2 rootfs into the enclave image rather than a stub.

Added ``./tests/sign`` to verify signing of EXT4-based images. Added negative test too.

### DIFF
--- a/samples/helloworld/ext2rootfs/Makefile
+++ b/samples/helloworld/ext2rootfs/Makefile
@@ -28,7 +28,9 @@ package.pem:
 
 package: package.pem ext2rootfs
 	echo "Generating a signed package"
-	@myst package-sgx --roothash=roothash appdir package.pem ../config.json
+	@myst package-sgx --roothash=roothash package.pem ../config.json
+
+export MYST_ROOTFS_PATH=$(CURDIR)/ext2rootfs
 
 run: package
 	echo "Running Mystikos packaged application. No myst exec-sgx necessary"
--- a/tests/sign/Makefile
+++ b/tests/sign/Makefile
@@ -1,0 +1,59 @@
+TOP=$(abspath ../..)
+include $(TOP)/defs.mak
+
+APPDIR = appdir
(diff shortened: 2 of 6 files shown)


salsal97 (approved):
Thank you for fixing this!

tools/myst/host/sign.c
```diff
@@ -440,10 +443,28 @@ int _sign(int argc, const char* argv[])
     assert(myst_validate_file_path(program_file));
     assert(myst_validate_file_path(temp_oeconfig_file));
 
+    /* if not a CPIO archive, create a zero-filled file with one page */
+    if (myst_cpio_test(rootfs) == -ENOTSUP)
```
jxyang, line 447: Should we check `roothashes_opt` for a quicker determination?

tests/sign/Makefile
```diff
@@ -0,0 +1,56 @@
+TOP=$(abspath ../..)
```
salsal97, line 1: This test should live in the tests/myst directory with the other signing tests

paulcallen (approved):

samples/helloworld/ext2rootfs/Makefile
```diff
@@ -28,7 +28,9 @@ package.pem:
 
 package: package.pem ext2rootfs
 	echo "Generating a signed package"
-	@myst package-sgx --roothash=roothash appdir package.pem ../config.json
+	@myst package-sgx --roothash=roothash package.pem ../config.json
```
radhikaj, line 31: What does it mean when both roothash and appdir are present in args? Should this be prevented(we should not change what we have, just wondering..)

radhikaj: The TEE_aware sample also takes appdir as a command line arg

radhikaj (approved):
Thanks for the fix @mikbras

PR #1465, opened 2023-03-14
add injection attack check

salsal97:
This should fix https://github.com/deislabs/mystikos/security/code-scanning/3479


CyanDevs:
What happens if the user puts in an absolute file path?
Is it enough to consider anything above to the current directory to be uncontrolled?

salsal97:
This PR is not required

PR #1493, opened 2023-08-17
Fix issue 1467

isil-oz:


isil-oz:
@microsoft-github-policy-service agree

kernel/mmanutils.c
```diff
@@ -664,6 +664,21 @@ static int _mmap_fd_checks(int prot, int flags, int fd)
...
+{
+    int ret = 0;
+    long new_usage = _mman.current_usage + length;
+    if(new_usage > _mman.max_usage)
+    {
+        _mman.max_usage = new_usage;
+    }
+
+    _mman.current_usage = new_usage;
+    //printf("max_usage %ld current_usage %ld\n", _mman.max_usage, _mman.current_usage);
```
vtikoo, line 677: Remove commented line

kernel/mmanutils.c
```diff
@@ -664,6 +664,21 @@ static int _mmap_fd_checks(int prot, int flags, int fd)
     return ret;
 }
 
+int set_mem_usage(long length)
```
vtikoo, line 667: Consider function names like `update_mem_usage` or `register_mem_usage`. This function is accounting for the memory usage update caused by the current allocation/deallocation.

kernel/mmanutils.c
```diff
@@ -664,6 +664,21 @@ static int _mmap_fd_checks(int prot, int flags, int fd)
...
+{
+    int ret = 0;
```
vtikoo, line 669: This function always returns 0, consider changing return type to void.

kernel/mmanutils.c
```diff
@@ -751,6 +766,9 @@ long myst_mmap(
         ECHECK(
             myst_mman_mmap(&_mman, addr, length, prot, tflags, (void**)&ret));
 
+        //Update mem usage myst_mmap
```
vtikoo, line 769: As max usage is only displayed if `--perf` was passed on command line, its better to apply the same condition here.

kernel/mmanutils.c
```diff
@@ -841,6 +859,9 @@ void* myst_mremap(
     ECHECK(
         myst_mman_mremap(&_mman, old_address, old_size, new_size, flags, &p));
 
+    //Update mem usage myst_mman_mremap
+    set_mem_usage((long)new_size - (long)old_size);
```
vtikoo, line 863: Same as above, put behind `--perf` option check.

kernel/mmanutils.c
```diff
@@ -928,6 +949,10 @@ int __myst_munmap(void* addr, size_t length, fdlist_t** head_out)
     ECHECK(myst_round_up(length, PAGE_SIZE, &length));
 
     ECHECK(myst_mman_munmap(&_mman, addr, length));
+
+    //Update mem usage myst_mman_munmap
+    set_mem_usage(-1 * (long)length);
```
vtikoo, line 954: Same as above, put behind `--perf` option check.

kernel/enter.c
```diff
@@ -977,6 +980,17 @@ int myst_enter_kernel(myst_kernel_args_t* args)
          */
         myst_set_fsbase(thread->target_td);
 
+        /* print out the maximum value it saw during an application's lifetime at shutdown */
```
vtikoo, line 983:
```suggestion
        /* print out the maximum memory usage seen during an application's lifetime at shutdown */
```

vtikoo (changes requested):
Thanks for the PR @isil-oz, overall looks good! Please take a look at the comments.
Also please squash your commits into a single commit, and provide a meaningful commit message.

vtikoo:
We don't have `--perf` documented in the usage strings, please document the existing functionality and the new additions you've made. 
The usage strings are here -
Linux target - https://github.com/deislabs/mystikos/blob/main/tools/myst/host/exec_linux.c#L126
SGX target - https://github.com/deislabs/mystikos/blob/main/tools/myst/host/exec.c#L461

For existing `--perf` functionality you can add - reports time taken to boot and on shutdown reports time spent on each syscall.

isil-oz:
> We don't have `--perf` documented in the usage strings, please document the existing functionality and the new additions you've made. The usage strings are here - Linux target - https://github.com/deislabs/mystikos/blob/main/tools/myst/host/exec_linux.c#L126 SGX target - https://github.com/deislabs/mystikos/blob/main/tools/myst/host/exec.c#L461
> 
> For existing `--perf` functionality you can add - reports time taken to boot and on shutdown reports time spent on each syscall.

Done.

vtikoo (approved):
PR looks good! @mikbras had a suggestion about "peak_usage" being more clear than "max_usage"(and I agree).

isil-oz:
> PR looks good! @mikbras had a suggestion about "peak_usage" being more clear than "max_usage"(and I agree).

Updated "max" with "peak".

PR #1495, opened 2023-09-13
Raising heap allocation size to accomodate for OpenSSL 3

justanotherminh:
OpenEnclave, which is used by Mystikos to perform enclave-related tasks, now uses OpenSSL 3 in the enclave side. OpenSSL 3 requires significantly more heap memory than OpenSSL 1, which is why gencreds was failing due to generate_x509_self_signed_certificate not having enough heap memory.


asvrada (approved):

CyanDevs (changes requested):
Could you please add a bit more context to your PR description so we understand why we need to increase the heap size?

Also, please don't include merge commits in your PR: [Merge branch 'deislabs:main' into openssl3](https://github.com/deislabs/mystikos/pull/1495/commits/836d89c07dbaecb6e1ffc3a8e9a42064d5c0db96)

Thanks!

CyanDevs (approved):
Thanks for the changes!

vtikoo (approved):
LGTM, thanks for tracking down this issue!

PR #1499, opened 2023-09-22
Update Makefile to stop referencing defs from parent repo and use mys…

radhikaj:
…t from package rather than build tree

### DIFF
--- a/samples/confidential_ml/Makefile
+++ b/samples/confidential_ml/Makefile
@@ -1,7 +1,3 @@
-TOP = $(abspath ../..)
-include $(TOP)/defs.mak
-
-APPBUILDER    = $(TOP)/scripts/appbuilder
 APP_NAME      = python3
 
 APP_PATH      = /usr/local/bin/python3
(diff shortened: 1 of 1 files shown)


vtikoo:
Would this have an effect on pipelines running this sample?

radhikaj:
Looks like the PR pipeline is green on this PR. So I dont believe so.
This sample was using myst from the build tree, the others were using myst from the Mystikos install. The latter is the preference since samples are developer facing.

samples/confidential_ml/Makefile
```diff
@@ -13,12 +9,12 @@ SHUTDOWN_WAIT = 2
...
 
 appdir:
-	$(APPBUILDER) -m -d Dockerfile
+	@myst-appbuilder -m -d Dockerfile
 
 $(PACKAGE): appdir package.pem
-	$(MYST) mkext2 appdir rootfs
-	$(MYST) fssig --roothash rootfs > roothash
-	$(MYST) package-sgx --roothash=roothash package.pem config.json
+	@myst mkext2 appdir rootfs
```
vtikoo, line 15: I'd prefer having it without the @, that way we can see what commands were run in the pipeline logs. For end-user it is not a big difference in experience.

PR #1507, opened 2023-11-02
make the dockerfile stand-alone

salsal97:


asvrada (approved):

samples/docker_aks/Dockerfile
```diff
@@ -1,7 +1,17 @@
+FROM ubuntu:20.04 as builder
+
+RUN apt-get update && apt-get install -y wget gcc
```
CyanDevs, line 3: nit: good practice to apt clean and remove temp files afterwards -- this prevents unnecessary files in the layer
curl may also be an option to use so you don't have to install wget

samples/docker_aks/Dockerfile
```diff
@@ -1,7 +1,17 @@
...
+
+# Using these as links so that this Dockerfile is stand-alone
+RUN wget https://raw.githubusercontent.com/deislabs/mystikos/main/samples/docker_aks/hello.c
+RUN wget https://raw.githubusercontent.com/deislabs/mystikos/main/samples/docker_aks/config.json
```
CyanDevs, line 7: nit: you can wget <a> <b> to merge these layers

samples/docker_aks/Makefile
```diff
@@ -5,10 +5,8 @@ APPDIR        := appdir
 BUILD_DIR     := $(CURDIR)
 CFLAGS        := -fPIC
 DOCKER_IMG    := mystikos-hello
-MYST_TARBALL := mystikos.tar.gz
-PSW_VERSION   := 2.15.100
-MYST_RELEASE_VERSION := 0.5.0
-PACKAGE := myst/bin/hello
+MYST_TARBALL  := mystikos.tar.gz
+PACKAGE 	  := myst/bin/hello
```
CyanDevs, line 9: nit: spacing

samples/docker_aks/Makefile
```diff
@@ -5,10 +5,8 @@ APPDIR        := appdir
 BUILD_DIR     := $(CURDIR)
 CFLAGS        := -fPIC
 DOCKER_IMG    := mystikos-hello
-MYST_TARBALL := mystikos.tar.gz
-PSW_VERSION   := 2.15.100
```
CyanDevs, line 9: I've been trying to find where PSW_VERSION and MYST_RELEASE_VERSION was used in the Dockerfile but I don't see it anywhere. Were these just unused which is why you removed them?

CyanDevs (approved):

samples/docker_aks/Makefile
salsal97: Yeah they were used in an older version

samples/docker_aks/Dockerfile
salsal97: Would they not get deleted automatically since this is the first layer?

CyanDevs: Yup, it would still be discarded anyways in the final image so it's not important to minimize this intermediate layer.

PR #1512, opened 2024-01-02
language check for mystikos

salsal97:


ext2/ext2.c
```diff
@@ -2411,7 +2411,7 @@ static int _ftruncate(ext2_t* ext2, myst_file_t* file, off_t length, bool isdir)
     }
     else if (length > file_size)
     {
-        /* make file larger (with a hole) */
+        /* make file larger (with a blank space) */
```
vtikoo, line 2414: A file hole is a standard file system concept. Is it possible to turn off the check explicitly for this block of code?

salsal97: The software takes in files at a time, so unfortunately, it isnt possible to remove a chunk. However, I'ved updated the term to "a file hole" which should keep the concept as well as fix the issue for us. 

Also - all the fixes I made were for sev 1s and 2s.

CyanDevs (approved):

PR #1521, opened 2024-04-20
Create parent directories of mount point target

vtikoo:
During startup mystikos creates mount point target directories if they do not exist, but not its parent directories. This PR changes that to create the intermediate path components as well.


mikbras (approved):
Looks good. Thank you for doing this.

PR #1524, opened 2024-04-29
Updates to Azure SDK and confidential_ml sample

CyanDevs:
* Update storage account
* Clean up unused variables
* Add retry to pycurl download
* Add wait time limit to prevent infinite waits


samples/confidential_ml/src/service_app.py
```diff
@@ -42,17 +42,38 @@ def evaluate_sample():
...
+            cl.setopt(cl.URL, url)
+            #cl.setopt(pycurl.VERBOSE, 1)
+            cl.setopt(pycurl.FOLLOWLOCATION, 1)
+            cl.setopt(pycurl.WRITEFUNCTION, output.write)
+            cl.perform()
+            cl.close()
+            if output.getvalue():
+                # aes_obj.decrypt() expects input to be a multiple of 16 in length
+                # because AES encrypts in blocksize of 16 bytes.
+                if len(output.getvalue()) % 16 != 0:
```
vtikoo, line 60: Just curious, is this to heuristic to prevent download errors?

vtikoo (approved):
LGTM!

samples/confidential_ml/src/service_app.py
CyanDevs: Yes, there are two cases that are handled here:
1. (transient) network issues, which results in pycurl.error and will be retired up to `max_retries`
2. A successful request but an unexpected response is returned. For example, Azure storage API can return an error instead of the file.

radhikaj (approved):

PR #1529, opened 2024-06-25
Move SQL DB to useast2

CyanDevs:
SQL DB moved to useast2 (from useast and canadacentral). This affects Code Coverage, and solutions tests. Specifically, the python_app and sql_ae solutions were updated to use MSI. There is still support for user/password even though we do not use it in the test.

This also updates to ODBC Driver 18 and to use Entra ID for authentication with MSI.


CyanDevs:
Note: code coverage tests need to be fixed before this is thoroughly tested.

radhikaj (approved):

asvrada (approved):

